PR #192, opened 2017-04-26
detect user's os, move to the top and open corresponding tab

ogogorev:


bigbes:
Merged by hand. Thank you!

PR #462, opened 2018-03-15
fix SUBSCRIBE command code

rkravchik:


rkravchik:
1.9, 2.0 has the same mistake.

Issue #575, opened 2018-07-23
[vshard] Double i in the caption and in the url


msiomkin:
https://tarantool.io/en/doc/1.9/reference/reference_rock/vshard/#archiitecture

lenkis:
Closed within PR https://github.com/tarantool/doc/pull/658

Issue #602, opened 2018-08-21
[vshard] Document vshard.storage.bucket_ref/unref


TarantoolBot:
Bucket ref is an in-memory counter that works like
bucket_pin/unpin, but has several differences:

* It is never persisted. It is ok since refs are intended to
  forbid bucket transfer during a request execution, but on
  restart all requests are dropped;

* Has 2 types: RO and RW - read-only and read-write. When a
  bucket has RW refs, it can not be sent. But when the rebalancer
  wants it to be sent, it locks the bucket for new write
  requests, waits until all current ones are finished and sends
  it. When a bucket has RO refs, it can be sent, but can not be
  dropped. Such a bucket can even enter GARBAGE or SENT state,
  but its data is kept until last reader left. A bucket can have
  both RO and RW refs;

* Evidently, it is countable.

Full API:
```Lua
--
-- Make RO/RW ref.
-- @param bucket_id Identifier of a bucket to ref.
-- @param mode 'read' or 'write'.
--
-- @retval true Success.
-- @retval nil, error Can not ref. An error object is
--         in a second value.
vshard.storage.bucket_ref(bucket_id, mode)

--
-- Remove RO/RW ref.
-- @param bucket_id Identifier of a bucket to unref.
-- @param mode 'read' or 'write'.
--
-- @retval true Success.
-- @retval nil, error Can not unref. An error object
--         is in a second value.
--
vshard.storage.bucket_unref(bucket_id, mode)

-- Aliases for the functions above with the predefined
-- mode.
vshard.storage.bucket_refro/refrw
vshard.storage.bucket_unrefro/unrefrw
```
Requested by @Gerold103 in https://github.com/tarantool/vshard/commit/630e6076383c8a8304279009b871ddcecc939552.

Gerold103:
Also, please, note that ref/unref is called automatically when `vshard.router.call` is used or `vshard.storage.call`. For raw API like `r = vshard.router.route() r:callro/callrw` you should ref for yourself inside your function. Be attentive that you **shall** call unref after ref, or else the bucket will be pinned to the storage until the instance restart.

Gerold103:
Besides, you can watch how many refs a bucket has via `vshard.storage.buckets_info([bucket_id])` - bucket_id here is optional. Example:
```YAML
vshard.storage.buckets_info(1)
---
- 1:
    status: active
    ref_rw: 1
    ref_ro: 1
    ro_lock: true
    rw_lock: true
    id: 1
```

[...]

Issue #591, opened 2018-08-08
 [vshard] Document vshard.router.new


TarantoolBot:
Now vshard supports multiple routers in one instance. For this a new API method is exposed:
```Lua
--
-- Create a new router and configure it with @a cfg.
-- @param name Router name, accessible via router_object.name.
-- @param cfg Router configuration.
--
-- @retval Router instance on success.
-- @retval Nil and error object on an error.
vshard.router.new(name, cfg)
```
This method creates a new router instance named `name` and configures it with `cfg`. The later is the same parameter as for `vshard.router.cfg`. `Name` is used as a prefix in logs of the corresponding router and should be unique in the instance. Each router can be connected to any vshard cluster. To the same cluster multiple routers can be connected. A router created via `new` works exactly the same as a static one but for its methods `:method_name(...)` should be used instead of `.method_name(...)`. The old API still works (as `vshard.router.method_name(...)`). Additionally, an old router can be taken via `vshard.router.static` and used as any other created via `new`. **Note** that the `box.cfg` is shared among routers of a single instance.
Requested by @Gerold103 in https://github.com/tarantool/vshard/issues/130.

Issue #577, opened 2018-07-23
[vshard] Tuples vs fields


msiomkin:
Why tuples? I think fields are meant.

https://tarantool.io/en/doc/1.9/reference/reference_rock/vshard/

The _bucket system space
The _bucket system space of each replica set stores the ids of buckets present in the replica set. The space contains the following **tuples:**

bucket – bucket id
status – state of the bucket
destination – uuid of the destination replica set

PR #658, opened 2018-11-14
vshard fidxes

lenkis:
A cumulative update for 4 tickets closed by @llelik8
gh-576 - Ambiguous vshard bucket status description
gh-577 - Tuples vs fields
gh-591 - Document router.new
gh-602 Add bucket ref description and API


lenkis:
Also closes gh-575 Double i in the caption and in the url

PR #918, opened 2019-09-21
Freebsd doc update

daydrim:
Update FreeBSD download os-installation 2.0 section


lenkis:
Site sources were migrated to another repo. I applied download-related changes there, here's the result: https://www.tarantool.io/en/download/os-installation/2.2/freebsd/

Issue #874, opened 2019-08-22
 tarantoolctl allows to start instances without a box.cfg{}


TarantoolBot:
tarantoolctl now works for instances without box.cfg{} or
with delayed box.cfg{} call. This can be managed instances which receive
configuration from external server.

For such instances `tarantoolctl start` goes to background when
box.cfg{} is called, so it will wait until options for box.cfg are received.
However this is not the case for daemon management systems like systemd,
as they handle backgrounding on their side.
Requested by @mmelentiev-mail in https://github.com/tarantool/tarantool/commit/17df9edf013fd262be1b2ffd34d5080d20bc716e.

v-izmalkov:
Have not find any info leading to believe its version 2.2 specific. [This](https://github.com/tarantool/tarantool/commit/17df9edf013fd262be1b2ffd34d5080d20bc716e) patch is in master.

v-izmalkov:
Checked in 2.2 and master (c) Author. 
So this branch to be merged in 2.2. No such changes to 1.10.

PR #941, opened 2019-10-03
Fix gh-874 tarantoolctl allows to start instances without a box.cfg{}

v-izmalkov:
Added a description to Tarantoolctl reference. Didn't find any more suitable places.

### DIFF
--- a/doc/1.10/reference/tarantoolctl.rst
+++ b/doc/1.10/reference/tarantoolctl.rst
@@ -48,6 +48,18 @@ Commands for managing Tarantool instances
         Additionally, this command sets the TARANTOOLCTL environment variable to
         'true', to mark that the instance was started by ``tarantoolctl``.
 
+        .. NOTE::
+
+            ``tarantoolctl`` works for instances without ``box.cfg{}`` called or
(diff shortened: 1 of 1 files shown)


doc/1.10/reference/tarantoolctl.rst
```diff
@@ -48,6 +48,18 @@ Commands for managing Tarantool instances
         Additionally, this command sets the TARANTOOLCTL environment variable to
         'true', to mark that the instance was started by ``tarantoolctl``.
 
+        .. NOTE::
+
+            ``tarantoolctl`` works for instances without ``box.cfg{}`` called or
+            with delayed ``box.cfg{}`` call.
+
+            For example, this can be managed instances which receive configuration
```
lenkis: Bad English + some markup issues, please fix:

For example, this can be used to manage instances which receive
configuration from an external server. For such instances, ``tarantoolctl start``
goes to background when ``box.cfg{}`` is called, so it will wait until options
for ``box.cfg`` are received. However this is not the case for daemon
management systems like ``systemd``, as they handle backgrounding on
their side.

Issue #586, opened 2018-08-01
Document maximum tuple count in space


kyukhin:
Number of tuples in space is limited.
Need to describe this limit in documentation.

sofiayem:
Waiting for the exact number of tuples (the limit). Cannot document without it.

lenkis:
@kyukhin please let us know the limit

kyukhin:
It seems that there's no limit.

lenkis:
Let's document this explicitly in https://www.tarantool.io/en/doc/1.10/book/box/data_model/#tuple

olegrok:
This ticket says that https://github.com/tarantool/tarantool/issues/3594 tuple count is limited

@kyukhin @lenkis

olegrok:
Let's reopen it. Because of https://github.com/tarantool/doc/pull/1033

I think we should clarify it

Totktonada:
We'll look into this in the scope of https://github.com/tarantool/tarantool/issues/3594 Let's consider this issue blocked for now.

p7nov:
Update: https://github.com/tarantool/tarantool/issues/3594#issuecomment-1767159850
Need to check the limit without a hash index.

p7nov:
Closing: as per comments in the issue tarantool/tarantool#3594, there is no known limit on the tuple count (regardless of indexes)

PR #1033, opened 2019-12-19
Added description of limitations for number of tuples in index

bhavenger:
Since we found out about this limitation of the index size in our own experience ([4678](https://github.com/tarantool/tarantool/issues/4678)) - I am adding information about the limitation in the documentation.


olegrok:
It seems that we should reopen - https://github.com/tarantool/doc/issues/586

lenkis:
blocked by https://github.com/tarantool/tarantool/issues/3594

Issue #1045, opened 2019-12-27
feedback: Tarantool Cartridge developer’s guide


TarantoolBot:
> <…>  workdir = ...,
>   advertise_uri = ...,
>   cluster_cookie = **...,
> |  ...
> })|**
> ...
> 
> 
> The cartridge.cfg() call renders the instance operable vi<…>

https://www.tarantool.io/en/doc/1.10/book/cartridge/cartridge_dev/

it's better to mention box.cfg parameters in the example, like `cartridge.cfg({...}, {memtx_memory=1000000000}`

PR #1076, opened 2020-01-23
fixes gh-1045 add box options to the example

Onvember:


doc/book/cartridge/cartridge_dev.rst
```diff
@@ -135,11 +135,14 @@ initialization function:
    local cartridge = require('cartridge')
    ...
    cartridge.cfg({
+   -- cartridge options example
      workdir = ...,
      advertise_uri = ...,
      cluster_cookie = ...,
```
lenkis: please add sample values to the example instead of "..."

Issue #1070, opened 2020-01-17
feedback: Submodule box.stat


TarantoolBot:
> <…> 
> tarantool> box.stat() -- return 10 tables
> ---
> - DELETE:
>     total: 1873949
>     rps: 123
>   SELECT:
>   <…>

https://www.tarantool.io/en/doc/2.2/book/box/box_stat/

Starting from 2.*, there are 11 tables returned for the request statistics. This one is added for SQL requests:
```
EXECUTE:
  total:
  rps: 
```

PR #1078, opened 2020-01-23
gh-1070 add 11th table to box.stat()

Onvember:


lenkis:
All these changes are already in branch 2.3

PR #1263, opened 2020-04-21
Apply minor fixes to getting_started_php.rst

rybakit:


lenkis:
@rybakit thanks a lot!

Issue #1223, opened 2020-04-07
 always promote error created via box.error() to diag


TarantoolBot:
box.error() is able to accept two types of argument: either pair of code
and reason (box.error{code = 555, reason = 'Arbitrary message'}) or error
object (box.error(err)). In the first case error is promoted to
diagnostic area, meanwhile in the latter - it is not:
```
e1 = box.error.new({code = 111, reason = "cause"})
box.error({code = 111, reason = "err"})
- error: err
box.error.last()
- err
box.error(e1)
- error: cause
box.error.last()
- err
```
From now box.error(e1) sets error to diagnostic area as well:
```
box.error(e1)
- error: cause
box.error.last()
- cause
```
Requested by @Korablev77 in https://github.com/tarantool/tarantool/commit/4bcaf15ecf4cebeaa6608707ef32221a70657ac3.

lenkis:
@Korablev77 is this 2.4-specific?

Korablev77:
@lenkis Yes.

PR #1290, opened 2020-04-29
fix gh-1223 add note for box.error about promoting to diagnostic area

Onvember:


doc/reference/reference_lua/box_error.rst
```diff
@@ -77,10 +77,22 @@ Below is a list of all ``box.error`` functions.
 
     :except: whatever is specified in errcode-number.
 
+    ``box.error()`` is able to accept two types of arguments: either a pair of
```
lenkis, line 80: for easier skimming let's make this a bullet list:

```
    ``box.error()`` accepts two sets of arguments:
    
    * error code and reason (``box.error{code = 555, reason = 'Arbitrary message'}``), or
    * error object (``box.error(err)``).

    In both cases the error is promoted as the last error.
```

Onvember: fixed in d7a0ca6

Issue #1228, opened 2020-04-13
 Document uuid field type.


TarantoolBot:
There's a new field type -- UUID, it accepts values returned by
`uuid.new()`.

The index may be either unique or non-unique, nullable or non-nullable,
and may be a primary key.

The values in an index are ordered lexicographically by their string
representation.

To create an index over a uuid field for space `test`, say:
```
box.space.test:create_index("pk", {parts={1, 'uuid'}})
```
Now you may insert uuids into the space:
```
tarantool> box.space.test:insert{uuid.new()}
---
- [e631fdcc-0e8a-4d2f-83fd-b0ce6762b13f]
...

tarantool> box.space.test:insert{uuid.fromstr('64d22e4d-ac92-4a23-899a-e59f34af5479')}
---
- [64d22e4d-ac92-4a23-899a-e59f34af5479]
...

tarantool> box.space.test:select{}
---
- - [64d22e4d-ac92-4a23-899a-e59f34af5479]
  - [e631fdcc-0e8a-4d2f-83fd-b0ce6762b13f]
...

```
Requested by @sergepetrenko in https://github.com/tarantool/tarantool/commit/b238def8065d20070dcdc50b54c2536f1de4c7c7.

PR #1297, opened 2020-05-12
fix gh-1228 add info about uuid field type and uuid.new()

Onvember:


lenkis:
please add missing field types to this table as well: https://www.tarantool.io/en/doc/2.4/book/box/data_model/#indexed-field-types

Issue #1229, opened 2020-04-13
 API for accessing _session_settings space.


TarantoolBot:
There are two ways of updating values of session settings:
via Lua and SQL.

Lua:
box.session.settings is a table, which is always accessible
to user. The syntax is the following:
`box.session.settings.<setting_name> = <new_value>`.

Example of usage:
```
tarantool> box.session.settings.sql_default_engine
---
- memtx
...

tarantool> box.session.settings.sql_default_engine = 'vinyl'
---
...

```

The table itself represents the (unordered) result of select
from _session_settings space.

SQL:
Instead of typing long UPDATE query one can use the SET SESSION command:
`box.execute([[SET SESSION "<setting_name>" = <new_value>]])`.
Note, that this query is case sensitive so the name must be quoted.
Also, SET SESSION doesn't provide any implicit casts, so <new_value> must
be of the type corresponding to the setting being updated.

Example:
```
tarantool> box.execute([[set session "sql_default_engine" = 'memtx']])
---
- row_count: 1
...

tarantool> box.execute([[set session "sql_defer_foreign_keys" = true]])
---
- row_count: 1
...

```
Requested by @slumber in https://github.com/tarantool/tarantool/commit/6706f6591fffec28a29498973ecec85fbfa5a8e2.

pgulutzan:
Since there is a slight chance that this will cause existing applications to fail, it might be right to put a notice about the change in the version-2.3 manual.

lenkis:
@pgulutzan Peter, please help with this ticket. Feel free to make use of Natalia's commit https://github.com/tarantool/doc/commit/6c5f1aa00d0256901d39730e46364d5dd0d1a50c which is a draft for the Lua part of this ticket.

pgulutzan:
@lenkis 

Re Lua:

The words
"The table itself represents the (unordered) result of select
from _session_settings space."
are not currently true. The selecting is done via the primary-key
index, therefore results are in order by name.

I will be more clear about my suggestion about putting a change
notice in the version-2.3 manual. We have no "Release Notes"
for version 2.3, therefore this should go in "Release Notes"
for Release 2.2.1.
"
Deprecations
    (SQL) SESSION will be a reserved word in a later version so objects named SESSION will be illegal. Part of issue 4711.
"

Re SQL:

I can change sql.rst (without closing the issue) by
adding to the reserved-word list,
modifying the current words about _session_settings
"may be changed by updating the box.space._session_settings system table."
[...]

PR #1306, opened 2020-05-13
gh-1229 add API for accessing _session_settings space

Onvember:


lenkis:
To be finalized by Peter Gulutzan in a separate thread.

Issue #1307, opened 2020-05-13
feedback: Submodule box.space


TarantoolBot:
> <…>ter, then it’s as if
> the before_replace function wasn’t called;
> **|if the value is something else, then execution proceeds,
> inserting|replacing the new value|.**
> 
> However, if a trigger function returns an old tuple, or if a
> <…>

https://www.tarantool.io/en/doc/2.3/reference/reference_lua/box_space/#box-space-before-replace

if value is an other tuple, then it is used for insert/replace.

PR #1356, opened 2020-06-03
fixes gh-1307 fix a phrase and code-blocks mark-up

Onvember:


doc/reference/reference_lua/box_space.rst
```diff
@@ -1024,12 +1024,11 @@ Below is a list of all ``box.space`` functions and members.
...
         * if the value is nil, then the tuple will be deleted;
         * if the value is the same as the old parameter, then no
-          `on_replace`` function will be called and the data
+          ``on_replace`` function will be called and the data
           change will be skipped
         * if the value is the same as the new parameter, then it's as if
           the ``before_replace`` function wasn't called;
-        * if the value is something else, then execution proceeds,
-          inserting|replacing the new value.
+        * if value is an other tuple, then it is used for insert/replace.
```
lenkis, line 1031:
```suggestion
        * if the value is some other tuple, then it is used for insert/replace.
```

Issue #1326, opened 2020-05-21
feedback: Module popen


TarantoolBot:
> <…>s.stdout and opts.stderr to true).
> Possible raised errors are:
> 
> **|“IllegalParams:    incorrect type or value of a parameter”|**
> “IllegalParams:    called on a closed handle”
> “IllegalParams:  <…>

https://www.tarantool.io/en/doc/2.4/reference/reference_lua/popen/#lua-function.popen_handle.read

The error descriptions are quoted and it looks like those descriptions are exact error messages. It is not so, it is list of all possible causes.

See, https://github.com/tarantool/doc/issues/1248 list them w/o quotes.

PR #1361, opened 2020-06-05
gh-1326 fix quoted lines

Onvember:


Onvember:
@Totktonada it will look like italic, without backticks. Like this:
_IllegalParams: incorrect type or value of a parameter_

Will it be ok?

Totktonada:
Okay.

We can play with wording to highlight the fact that it is just descriptions for different situations. It is up to you. I'm okay if we'll just don't format it as a string / code.

doc/reference/reference_lua/popen.rst
```diff
@@ -100,7 +100,7 @@ Below is a list of all ``popen`` functions and handle methods.
              (if failure) ``nil, err``
 
     Possible errors: if a parameter is incorrect, the result is
-    "IllegalParams: incorrect type or value of a parameter".
+    `IllegalParams: incorrect type or value of a parameter`.
```
lenkis, line 103: please remove all kinds of quotes and back-ticks: this should be plain text; otherwise the descriptions still look like exact error messages

Onvember: fixed in 89e9596c310e981d07f3835e085ebb1e6dae9ac4

PR #1447, opened 2020-08-07
Update release policy page

veod32:


doc/dev_guide/release_management.rst
```diff
@@ -10,79 +10,79 @@ Release management
 Release policy
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-A Tarantool release is identified by three digits, for example, 1.7.7.
-We use these digits according to their definitions provided at http://semver.org:
+A Tarantool release is identified by three digits, for example, 1.10.7:
 
-* The first digit stands for MAJOR release. A **major** release may contain
+* The first digit stands for ?a MAJOR release. A MAJOR release may contain
```
Totktonada, line 15: A MINOR release may contain incompatible changes too.

doc/dev_guide/release_management.rst
```diff
@@ -10,79 +10,79 @@ Release management
...
   *incompatible changes*.
-* The second digit stands for MINOR release, it does not contain incompatible
+* The second digit stands for ?a MINOR release. It does not contain incompatible
   changes, and is used for introducing backward-compatible *features*.
```
Totktonada, line 18: Nope. We may introduce a backward-incompatible in a MINOR release.

doc/dev_guide/release_management.rst
```diff
@@ -10,79 +10,79 @@ Release management
...
+into the current alpha release:
 
 * **LTS** is a stable release which does not receive new features, and only gets
-  backward compatible fixes. Hence, following the rules of semver, LTS release
+  backward compatible fixes. Hence, LTS release
   never has its MAJOR or MINOR version increased, and only gets PATCH level
   releases.
 
-* **STABLE** is our current stable release, which may receive new features.
+* **STABLE** are our current stable releases which may not receive new features.
```
Totktonada, line 73: may not → does not?

Totktonada:
> In all supported releases, we also release a PATCH release as soon as we find and fix an outstanding CVE/vulnerability.

@kyukhin Please, decide whether this outstanding situation should be handled this way (via publishing a new relase out of usual timelines).

kyukhin:
> > In all supported releases, we also release a PATCH release as soon as we find and fix an outstanding CVE/vulnerability.
> 
> @kyukhin Please, decide whether this outstanding situation should be handled this way (via publishing a new release out of usual timelines).
> 
> Since our repositories have per-push packages, it is not necessary, maybe. However we should notify users about necessity to update and properly describe reasons.

Maybe just notify users about the patch. I see no reason for tagging new release in the case.

Totktonada:
@veod32 Please, adjust the cited sentence accordingly.

doc/dev_guide/release_management.rst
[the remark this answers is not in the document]
veod32: It's from draft. Fixed.

doc/dev_guide/release_management.rst
```diff
@@ -10,79 +10,79 @@ Release management
...
 
-* The first digit stands for MAJOR release. A **major** release may contain
+* The first digit stands for ?a MAJOR release. A MAJOR release may contain
   *incompatible changes*.
-* The second digit stands for MINOR release, it does not contain incompatible
+* The second digit stands for ?a MINOR release. It does not contain incompatible
   changes, and is used for introducing backward-compatible *features*.
-* The third digit is for PATCH releases that contain only backward-compatible
-  *bug fixes*.
+* The third digit is for [TODO]PATCH releases that contain only
```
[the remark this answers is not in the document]
veod32: Fixed.

doc/dev_guide/release_management.rst
veod32: corrected

doc/dev_guide/release_management.rst
veod32: re-written

doc/dev_guide/release_management.rst
veod32: re-written

veod32:
@Totktonada All review comments have been resolved. Please review once again.

doc/dev_guide/release_management.rst
```diff
@@ -10,92 +10,134 @@ Release management
...
-A Tarantool release is identified by three digits, for example, 1.7.7.
-We use these digits according to their definitions provided at http://semver.org:
+A Tarantool release is identified by three digits, for example, 1.10.7:
+
+* The first digit stands for a MAJOR release that introduces
+  some *major changes*. Up to now, there has been only one major release jump
+  when we delivered the 2.x release series with the SQL support.
+* The second digit stands for a MINOR release that is used for
+  introducing new *features*. Some of the features may have no backward
+  compatibility.
```
Totktonada, line 20: > The second digit stands for a MINOR release that is used for introducing new *features*.

I would also add that we call it a release series. Say, add:

> [...] a MINOR releases (so called release series) [...]

'features without backward compatibility' sounds strange. If a feature was introduced in `x.y` and is not present in `x.(y-1)`, then the term 'backward compatibility' is not applicable here.

I would say about backward compatibility without context of a feature. Backward incompatible change may be introduced as part of refactoring, which reduces code complexity — not as part of a feature. I would rephrase it like:

> A release series may contain backward incompatible changes in the sense that existing Lua, SQL or C code that are run on a current release series may not be run with the same effect on a future series. However we don't exploit this rule and don't make incompatible changes without appropriate reason. We usually deliver information how mature a functionality is through release notes. Please, note that binary data layout is always compatible with a previous series as well as with LTS series (an `x.y` instance can be started on top of `x.(y+1)` or `1.10.z` data), binary protocol is compatible too (as client-server as well as replication protocol).

Hm, in appears to be large. Maybe it worth to just say that backward incompatible changes are possible between release series here and describe exact meaning of this somewhere below.

Let's consider the comment as source of information and rephrase it for better readability, where necessary.

doc/dev_guide/release_management.rst
```diff
@@ -10,92 +10,134 @@ Release management
...
+* The third digit is for PATCH releases that contain only
+  backward-compatible *bug fixes*.
```
Totktonada, line 22: It is right by meaning, but technically it is true since a first beta release within a series. An alpha (`x.y.0`) is not a release in fact, it is more like a point from where a series starts to live. Look at the diagram:

```
2.2 series ---- 2.2.1 ---- 2.2.2 ---- 2.2.3
                  |   
                  V   
2.3 series .... 2.3.0 ---- 2.3.1 ---- 2.3.2 ---- 2.3.3
                             |   
                             V   
2.4 series ............... 2.4.0 ---- 2.4.1 ---- 2.4.2
                                        |   
                                        V   
2.5 series .......................... 2.5.0 ---- 2.5.1
                                                   |   
                                                   V   
2.6 series ..................................... 2.6.0

--------------------- time axis --------------------->
```

`x.y.0` → `x.y.1` is where features are land and backward incompatible changes are possible. After `x.y.1` only backward-compatible bugfixes are land.

Another interesting thing is that `x.y.0` is equivalent of `x.(y-1).1` in practice.

doc/dev_guide/release_management.rst
```diff
@@ -10,92 +10,134 @@ Release management
...
+   such as 2.3.0, 2.4.0, and so on. It is not what an alpha release usually
+   means in the software release life cycle but rather the current trunk version
+   which is under heavy development and can be unstable.
+   The current alpha version always lives in the master branch.
+
+2. **Beta**. When all the features planned are implemented, we fork a new branch
+   from the master branch and tag it as a new beta version.
+   It contains ``1`` for the PATCH digit, e.g., 2.3.1, 2.4.1, and so on.
+   This version is still unstable although there're no critical regressions
+   in it since the last stable release. We do this quarterly as well.
```
Totktonada, line 43: I would move accents a bit:

* From 'is unstable' to 'is not proven by time to be stable (a feature set is just frozen)'.
* From 'no critical regressions' to 'no known critical regressions'.

Like so:

> This version is not proven by time to be stable (a feature set is just frozen) although there're no known critical regressions in it since the last stable release.

doc/dev_guide/release_management.rst
```diff
@@ -10,92 +10,134 @@ Release management
...
-  backward compatible fixes. Hence, following the rules of semver, LTS release
+* **LTS** is a stable release which does not receive new features and only gets
+  backward compatible fixes. Hence, LTS release
   never has its MAJOR or MINOR version increased, and only gets PATCH level
   releases.
 
-* **STABLE** is our current stable release, which may receive new features.
+* **STABLE** are our current stable releases. They also receive no new features
+  and only backward compatible fixes.
   When the next STABLE version is published, MINOR version is incremented.
```
Totktonada, line 77: > When the next STABLE version is published, MINOR version is incremented.

It is true if we tell about a stable release series. Otherwise we should say about PATCH version increment.

In fact this sentence repeats the information above and I don't see much sense to keep it here.

doc/dev_guide/release_management.rst
```diff
@@ -10,92 +10,134 @@ Release management
...
-  cycle described in the beginning. While NEXT release is in alpha state,
-  its MINOR is frozen at 0 and is only increased when the release reaches
-  BETA status. Once the NEXT release becomes STABLE, we switch the vehicle for
-  delivery of minor features, designating the previous stable release as LTS,
-  and releasing it with MINOR set to 10.
+  We maintain PATCH level releases for two STABLE releases,
+  the current and the previous one, to preserve support continuity.
+
+* **NEXT** Beta is our next MINOR release, and it follows the maturity cycle
+  described in the beginning.
```
Totktonada, line 82: Is there any reason to introduce the second term 'next' with the same meaning as 'beta'? I generally don't like this: it is better to have one term for one thing. This way it is easier to understand.

doc/dev_guide/release_management.rst
```diff
@@ -10,92 +10,134 @@ Release management
...
 
 To sum up, once a quarter we release:
 
-* the next LTS release, e.g. 2.10.6, 2.10.7 or 2.10.8
-* the next STABLE release, e.g. 3.6, 3.7 or 3.8
-* (optionally) an alpha or beta version of the NEXT release,
-  e.g. 4.0.1, 4.0.2 or 4.0.3
+* the next LTS release, e.g., 1.10.6, 1.10.7, and or so on
+* the next STABLE release, e.g., 2.4.2 or 2.3.3
```
Totktonada, line 87: In fact we release two stable version simultaneously (draw a vertical line on the diagram above to see visual representation of this).

doc/dev_guide/release_management.rst
```diff
@@ -10,92 +10,134 @@ Release management
...
 
 Example version identifier:
 
-* 2.0.3 - third alpha of 2.0 release
-* 2.1.3 - a beta of 2.0 release
-* 2.2 - a stable version of 2.0 series, but not an LTS yet
-* 2.10 - an LTS release
+* 2.5.0 -- current alpha version under development
+* 2.6.1 -- beta version of the 2.6 release
+* 2.3.3 or 2.4.2 -- stable versions of the 2.3 or 2.4 releases, but not an LTS yet
```
Totktonada, line 102: Not every release series becomes LTS, so 'yet' here looks redundant for me.

doc/dev_guide/release_management.rst
Totktonada: NB: If we'll place the diagram into the documentation, it worth to add some terminator symbol after `x.y.3`.

veod32:
@Totktonada I've resolved all the comments. Also I restructured the content and re-wrote some of the paragraphs as we discussed. Some of the paragraphs are deleted as the contained the duplication of info.
Please review once again.

doc/dev_guide/release_management.rst
```diff
@@ -10,92 +10,166 @@ Release management
...
+If we look at the release diagram above, it means that the bug fixes are to be
+added into 1.10, 2.4, 2.5, and 2.6 release series.
 
-* **LTS (Long Term Support)** releases that are supported for 3 years
-  (community) and up to 5 years (paying customers). **LTS** release
-  is identified by MINOR version 10.
-* **Standard stable releases** are only supported a few months after the next
-  stable is out.
+To sum it up, once a quarter we release (?see the release diagram above for
+reference):
```
Totktonada, line 105: Is this question mark part of the syntax or just typo?

doc/dev_guide/release_management.rst
```diff
@@ -10,92 +10,166 @@ Release management
...
-"Support" means that we continue fixing bugs in a release.
+* next LTS release, e.g., 1.10.7
+* two stable releases, e.g., 2.3.3 and 2.4.2
+* beta version of the next release series, e.g., 2.5.1.
 
-We add commits simultaneously to three MAJOR releases:
+In all supported releases, when we find and fix an outstanding CVE/vulnerability,
+we release a patch for that but do not tag a new PATCH level version.
+Users will be informed about such critical patches via official Tarantool news
+channel (`tarantool_news <https://t.me/tarantool_news>`_).
```
Totktonada, line 114: > we release a patch <...> but do not tag a new PATCH level version.

Maybe it is not good to name the process of pushing a fix as 'release a patch'. It may be confusing in the context. I would think how to get rid of possible ambiguity.

doc/dev_guide/release_management.rst
```diff
@@ -10,92 +10,166 @@ Release management
...
 
-We also publish nightly builds, and use the fourth slot in the version
-identifier to designate the nightly build number.
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Release list
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Example version identifier:
+Below is the table containing all Tarantool releases starting from 1.10.0 up to
+the current latest versions. For each release series, releases are sorted out
```
Totktonada, line 143: Maybe mention a date to don't be obligated to update the table at every release? Like so:

> up to the current latest versions (as on 2020-09-07).

Totktonada (approved):
LGTM.

I gave several minor comments, but they are tiny and there is no need to re-review it with me.

Maybe @kyukhin want to take a glance at the final variant.

kyukhin (approved):

doc/dev_guide/release_management.rst
```diff
@@ -10,92 +10,166 @@ Release management
...
-* **NEXT** is our next MAJOR release, and it follows the maturity
-  cycle described in the beginning. While NEXT release is in alpha state,
-  its MINOR is frozen at 0 and is only increased when the release reaches
-  BETA status. Once the NEXT release becomes STABLE, we switch the vehicle for
-  delivery of minor features, designating the previous stable release as LTS,
-  and releasing it with MINOR set to 10.
+.. note::
 
-To sum up, once a quarter we release:
+   A release series may contain backward incompatible changes in a sense that
```
Totktonada, line 123: Sorry for the late feedback, but I would say 'may introduce' instead of 'may contain'. It would give better understanding where exactly such changes may appear.

doc/dev_guide/release_management.rst
```diff
@@ -10,92 +10,166 @@ Release management
...
+   incoming bugs, we declare this version stable. It is tagged with ``2`` for
+   the PATCH digit, e.g., 2.3.2, 2.4.2, and so on.
+
+   We support such version for 3 months while making another stable release
+   by fixing all bugs found. We release it in a quarter. This last tag
+   contains ``3`` for the PATCH digit, e.g., 2.3.3, 2.4.3, and so on.
+   After the tag is set, no new changes are allowed to the release branch,
+   and it is declared deprecated and superseded by a newer MINOR version.
+
+   Stable versions don't receive any new features and only gets backward
```
lenkis, line 55:
```suggestion
   Stable versions don't receive any new features and only get backward
```

doc/dev_guide/release_management.rst
```diff
@@ -10,92 +10,166 @@ Release management
...
 
-"Support" means that we continue fixing bugs in a release.
+* next LTS release, e.g., 1.10.7
+* two stable releases, e.g., 2.3.3 and 2.4.2
+* beta version of the next release series, e.g., 2.5.1.
 
-We add commits simultaneously to three MAJOR releases:
+In all supported releases, when we find and fix an outstanding CVE/vulnerability,
+we deliver a patch for that but do not tag a new PATCH level version.
+Users will be informed about such critical patches via official Tarantool news
```
lenkis, line 113:
```suggestion
Users will be informed about such critical patches via the official Tarantool news
```

doc/dev_guide/release_management.rst
```diff
@@ -10,92 +10,166 @@ Release management
...
 
-* the next LTS release, e.g. 2.10.6, 2.10.7 or 2.10.8
-* the next STABLE release, e.g. 3.6, 3.7 or 3.8
-* (optionally) an alpha or beta version of the NEXT release,
-  e.g. 4.0.1, 4.0.2 or 4.0.3
+   Please note that binary data layout
+   is always compatible with the previous series as well as with the LTS series
+   (an instance of ``X.Y`` version can be started on top of ``X.(Y+1)``
+   or ``1.10.z`` data); binary protocol is compatible too
+   (as client-server as well as replication protocol).
```
lenkis, line 134:
```suggestion
   (client-server as well as replication protocol).
```

doc/dev_guide/release_management.rst
```diff
@@ -10,92 +10,166 @@ Release management
...
-identifier to designate the nightly build number.
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Release list
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Example version identifier:
+Below is the table containing all Tarantool releases starting from 1.10.0 up to
+the current latest versions (as of September 1, 2020). For each release series,
+releases are sorted out as alpha, beta, and stable ones.
+
```
lenkis, line 145: Why no "table" container? Like this:

```
.. container:: table

    .. rst-class:: left-align-column-1
    .. rst-class:: left-align-column-2 
```

lenkis (changes requested):

doc/dev_guide/release_management.rst
veod32: Corrected.

Issue #986, opened 2019-11-15
feedback: Tarantool Cartridge developer’s guide


TarantoolBot:
> **|A Tarantool Cartridge cluster segregates instance functionality in a role-based
> way. |Cluster** roles are Lua modules that implement some instance-speci<…>

https://www.tarantool.io/en/doc/2.2/book/cartridge/cartridge_dev/#cluster-roles

I cannot find anywhere info that role is enabled per replicaset. This is an important information, I think.

lenkis:
Already in https://github.com/tarantool/cartridge/blob/master/rst/cartridge_dev.rst

PR #1506, opened 2020-09-08
gh-986 clarify info about cluster roles + translation

Onvember:


lenkis:
Already integrated into https://github.com/tarantool/cartridge/blob/master/rst/cartridge_dev.rst

Issue #1419, opened 2020-07-22
feedback: Module fiber


TarantoolBot:
> <…>uch as pthread_cond_wait() plus pthread_cond_signal().
> 
> 
> 
> Index
> **|Below is a list of all fiber functions and members.|**
> 
> 
> 
> 
> 
> 
> 
> Name
> Use
> 
> 
> 
> fiber.create()
> Create and start a fiber
> 
> fib<…>

https://www.tarantool.io/en/doc/2.3/reference/reference_lua/fiber/

The government conceals information about `fiber.clock()` function! It exists, but isn't mentioned in the documentation.

Let's reveal this truth.

lenkis:
`fiber.clock()` arrived since 1.7.5-3-gd988d7fb9. But we don't support 1.7 doc any longer, so please start with 1.10 doc.

PR #1553, opened 2020-09-22
gh-1419 describe fiber.clock() + translation

Onvember:


doc/reference/reference_lua/fiber.rst
```diff
@@ -779,6 +785,46 @@ recommended.
             - 1448466351270762
             ...
 
+.. _fiber-clock:
+
+.. function:: clock()
+
+    Get the monotonic time in seconds. It is better to use ``fiber.clock()`` for
+    calculating timeouts instead of ``fiber.time()`` because ``fiber.time()``
```
lenkis, line 793: maybe add a link here?

...instead of :ref:`fiber.time() <fiber-time>` because ...

doc/reference/reference_lua/fiber.rst
```diff
@@ -779,6 +785,46 @@ recommended.
...
+        tarantool> print(fiber.time(), fiber.time()-start)
+        1600785979.8291 1600537279.241
+        ---
+        ...
+
+.. _fiber-clock64:
+
+.. function:: clock64()
+
+    Same as ``fiber.clock()`` but in microseconds.
```
lenkis, line 821:
```suggestion
    Same as :ref:`fiber.clock() <fiber-clock>` but in microseconds.
```

doc/reference/reference_lua/fiber.rst
```diff
@@ -779,6 +785,46 @@ recommended.
...
+.. _fiber-clock:
+
+.. function:: clock()
+
+    Get the monotonic time in seconds. It is better to use ``fiber.clock()`` for
+    calculating timeouts instead of ``fiber.time()`` because ``fiber.time()``
+    reports real time so it is affected by system time changes.
+
+    :Return: a floating-point number of seconds, representing elapsed wall-clock
+             time since some time in the past that is guaranteed not to change
```
lenkis, line 797: since some time in the past = since 01.01.1970? (the Unix epoch is 00:00:00 UTC on 1 January 1970, so most probably here the same starting point is used)

lenkis (changes requested):

doc/reference/reference_lua/fiber.rst
olegrok: No it's not a Unix epoch and "some time in the past" is correct phrase.

It was inspired by [MPI Wtime](https://mpi.deino.net/mpi_functions/MPI_Wtime.html) or [OpeMP get_wtime](https://www.openmp.org/spec-html/5.0/openmpsu160.html)

> a value equal to the elapsed wall clock time in seconds since some time-in-the-past. The actual time-in-the-past is arbitrary, but it is guaranteed not to change during the execution of the application program.

lenkis: Thanks for explaining. Let's leave the text as it is -- it's quite enough for users.

doc/reference/reference_lua/fiber.rst
Onvember: fixed in 1c92371

doc/reference/reference_lua/fiber.rst
Onvember: fixed in 1c92371

PR #1599, opened 2020-10-13
Update cookbook.rst

aak74:
changes to fit version 2+
see: https://github.com/tarantool/http/issues/97


Onvember:
thank you for the update! I used it in https://github.com/tarantool/doc/pull/1618

PR #1634, opened 2020-11-06
Add transactional manager and memtx_use_mvcc_engine option description

veod32:


doc/book/box/atomic.rst
```diff
@@ -244,3 +244,27 @@ via telnet, via one of the connectors, or via the
...
+
+--------------------------------------------------------------------------------
+Transactional manager
+--------------------------------------------------------------------------------
+
+Since version 2.6.1, Tarantool has another option for transaction behaviour that
+allows yielding inside a memtx transaction. This is controled by
+the *transactional manager*.
+
+The transactional manager is designed for isolation of concurrent transactions.
```
alyapunov, line 258: I would add "and provides serializable transaction isolation level".

doc/book/box/atomic.rst
```diff
@@ -244,3 +244,27 @@ via telnet, via one of the connectors, or via the
...
+It consists of two parts:
+
+* *MVCC engine*
+* *conflict manager*.
+
+The MVCC engine provides personal read views for transactions if necessary.
+The conflict manager tracks transactions' changes and determines their correctness
+in serialization order. Of course, once yielded, a transaction could interfere
+with other transactions and could be aborted due to conflict.
```
alyapunov, line 267: Another thing that worth to be mentioned is that the transaction manager provides non classic snapshot isolation level. That means that a transaction can get a consistent snapshot of DB (that's common) but that snapshot is not necessary bound with the moment of the beginning of the transaction (that's not common). Conflict manager makes decisions of whether and when each transaction gets which snapshot. That allows to avoid some conflicts comparing with classical snapshot isolation approach.

doc/book/box/atomic.rst
veod32: added

doc/book/box/atomic.rst
veod32: Added this info

PR #1680, opened 2020-12-07
Replace hyphen and n-dash with m-dashes.

### LINKED ISSUE #1679: feedback: Создаем свою первую базу данных на Tarantool | Tarantool
> <…>ое уже установлено и настроено на платформе Linux. Данные **образы| -|** это самый простой способ установить и запустить Tarantool.

https://www.tarantool.io/ru/doc/latest/getting_started/getting_started_db/

тут должно быть длинное тире (m-dash)

NickVolynkin:


veod32:
Actually, it's a rendering bug: we have an agreement that we put two hyphens for a dash in the source RST files and they should be replaced with the m-dash during rendering. So, better to fix the rendering bug to resolve the issue for the entire documentation in both English and Russian.

NickVolynkin:
If a double hyphen makes an m-dash, what makes an n-dash?

veod32:
I don't remember we've ever used n-dash in English text. So, if you can bring in the n-dash use cases, we can reconsider the rules and the process. As for Russian text, we thought of following the corporate voice-and-tone https://paradigm.mail.ru/voice_and_tone/text_sign, but it hasn't been put into practice yet. Anyway, it's an area of improvement, so if you have suggestions, we can introduce some changes. I think it's better to do this for the entire documentation scope by introducing changes in
1. Style guide
2. Rendering RST into HTML and other output formats

PR #1692, opened 2020-12-14
gh-1031 refactor C Style Guide

### LINKED ISSUE #1031: [3pt] Cleanup Tarantool C Style Guide
https://www.tarantool.io/en/doc/1.10/dev_guide/c_style_guide/

The following characters look inapplicable in context of tarantool:

* Chapter 10: Kconfig configuration files
* Chapter 13: Printing kernel messages
* Chapter 14: Allocating memory

Chapters that should be adjusted:

* Chapter 17: Don’t re-invent the kernel macros
  - Update it with macros from `src/trivia/util.h`.

[veod32]: The issue estimation of 48pt is subject to changed if we understand the scope of work is bigger than the current one above.

Onvember:
The current page about C style is quite old and confusing so I've tried to make it better. 

Tarantool style features will be rendered like this:
image

**!!!NOTE FOR REVIEWERS!!!**
The "c_style_guide.rst" and "c_style_guide_new.rst" files are identical. I added a "c_style_guide_new.rst" just to make sure that git comparisons with the old version would not get in the way during the review. I will remove extra files after review. Feel free to comment on any of these files.

fixes #1031


doc/dev_guide/c_style_guide.rst
```diff
@@ -2,124 +2,36 @@
...
-and unloaded multiple times, reload should not lead to a memory leak.
-
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-         Function naming
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-Our convention is to use:
+`developer discussion list <https://groups.google.com/forum/#!forum/tarantool>`_
+or on `StackOverflow <https://stackoverflow.com/questions/tagged/tarantool>`_.
+Additionally, some engineers are always present on #tarantool channel on
```
Gerold103, line 13: I know that it is not practiced much in the doc team, but usually we try to make changes atomic. Here I see changes not only about C style, which makes it hard to review the C style updates. The diff is simply too big.
If the doc team does not care about atomicity, then nevermind. Otherwise I would split these non-style changes into a different commit.

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,1046 @@
...
+We use Git for revision control. The latest development is happening in the
+default branch (currently ``master``). Our git repository is hosted on GitHub,
+and can be checked out with ``git clone git://github.com/tarantool/tarantool.git``
+(anonymous read-only access).
+
+If you have any questions about Tarantool internals, please post them on the
+`developer discussion list <https://groups.google.com/forum/#!forum/tarantool>`_
+or on `StackOverflow <https://stackoverflow.com/questions/tagged/tarantool>`_.
+Additionally, some engineers are always present on #tarantool channel on
+irc.freenode.net.
```
Gerold103, line 14: If we don't care about atomicity of the changes, I would delete the google group and irc.freenode.net mentioning. The latter does not even work.

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,1046 @@
...
+
+================================================================================
+                           Linux kernel coding style
+================================================================================
+
+This is a short document describing the preferred coding style for the
+linux kernel. Coding style is very personal, and I won't **force** my
+views on anybody, but this is what goes for anything that I have to be
+able to maintain, and I'd prefer it for most other things too.  Please
+at least consider the points made here.
```
Gerold103, line 37: This paragraph could make a user think he has a choice not to use this guide, which is not true. We exactly **force** our style.

I guess this is a direct cite from the kernel guide, which adds +1 point to the option of rewriting the guide entirely.

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,1046 @@
...
+.. admonition:: Tarantool Style
+    :class: FACT
+
+    As for comments, the same limit of 80 columns is applied.
+
+Statements longer than 80 columns will be broken into sensible chunks, unless
+exceeding 80 columns significantly increases readability and does not hide
+information. Descendants are always substantially shorter than the parent and
+are placed substantially to the right. The same applies to function headers
+with a long argument list. However, never break user-visible strings such as
```
Gerold103, line 126: We allow to break strings.

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,1046 @@
...
+25-line terminal screens here), you have more empty lines to put
+comments on.
+
+Do not unnecessarily use braces where a single statement will do.
+
+.. code-block:: c
+
+    if (condition)
+      action();
+    and
```
Gerold103, line 214: "and" is not supposed to be a part of the code block.

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,1046 @@
...
+      do_that();
+    } else {
+      otherwise();
+    }
+
+**********************
+Chapter 3.1: Spaces
+**********************
+
+Tarantool style for use of spaces depends (mostly) on
```
Gerold103, line 239: Does not look like a direct cite. And is not different from Linux. Should we even change anything here? I am talking about using "Tarantool" word in the kernel style guide copy-paste.

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,1046 @@
...
+However, some such editors do not remove the whitespace if you end up not
+putting a line of code there, such as if you leave a blank line. As a result,
+you end up with lines containing trailing whitespace.
+
+Git will warn you about patches that introduce trailing whitespace, and can
+optionally strip the trailing whitespace for you; however, if applying a series
+of patches, this may make later patches in the series fail by changing their
+context lines.
+
+.. admonition:: Tarantool Style
```
Gerold103, line 306: This note better be moved a little bit above, to the text talking about unary operators (because this is one).

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,1046 @@
...
+
+   When editing existing code which already uses one or the other set
+   of types, you should conform to the existing choices in that code.
+
+#. Types safe for use in userspace.
+
+   In certain structures which are visible to userspace, we cannot
+   require C99 types and cannot use the ``u32`` form above. Thus, we
+   use __u32 and similar types in all structures which are shared
+   with userspace.
```
Gerold103, line 431: Another +1 point to the option of writing our own style guide - this useless information. We are userspace, so this can't be applied to Tarantool. There are many places like this.

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,1046 @@
...
+closing function brace line.  E.g.:
+
+.. code-block:: c
+
+  int
+  system_is_up(void)
+  {
+    return system_state == SYSTEM_RUNNING;
+  }
+  EXPORT_SYMBOL(system_is_up);
```
Gerold103, line 479: Another unnecessary info from Linux guide - we don't have `EXPORT`.

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,1046 @@
...
+    Javadoc flavor, i.e. `@tag` rather than `\\tag`.
+    The main tags in use are ``@param``, ``@retval``, ``@return``, ``@see``,
+    ``@note`` and ``@todo``.
+
+    Every function, except perhaps a very short and obvious one, should have a
+    comment. A sample function comment may look like below:
+
+    .. code-block:: c
+
+        /** Write all data to a descriptor.
```
Gerold103, line 595: We don't write text on the same line as the comment opening (`/**`). It should be empty.

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,1046 @@
...
+         *
+         * This function is equivalent to 'write', except it would ensure
+         * that all data is written to the file unless a non-ignorable
+         * error occurs.
+         *
+         * @retval 0  Success
+         *
```
Gerold103, line 602: Don't need to put empty lines between doxygen commands. At least if they are short.

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,1046 @@
...
+         * @retval  1  An error occurred (not EINTR)
+         * /
```
Gerold103, line 604: It is not a valid C comment closure - should be `*/`, not `* /`.

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,1046 @@
...
+        static int
+        write_all(int fd, void \*data, size_t len);
+
+    It's also important to comment data, whether they are basic types or derived
```
Gerold103, line 608: data -> data types.

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,1046 @@
...
+    types. To this end, use just one data declaration per line (no commas for
+    multiple data declarations). This leaves you room for a small comment on each
+    item, explaining its use.
```
Gerold103, line 611: I couldn't parse the last 2 sentences.

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,1046 @@
...
+                ;; Enable kernel mode for the appropriate files
+                (when (and filename
+                           (string-match (expand-file-name "~/src/linux-trees")
+                                         filename))
+                  (setq indent-tabs-mode t)
+                  (setq show-trailing-whitespace t)
+                  (c-set-style "linux-tabs-only")))))
+
+This will make emacs go better with the kernel coding style for C
+files below ``~/src/linux-trees``.
```
Gerold103, line 694: This is not related to Tarantool. It assumes you have a special file with rules, and even if you really have it - it contains Linux rules, without our corrections. Another point why citing the kernel guide is a bad idea.

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,1046 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Chapter 11: Allocating memory
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. admonition:: Tarantool Style
+    :class: FACT
+
+    Prefer the supplied slab (salloc) and pool (palloc) allocators to malloc()/free()
+    for any performance-intensive or large  memory allocations. Repetitive use of
+    malloc()/free() can lead to memory fragmentation and should therefore be avoided.
```
Gerold103, line 801: We don't use `salloc` nor `palloc`. Also `malloc` is not a big deal when is used in a non-perf critical code. For perf critical code there are various options and hacks, not only slab allocator.

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,1046 @@
...
+
+Mixing up these two sorts of representations is a fertile source of
+difficult-to-find bugs.  If the C language included a strong distinction
+between integers and booleans then the compiler would find these mistakes
+for us... but it doesn't.  To help prevent such bugs, always follow this
+convention:
+
+  If the name of a function is an action or an imperative command,
+  the function should return an error-code integer.  If the name
+  is a predicate, the function should return a "succeeded" boolean.
```
Gerold103, line 855: Not correct for Tarantool. In 99.99999% of all cases we return 0 on success, non-zero on error (-1 usually). Errors are saved into a diagnostics area which is global per fiber. We never return error codes as a result of a function.

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,1046 @@
...
+remain easy to follow.
+
+Prefer to compile out entire functions, rather than portions of functions or
+portions of expressions.  Rather than putting an ifdef in an expression, factor
+out part or all of the expression into a separate helper function and apply the
+conditional to that function.
+
+If you have a function or variable which may potentially go unused in a
+particular configuration, and the compiler would warn about its definition
+going unused, mark the definition as __maybe_unused rather than wrapping it in
```
Gerold103, line 957: We don't have `__maybe_unused`. If a function may not be used, we usually not compile it and use #if for this.

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,1046 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. admonition:: Tarantool Style
+    :class: FACT
+
+    Use header guards. Put the header guard in the first line in the header,
+    before the copyright or declarations. Use all-uppercase name for the header
+    guard. Derive the header guard name from the file name, and append _INCLUDED
+    to get a macro name. For example, core/log_io.h -> CORE_LOG_IO_H_INCLUDED. In
+    ``.c`` (implementation) file, include the respective declaration header before
```
Gerold103, line 999: We use `#pragma once` in the new code instead of macro guards.

doc/dev_guide/c_style_guide_new.rst
Onvember: Can I just remove this chapter or the part about `indent` below is ok?

doc/dev_guide/c_style_guide_new.rst
Onvember: It [came](https://www.tarantool.io/en/doc/latest/dev_guide/c_style_guide/#allocating-memory) from the previous version of this guide. Should I remove this statement at all?

doc/dev_guide/c_style_guide_new.rst
Onvember: Should I rewrite just the convention according to your comment, or the whole chapter is misleading?

doc/dev_guide/c_style_guide_new.rst
Onvember: I need more details to rewrite this part

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,1013 @@
...
+Prefer to compile out entire functions, rather than portions of functions or
+portions of expressions. Rather than putting an ifdef in an expression, factor
+out part or all of the expression into a separate helper function and apply the
+conditional to that function.
+
+If you have a function or variable which may potentially go unused in a
+particular configuration, and the compiler would warn about its definition
+going unused, mark the definition as __maybe_unused rather than wrapping it in
+a preprocessor conditional. (However, if a function or variable *always* goes
+unused, delete it.)
```
Onvember, line 926: Will this be correct?

```suggestion
going unused, do not compile it and use #if for this.
(However, if a function or variable *always* goes unused, delete it.)
```

doc/dev_guide/c_style_guide_new.rst
Onvember: Will this be correct?

```suggestion
going unused, do not compile it and use #if for this.
(However, if a function or variable *always* goes unused, delete it.)
```

doc/dev_guide/c_style_guide_new.rst
Gerold103: Looks irrelevant as well.

doc/dev_guide/c_style_guide_new.rst
Gerold103: I would say something like "prefer specialized allocators like region, mempool, smalloc" instead of "salloc/palloc". For perf critical data, which is almost everything with rare exceptions. "salloc/palloc" didn't ever exist in our code, so the previous version of the guide was also wrong.

doc/dev_guide/c_style_guide_new.rst
Gerold103: Only the last paragraph in the chapter looks kind of relevant. The rest is misleading.

doc/dev_guide/c_style_guide_new.rst
Gerold103: Yeah, that is better. We also have C++ templates. I guess it might be a separate task to describe when it is appropriate to use them.

doc/dev_guide/c_style_guide_new.rst
Gerold103: It is that simple - just use `#pragma once` in the headers. If we need to add some water, then we can say that as the header guards we refer to this construction:
```C
#ifndef THE_HEADER_IS_INCLUDED
#define THE_HEADER_IS_INCLUDED

// ... the header code ...

#endif // THE_HEADER_IS_INCLUDED
```
It works fine, but the guard name `THE_HEADER_IS_INCLUDED` tends to outdate when the file is moved or renamed. This is especially painful with multiple files having the same name in the project, but different path. For instance, we have 3 `error.h` files, which means for each of them we need to invent a new header guard name, and not forget to update them if the files are moved or renamed.

For that reason we use `#pragma once` in all the new code, which shortens the header file down to this:
```C
#pragma once

// ... header code ...
```

(Header guard and pragma once - both prevent inclusion of the same file multiple times in C when `#include "<my_file_name>.h"` is used.)

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,920 @@
...
+This is a short document describing the preferred coding style for the
+Tarantool developers and contributors. We **insist** on following these rules
+in order to make our code consistent and understandable to any developer.
+
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Chapter 1: Indentation
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Tabs are 8 characters, and thus indentations are
+also 8 characters. There are heretic movements that try to make indentations
```
Gerold103, line 36: It is worth mentioning that we use 8-width tabs, not 8 whitespaces.

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,920 @@
...
+
+Don't put multiple statements on a single line unless you have
+something to hide:
+
+..  code-block:: c
+
+    if (condition) do_this;
+      do_something_everytime;
+
+Don't put multiple assignments on a single line either. Kernel coding style
```
Gerold103, line 86: It is not Kernel coding style anymore. We must purge all mentioning of the kernel from the doc.

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,920 @@
...
+..  code-block:: c
+
+    struct virtual_container *a;
+
+you can actually tell what ``a`` is.
+
+Lots of people think that typedefs ``help readability``. Not so. They are
+useful only for:
+
+#.  totally opaque objects (where the typedef is actively used to **hide**
```
Gerold103, line 356: `Totally` should start with a capital letter?

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,920 @@
...
+
+        Again - there needs to be a **reason** for this. If something is
+        ``unsigned long``, then there's no reason to do
+        typedef unsigned long myflags_t;
+
+    but if there is a clear reason for why it under certain circumstances
+    might be an ``unsigned int`` and under other configurations might be
+    ``unsigned long``, then by all means go ahead and use a typedef.
+
+#.  when you use sparse to literally create a **new** type for
```
Gerold103, line 384: Ditto.

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,920 @@
...
+    type-checking.
+
+#.  New types which are identical to standard C99 types, in certain
+    exceptional circumstances.
+
+    Although it would only take a short amount of time for the eyes and
+    brain to become accustomed to the standard types like ``uint32_t``,
+    some people object to their use anyway.
+
+    Therefore, the Linux-specific ``u8/u16/u32/u64`` types and their
```
Gerold103, line 394: We don't use these types. They can only be found in old SQL code, but we are getting rid of them.

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,920 @@
...
+
+Another measure of the function is the number of local variables. They
+shouldn't exceed 5-10, or you're doing something wrong. Re-think the
+function, and split it into smaller pieces. A human brain can
+generally easily keep track of about 7 different things, anything more
+and it gets confused. You know you're brilliant, but maybe you'd like
+to understand what you did 2 weeks from now.
+
+In function prototypes, include parameter names with their data types.
+Although this is not required by the C language, it is preferred in Linux
```
Gerold103, line 438: "in Linux" -> "in Tarantool". Might be good to find all the usages of word "Linux" and replace with "Tarantool" where appropriate.

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,920 @@
...
+
+    /**
+     * Write all data to a descriptor.
+     *
+     * This function is equivalent to 'write', except it would ensure
+     * that all data is written to the file unless a non-ignorable
+     * error occurs.
+     *
+     * @retval 0  Success
+     * @retval  1  An error occurred (not EINTR)
```
Gerold103, line 552: Excess whitespaces around `1`.

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,920 @@
...
+     */
+    static int
+    write_all(int fd, void \*data, size_t len);
```
Gerold103, line 555: Should not have `\` before `*`.

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,920 @@
...
+          typeof(x) ret;          \
+          ret = calc_ret(x);      \
+          (ret);                  \
+        })
+
+    ret is a common name for a local variable - ``__foo_ret`` is less likely
+    to collide with an existing variable.
+
+    The cpp manual deals with macros exhaustively. The gcc internals manual also
+    covers RTL which is used frequently with assembly language in the kernel.
```
Gerold103, line 675: Should not mention the kernel probably.

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,920 @@
...
+available for the pagecache. Just think about it; a pagecache miss causes a
+disk seek, which easily takes 5 milliseconds. There are a LOT of cpu cycles
+that can go into these 5 milliseconds.
+
+A reasonable rule of thumb is to not put inline at functions that have more
+than 3 lines of code in them. An exception to this rule are the cases where
+a parameter is known to be a compiletime constant, and as a result of this
+constantness you *know* the compiler will be able to optimize most of your
+function away at compile time. For a good example of this later case, see
+the kmalloc() inline function.
```
Gerold103, line 710: `kmalloc()` is a kernel function, we should not mention it.

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,920 @@
...
+
+In 99.99999% of all cases in Tarantool we return 0 on success, non-zero on error
+(-1 usually). Errors are saved into a diagnostics area which is global per fiber.
+We never return error codes as a result of a function.
+
+Functions whose return value is the actual result of a computation, rather
+than an indication of whether the computation succeeded, are not subject to
+this rule. Generally they indicate failure by returning some out-of-range
+result. Typical examples would be functions that return pointers; they use
+NULL or the ERR_PTR mechanism to report failure.
```
Gerold103, line 735: We don't have `ERR_PTR`, only `NULL`.

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,920 @@
...
+
+Do not include any of these in source files. People have their own personal
+editor configurations, and your source files should not override them. This
+includes markers for indentation and mode configuration. People may use their
+own custom mode, or may have some other magic method for making indentation
+work correctly.
+
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Chapter 14: Inline assembly
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
Gerold103, line 773: Tbh, I would drop chapter this entirely. We don't write in assembly except extremely rare exceptions which I can't even remember what was the last one.

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,920 @@
...
+use such conditionals in a header file defining functions for use in those .c
+files, providing no-op stub versions in the #else case, and then call those
+functions unconditionally from .c files. The compiler will avoid generating
+any code for the stub calls, producing identical results, but the logic will
+remain easy to follow.
+
+Prefer to compile out entire functions, rather than portions of functions or
+portions of expressions. Rather than putting an ``#ifdef`` in an expression,
+factor out part or all of the expression into a separate helper function and
+apply the conditional to that function.
```
Gerold103, line 818: Shouldn't it be "the conditional" -> "the condition"?

doc/dev_guide/c_style_guide_new.rst
```diff
@@ -0,0 +1,920 @@
...
+
+If you have a function or variable which may potentially go unused in a
+particular configuration, and the compiler would warn about its definition
+going unused, do not compile it and use #if for this.
+(However, if a function or variable *always* goes unused, delete it.)
+
+Within code, where possible, use the IS_ENABLED macro to convert a Kconfig
```
Gerold103, line 825: We don't have `IS_ENABLED` nor `Kconfig`. I would propose to drop the text from here to line 841.

NickVolynkin (approved):

PR #1707, opened 2020-12-17
gh-1685 split box submodules into separate pages for each function

Onvember:


Totktonada:
Don't forget about proper redirects (just as it was with #1711).

Issue #1664, opened 2020-11-27
[36pt] Provide an example to use yielding routines in finalizer


igormunkin:
Fiber switch is forbidden in `__gc` metamethod since [this change](https://github.com/tarantool/tarantool/issues/4518#issuecomment-704259323). However, one may need to use a yielding function to finalize the resources (e.g. to close the socket). We should provide the example with the proper way implementing such methods.

Since one can't use explicitly yielding function in `__gc` metamethod, this action should be scheduled to be guaranteed executed by the platform. @Gerold103 already filed [the ticket](https://github.com/tarantool/tarantool/issues/5544) regarding the Tarantool-wide scheduler for such lightweight tasks, __but__
1. This feature will be prioritized, designed, implemented and only then released, but one can't yield in scope of the finalizer already __now__
2. Even after being implemented this feature won't be backported to 1.10, but fiber switch in `__gc` is forbidden in LTS versions starting from [1.10.7-47-g8099cb053](https://github.com/tarantool/tarantool/commit/8099cb053)

---

Here is the short example to implement the valid finalizer for the particular FFI `<custom_t>` type.
* `simple.lua`
```lua
local ffi = require('ffi')
local fiber = require('fiber')

ffi.cdef('struct custom { int a; };')

local function __custom_gc(self)
  print(("Entered custom GC finalizer for %s... (before yield)"):format(self.a))
  fiber.yield()
  print(("Leaving custom GC finalizer for %s... (after yield)"):format(self.a))
end

local custom_t = ffi.metatype('struct custom', {
  __gc = function(self)
    -- XXX: Do not invoke yielding functions in __gc metamethod.
    -- Create a new fiber to be run after the execution leaves
    -- this routine.
    fiber.new(__custom_gc, self)
    print(("Finalization is scheduled for %s..."):format(self.a))
  end
})

-- Create a cdata object of <custom_t> type.
local c = custom_t(42)

-- Remove a single reference to that object.
c = nil

-- Run full GC cycle to purge the unreferenced object.
collectgarbage('collect')
-- > Finalization is scheduled for 42...

-- XXX: There is no finalization made until the running fiber
-- yields its execution. Let's do it now.
fiber.yield()
-- > Entered custom GC finalizer for 42... (before yield)
-- > Leaving custom GC finalizer for 42... (after yield)
```

---

Here is another simple example to implement the valid finalizer for the particular `<struct custom>` user type.
* `custom.c`
```c
#include <lauxlib.h>
[...]

PR #1708, opened 2020-12-17
gh-1664 Add usage examples of yielding routines in custom finalizer

veod32:


doc/book/admin/troubleshoot.rst
```diff
@@ -314,3 +314,379 @@ recommend to optimize your Tarantool application code).
...
+--------------------------------------------------------------------------------
+Problem: Fiber switch is forbidden in ``__gc`` metamethod
+--------------------------------------------------------------------------------
+
+~~~~~~~~~~~~~~~~~~~~~~~~
+Problem description
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+Fiber switch is forbidden in ``__gc`` metamethod since `this change <https://github.com/tarantool/tarantool/issues/4518#issuecomment-704259323>`_
+to avoid unexpected Lua OOM.
```
igormunkin, line 329: Well, I'm not sure, whether we need to describe this issue in a more precise way: on the one hand it looks making everything clearer for "experts"; on the other hand it might sound like some kind of elvish and reveals the details Tarantool user doesn't have to think about. Let's leave it as is for now.

igormunkin (approved):
@veod32, thanks! LGTM.

doc/book/admin/troubleshoot.rst
veod32: Ok. If we decide to describe the issue in more details, let's do it in a separate doc ticket.

PR #1783, opened 2021-01-28
Use Python 3.9 in Travis builds

NickVolynkin:


.travis.yml
```diff
@@ -1,4 +1,6 @@
-language: generic # don't install any environment
+language: python # don't install any environment
```
Totktonada, line 1: The comment becomes outdated.

NickVolynkin: @Totktonada yes, but Travis doesn't run the builds anyway, so I'm going to close this PR. We'll probably have to rebuild CI on GH Actions.

PR #1834, opened 2021-02-16
Discourage using i.e. and e.g.; fix #1833

### LINKED ISSUE #1833: Guidelines: don't use i.e. and e.g.
* Why it's bad practice
* What to use instead

NickVolynkin:


Onvember (approved):

doc/contributing/docs/style.rst
```diff
@@ -21,4 +21,19 @@ Wrong usage: "Replication allows multiple Tarantool *servers* to work on copies
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Don't use i.e. and e.g.
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Don't use the following contractions:
+
+*   `"i.e." <https://www.merriam-webster.com/dictionary/i.e.>`_
+    —from Latin "id est". Use "that is" or "which means" instead.
+*   `"e.g." <https://www.merriam-webster.com/dictionary/e.g.>`_
+    —from Latin "exempli gratia". Use "for example" instead.
```
Totktonada, line 35: Or 'such as'?

doc/contributing/docs/style.rst
```diff
@@ -21,4 +21,19 @@ Wrong usage: "Replication allows multiple Tarantool *servers* to work on copies
...
+
+Many people, especially non-native English speakers,
+aren't familiar with `these contractions
+<https://www.merriam-webster.com/words-at-play/ie-vs-eg-abbreviation-meaning-usage-difference>`_.
```
Totktonada, line 39: Correct usage should not be a problem for a reader I guess.

doc/contributing/docs/style.rst
veod32, line 35:
```suggestion
    —from Latin "exempli gratia". Use "for example" or "for instance" instead.
```

doc/contributing/docs/style.rst
NickVolynkin: People also misuse them, i.e. use "i.e." when they mean to say "for example". ;)

doc/contributing/docs/style.rst
NickVolynkin: This is a great advice in general, thank you. But I would object to using it in our docs because "instance" is a term that we use a lot. It's best if we reserve it for cases where it means a Tarantool instance.

doc/contributing/docs/style.rst
Totktonada: That's true (for me as well before I read links you pointed here). However I'm about reading. Is that a problem?

doc/contributing/docs/style.rst
Totktonada: Maybe it is more like 'spoken language', I don't know.

PR #1853, opened 2021-02-22
[2pt] Recommend writing in simple sentences

NickVolynkin:


doc/contributing/docs/style.rst
```diff
@@ -23,6 +23,44 @@ of the same databases."
...
+..  rst-class:: table-example
+..  container:: table
+
+    ..  rst-class:: left-align-column-1
+    ..  rst-class:: left-align-column-2
+
+    ..  list-table::
+        :header-rows: 1
+
+        *   *   Don't
```
Onvember, line 44: it more readable for me when rows and columns marked differently, for example
```suggestion
        *   -   Don't
```

NickVolynkin: That's a great idea, I like it.

PR #1855, opened 2021-02-22
[1pt] Set up a live-reload server with `make autobuild`

NickVolynkin:
When editing the documentation, you can set up a live-reload server.
It will build your documentation and serve it on http://127.0.0.1:8000,
Every time you save changes in the source files, it will rebuild the docs
and refresh the browser page.
```bash
docker run --rm -it -p 8000:8000 -v $(pwd):/doc tarantool-doc-builder sh -c "make autobuild"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@ Tarantool documentation source, published at https://www.tarantool.io/doc/.
 ## How to build Tarantool documentation using [Docker](https://www.docker.com)
 
 ### Build docker image
+
 ```bash
 docker build -t tarantool-doc-builder .
 ```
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -56,16 +56,28 @@ docker run --rm -it -v $(pwd):/doc tarantool-doc-builder sh -c "make update-po-f
...
 
 ### Run documentation locally on your machine
+
+When editing the documentation, you can set up a live-reload server.
+It will build your documentation and serve it on [127.0.0.1:8000](http://127.0.0.1:8000).
+Every time you make changes in the source files, it will rebuild the docs
+and refresh the browser page.
+
+```bash
+docker run --rm -it -p 8000:8000 -v $(pwd):/doc tarantool-doc-builder sh -c "make autobuild"
```
veod32, line 66: Everything works fine! The only suggestion is to restructure README: make it more clear that `make autobuild` is an alternative to `make html` and what step are necessary before `make autobuild`; and that one also needs to rebuild the docker image to make it work.

Issue #1757, opened 2021-01-18
[1pt] Python 2 is gone, need to update page "Building from source"


ligurio:
We are moving away from Python 2 to Python 3 (see https://github.com/tarantool/tarantool/issues/5652),
let's update text and package names in https://www.tarantool.io/en/doc/latest/dev_guide/building_from_source/

NickVolynkin:
@ligurio since which Tarantool versions is this change effective? Are they 2.5.3, 2.6.2, and 2.7.1?

ligurio:
@NickVolynkin to be honest we are still in progress with migration (relevant tasks are [1] and [2]).
I believe change will affect all supported Tarantool versions.

1. https://github.com/tarantool/tarantool/issues/5652
2. https://github.com/tarantool/test-run/pull/263

ligurio:
We are close to moving Tarantool testing to Python 3:

- [ ] tarantool/tarantool#5652 - to be merged next week
- [x] tarantool/test-run#263

Totktonada:
The transition to Python 3 is done.

PR #1879, opened 2021-03-01
Update instructions for building from source

### LINKED ISSUE #1640: [1pt] feedback: Building from source
> <…>n-msgpack python-yaml python-argparse python-six python-gevent$ **|yum install -y gcc gcc-c++ cmake make coreutils sed \
>       autoconf automake libtool zlib-devel \
>       readline-devel ncurses-devel openssl-devel \
>       libunwind-devel libicu-devel \
>       python python-pip python-setuptools python-devel \
>       python-msgpack python-yaml python-argparse python-six python-gevent|**
> 
> 
> 
> For CentOS 8, say:
> COPYyum install epel-release
> curl -s http<…>

https://www.tarantool.io/en/doc/2.2/dev_guide/building_from_source/

Please add "git" package to this snippet

### LINKED ISSUE #1702: [2pt] mopinion: mistake in building instruction
URL
https://www.tarantool.io/en/doc/latest/dev_guide/building_from_source/

'For CentOS 8, say...' fails on sed and coreutils

Onvember:
Fixes #1702
Fixes #1640

### DIFF
--- a/doc/dev_guide/building_from_source.rst
+++ b/doc/dev_guide/building_from_source.rst
@@ -38,11 +38,11 @@ the preferences can differ. But strategically the steps are always the same.
      option in step 5. For all platforms, this is ``python`` version 2.7+
      (but not 3.x). You need the following Python modules:
 
-       * `pyyaml <https://pypi.python.org/pypi/PyYAML>`_ version 3.10
-       * `argparse <https://pypi.python.org/pypi/argparse>`_ version 1.1
-       * `msgpack-python <https://pypi.python.org/pypi/msgpack-python>`_ version 0.4.6
-       * `gevent <https://pypi.python.org/pypi/gevent>`_ version 1.1.2
-       * `six <https://pypi.python.org/pypi/six>`_ version 1.8.0
+     * `pyyaml <https://pypi.python.org/pypi/PyYAML>`_ version 3.10
+     * `argparse <https://pypi.python.org/pypi/argparse>`_ version 1.1
+     * `msgpack-python <https://pypi.python.org/pypi/msgpack-python>`_ version 0.4.6
+     * `gevent <https://pypi.python.org/pypi/gevent>`_ version 1.1.2
+     * `six <https://pypi.python.org/pypi/six>`_ version 1.8.0
 
    To install all required dependencies, follow the instructions for your OS:
 
@@ -50,23 +50,23 @@ the preferences can differ. But strategically the steps are always the same.
 
      .. code-block:: console
 
-        $ apt install -y build-essential cmake make coreutils sed \
+        $ apt install -y build-essential git cmake make coreutils sed \
               autoconf automake libtool zlib1g-dev \
               libreadline-dev libncurses5-dev libssl-dev \
               libunwind-dev libicu-dev \
               python python-pip python-setuptools python-dev \
-              python-msgpack python-yaml python-argparse python-six python-gevent
+              python-msgpack python-yaml python-six python-gevent
 
    * For RHEL/CentOS (versions under 8)/Fedora, say:
 
      .. code-block:: console
 
-         $ yum install -y gcc gcc-c++ cmake make coreutils sed \
+         $ yum install -y gcc gcc-c++ git cmake make coreutils sed \
                autoconf automake libtool zlib-devel \
                readline-devel ncurses-devel openssl-devel \
                libunwind-devel libicu-devel \
                python python-pip python-setuptools python-devel \
-               python-msgpack python-yaml python-argparse python-six python-gevent
+               python-msgpack python-yaml python-six python-gevent
 
    * For CentOS 8, say:
 
@@ -74,7 +74,7 @@ the preferences can differ. But strategically the steps are always the same.
 
          $ yum install epel-release
          $ curl -s https://packagecloud.io/install/repositories/packpack/backports/script.rpm.sh | sudo bash
-         $ yum install -y gcc gcc-c++ cmake make coreutils sed \
+         $ yum install -y gcc gcc-c++ git cmake make coreutils sed \
                autoconf automake libtool zlib-devel \
                readline-devel ncurses-devel openssl-devel \
                libunwind-devel libicu-devel \
@@ -87,11 +87,11 @@ the preferences can differ. But strategically the steps are always the same.
 
      .. code-block:: console
 
-         $ brew install cmake make autoconf binutils zlib \
+         $ brew install cmake make git autoconf binutils zlib \
                 autoconf automake libtool \
                 readline ncurses openssl libunwind-headers icu4c \
                 && pip install python-daemon \
-                msgpack-python pyyaml configargparse six gevent
+                msgpack-python pyyaml six gevent
 
      .. NOTE::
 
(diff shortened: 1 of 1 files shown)


NickVolynkin:
@Totktonada I've got several concerns:

1. We don't freeze package versions here, so build environment isn't controlled. Someone can follow the instructions exactly and still face a problem. Moreover, we recommend exact versions for *some* packages in the instructions, but then don't repeat them in the code snippet.
2. Package lists and build instructions have no connection with our tests for the corresponding platforms. We will have to update instructions, and we will forget to do so.
3. Package lists and build instructions aren't in the form of code.

doc/dev_guide/building_from_source.rst
```diff
@@ -112,7 +112,7 @@ the preferences can differ. But strategically the steps are always the same.
                autoconf automake libtool \
                readline ncurses openssl libunwind icu \
                python27 py27-pip py27-setuptools py27-daemon \
-               py27-msgpack py27-yaml py27-argparse py27-six py27-gevent
+               py27-msgpack py27-yaml py27-six py27-gevent
```
NickVolynkin, line 115: Do we really use Python2? @Totktonada

Onvember: As I can see, related issues are still open so I decided lo leave it as is (https://github.com/tarantool/doc/issues/1757)

Totktonada: It is so for now. Should be resolved within two weeks I guess.

Totktonada:
> 1. We don't freeze package versions here, so build environment isn't controlled. Someone can follow the instructions exactly and still face a problem. Moreover, we recommend exact versions for some packages in the instructions, but then don't repeat them in the code snippet.

Packages are installed from the distribution specific repositories. Any binary distribution must ensure binary compatibility of packages within a distribution version. A user should not meet any problems related to packages version on a distribution we support.

I don't see frozen versions or recommendation to use it aside of PyYAML (the mentioned version is quite old anyway). I don't know why the instruction repeats `test-run/requirements.txt`. It should not.

Totktonada:
> 2. Package lists and build instructions have no connection with our tests for the corresponding platforms. We will have to update instructions, and we will forget to do so.

Do you propose to extract dependencies from `rpm/tarantool.spec` or `debian/control`? That's the good idea, but requires some research. Consider [this](https://github.com/packpack/packpack/blob/c1d29ef09c6a89a89358ef56d7eca12843490cc5/pack/rpm.mk#L164-L165) and [this](https://github.com/packpack/packpack/blob/c1d29ef09c6a89a89358ef56d7eca12843490cc5/pack/deb.mk#L133).

Totktonada:
> 3. Package lists and build instructions aren't in the form of code.

Is that about formatting on the web page?

doc/dev_guide/building_from_source.rst
```diff
@@ -38,43 +38,43 @@ the preferences can differ. But strategically the steps are always the same.
...
-       * `pyyaml <https://pypi.python.org/pypi/PyYAML>`_ version 3.10
-       * `argparse <https://pypi.python.org/pypi/argparse>`_ version 1.1
-       * `msgpack-python <https://pypi.python.org/pypi/msgpack-python>`_ version 0.4.6
-       * `gevent <https://pypi.python.org/pypi/gevent>`_ version 1.1.2
-       * `six <https://pypi.python.org/pypi/six>`_ version 1.8.0
+     * `pyyaml <https://pypi.python.org/pypi/PyYAML>`_ version 3.10
+     * `argparse <https://pypi.python.org/pypi/argparse>`_ version 1.1
+     * `msgpack-python <https://pypi.python.org/pypi/msgpack-python>`_ version 0.4.6
+     * `gevent <https://pypi.python.org/pypi/gevent>`_ version 1.1.2
+     * `six <https://pypi.python.org/pypi/six>`_ version 1.8.0
```
Totktonada, line 45: * argparse — it is built-in in Python 2.6+ AFAIR.
* msgpack-python is a submdoule of test-run, so it is not necessary to install it separately.
* All other versions seems obsoleted. And there is not any sense to list them, they're listed in `test-run/requirements.txt` in the format that pip understands. One who want to install it using a package manager can find related package names in `rpm/tarantol.spec` or `debian/control` (but I would left this exercise to a user, just mention where to find it).

doc/dev_guide/building_from_source.rst
```diff
@@ -38,43 +38,43 @@ the preferences can differ. But strategically the steps are always the same.
...
      .. code-block:: console
 
-         $ yum install -y gcc gcc-c++ cmake make coreutils sed \
+         $ yum install -y gcc gcc-c++ git cmake make coreutils sed \
                autoconf automake libtool zlib-devel \
                readline-devel ncurses-devel openssl-devel \
                libunwind-devel libicu-devel \
                python python-pip python-setuptools python-devel \
-               python-msgpack python-yaml python-argparse python-six python-gevent
+               python-msgpack python-yaml python-six python-gevent
```
Totktonada, line 69: That's is not so trivial considering that it is Python 2 packages on CentOS 7, but Python 3 ones on CentOS 8 / Fedora. I would just suggest to use `pip --user -r test-run/requirements.txt` or `pip -r test-run/requirements.txt` from a virtualenv environments. We can mention that the same can be installed with packages, but, to be honest, if one want, (s)he will do it anyway based on the requirements.txt content. So, let's stick to `pip` for simplicity.

Totktonada:
Please, also eliminate mentions of 2.2 branch. It is not supported for a long time and useless for a user in fact.

Totktonada:
> 5. Run the test suite.

This section is something really strance. You should just have all dependencies available, build tarantool and run `make test`. I don't know all those symlinks and PATH tweaks are here. Maybe they were required a long time ago (more than 4 years).

Totktonada:
> ``dpkg-buildpackage`` or ``rpmbuild``

To be honest, I'm not sure those tools will works seamlessly. We use packpack and it is the only reliable way to build tarantool package ATM.

I would just refer [this](https://github.com/tarantool/tarantool/wiki/Build-RPM-or-Deb-package-using-packpack) instruction.

Totktonada:
> $ # if you installed tarantool locally after build
> $ tarantool

It is either when we're about installing to some custom destdir and setting PATH (which is not actually necessary AFAIU). Or regarding installing directly to /usr or /usr/local bypassing a system package manager, which is not what I would suggest for anyone.

Totktonada:
> * ``Release`` -- used only if the highest performance is required

We don't deploy and don't verify such packages. And the point regarding better performance comparing to `RelWithDebInfo` is dubious. So I would remove mention of the `Release` build type.

Totktonada:
> * `OpenSSL <https://www.openssl.org>`_ library, version 1.0.1+

There are no problems with 1.1.* known to me.

Totktonada:
> > $ # if you installed tarantool locally after build
> > $ tarantool
> 
> It is either when we're about installing to some custom destdir and setting PATH (which is not actually necessary AFAIU). Or regarding installing directly to /usr or /usr/local bypassing a system package manager, which is not what I would suggest for anyone.

However, okay, if it is /usr/local, it is more or less okay. Still requires root priveleges and dubious suggestion. Debatable topic.

Onvember:
Outdated. See https://github.com/tarantool/doc/pull/1989

PR #1913, opened 2021-03-11
Include content on C++ connector API in the Connectors chapter

veod32:


doc/book/connectors/index.rst
```diff
@@ -105,25 +117,56 @@ script:
...
 
+.. _index_connector_csharp:
+
 .. include:: __csharp.rst
 
+.. _index_connector_c:
+
 .. include:: __c.rst
 
+.. _index_connector_cxx:
```
NickVolynkin, line 160: This is madness, really.
Let's make a child document rather than extending this document even more.

NickVolynkin (approved):

Issue #1845, opened 2021-02-19
[3pt] Convert Readme.md into rST, include in the Contribution guide/docs


NickVolynkin:
Time to add Readme in the docs.

- [x] Convert Readme.md into Readme.rst. Use pandoc:

    ```bash
    brew install pandoc
    pandoc -f markdown -t rst -o readme.rst readme.md
    ```
   Put the same content instead of readme.rst. Try to keep most lines unchanged.

- [x] Include the readme in a dummy document under under https://www.tarantool.io/en/doc/latest/contributing/docs/. Include only the content after the first heading.
- [x] If including directly from readme.rst doesn't work, copy it during build or make a symbolic link

PR #1938, opened 2021-03-15
Convert readme.md into .rst and include it in the docs

xuniq:
fixes gh-1845


NickVolynkin (changes requested):
Great job! But there should be one document, otherwise it'll be hard to update them.

- [ ] Put the same content instead of readme.rst. Try to keep most lines unchanged.
- [ ] Include the readme in a dummy document under contributing/docs. Include only content after the first heading.
- [ ] If including directly from readme.rst doesn't work, copy it during build

NickVolynkin (approved):

Issue #1912, opened 2021-03-11
[8pt] Push any branch to the test server on demand


artembo:
It's quite challenging to check how a documentation version from a branch looks like.

Need to create github actions workflow which will upload documentation to some s3 bucket and trigger downloading it by the test server.

NickVolynkin:
Done!

PR #1946, opened 2021-03-16
Deploy documentation to the dev server

artembo:
#1912

### DIFF
--- a/.github/workflows/deploy-branch.yml
+++ b/.github/workflows/deploy-branch.yml
@@ -1,0 +1,49 @@
+name: Deploy-branch
+
+on:
+  pull_request:
+    types:
+      - labeled
+jobs:
+  deploy-branch:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.label.name == 'deploy-branch' }}
+    container: tarantool/doc-builder:v1
+    env:
+      AWS_ACCESS_KEY_ID: ${{secrets.AWS_ACCESS_KEY_ID}}
+      AWS_SECRET_ACCESS_KEY: ${{secrets.AWS_SECRET_ACCESS_KEY}}
+      AWS_DEFAULT_REGION: ${{secrets.AWS_DEFAULT_REGION}}
+      S3_ENDPOINT_URL: ${{secrets.S3_ENDPOINT_URL}}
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-branch.yml
```diff
@@ -0,0 +1,46 @@
+name: Deploy-dev
```
NickVolynkin, line 1: let's name it `deploy-branch`, just like in the filename. 
Otherwise the page https://github.com/tarantool/doc/actions/workflows/deploy-branch.yml is a little confusing.

.github/workflows/deploy-branch.yml
```diff
@@ -0,0 +1,46 @@
...
+
+on:
+  pull_request:
```
NickVolynkin, line 4: Shouldn't work on each PR, but on manual dispatch instead.
```suggestion
  workflow_dispatch:
```

.github/workflows/deploy-branch.yml
```diff
@@ -0,0 +1,46 @@
...
+
+      - run: cmake .
+      - run: make pull-modules
+      - run: make build-modules
+      - run: make json
+      - run: make json-ru
+
+      - name: Set branch name from source branch
+        run: echo "BRANCH_NAME=${GITHUB_HEAD_REF#refs/heads/}" >> $GITHUB_ENV
+      - run: bash upload_output.sh
```
NickVolynkin, line 37: Will it remove the files from S3 when this deployment isn't needed anymore? How do we remove deployments?

NickVolynkin (changes requested):
It's a kind of magic!

.github/workflows/deploy-branch.yml
artembo: Fixed

.github/workflows/deploy-branch.yml
artembo: No, will create a new workflow for it later

.github/workflows/deploy-branch.yml
NickVolynkin: Is it?

NickVolynkin (approved):

Issue #1880, opened 2021-03-01
[1pt] rST reference: use *, then - in list-tables


NickVolynkin:
# rST reference: use *, then - in list-tables

Add the [recommendation about using * and - in list-tables](https://github.com/tarantool/doc/pull/1853#discussion_r582070266)
to the rST reference.

**Reviewer:** @NickVolynkin 

# DoD

- [ ] explain why it's good
- [ ] give an example

NickVolynkin:
@Onvember should we maybe apply the same rule to nested lists?

PR #2041, opened 2021-04-09
fixes gh-1880 recommend using *, then - in list-tables

Onvember:


doc/contributing/docs/markup/tables.rst
```diff
@@ -5,15 +5,15 @@ Tables are very useful and rST markup
...
-We prefer list-tables to create table of contents:
+We prefer list-tables because they allow you to put as much content as you need
+without painting ASCII-art borders:
 
-..  code-block:: rst
+..  literalinclude:: _includes/table.rst
+    :language: rst
 
-    ..  container:: table
+And this table will appear like this:
```
ValeriaKhokha, line 14: The idea this sentence expresses is very clear, yet I would suggest to rephrase it in order to avoid double "this", for instance "_This is how the table will look like:_".

doc/contributing/docs/markup/tables.rst
```diff
@@ -5,15 +5,15 @@ Tables are very useful and rST markup
...
 
-        ..  list-table::
-            :widths: 25 75
-            :header-rows: 1
+..  include:: _includes/table.rst
 
-            *   - Name
-                - Use
+Notice that we use ``*``, then ``-`` in tables because it is more readable
```
veod32, line 18:
```suggestion
Notice that we use ``*`` and then ``-`` in tables because it is more readable
```

doc/contributing/docs/markup/intro.rst
```diff
@@ -11,11 +11,11 @@ Text can be organized in bullet-lists:
...
 ..  code-block:: rst
 
-    - This is a bullet list.
+    *   This is a bullet list.
 
-    - Bullets can be "*", "+", or "-".
+    *   Bullets can be "*", "+", or "-".
 
-        * Lists can be nested. And it is good to indent them with 4 spaces.
+        -   Lists can be nested. And it is good to indent them with 4 spaces.
```
veod32, line 18: As for 4 space indent, I would suggest changing the wording and making it "a must" instead of "nice to have" as it says now. Because we have this in our doc guidelines https://www.tarantool.io/en/doc/latest/contributing/docs/markup/intro/#indentation

Onvember: I don't think it is necessary here)

Issue #2059, opened 2021-04-13
[3pt] implement ability to run multiple iproto threads


TarantoolBot:
Page: https://www.tarantool.io/en/doc/latest/dev_guide/internals/box_protocol/ ?

Implement ability to run multiple iproto threads, which is useful
in some specific workloads where iproto thread is the bottleneck
of throughput. To specify count of iproto threads, user should used
iproto_threads option in box.cfg. For example if user want to start
8 iproto threads, he must enter `box.cfg{iproto_threads=8}`. Default
iproto threads count == 1. This option is not dynamic, so user can't
change it after first setting, until server restart. Distribution of
connections per threads is managed by OS kernel.
Requested by @EvgenyMekhanik in https://github.com/tarantool/tarantool/commit/2ede3be3564f04ba2c80ed454a2ba6b0ca11450d.

NickVolynkin:
@pgulutzan would you like to take this issue?

pgulutzan:
@NickVolynkin: Yes, I will self-assign.

@EvgenyMekhanik  via TarantoolBot is suggesting a change of the "binary protocol" page
"Page: https://www.tarantool.io/en/doc/latest/dev_guide/internals/box_protocol/ ?"
bit it does not seem relevant, because I don't see that there's a change to message content.

Here is a terminology issue. In the manual the term "iproto thread" is not the preferred term.
The manual describes three threads: network thread, transaction thread
(also called transaction processor thread), WAL thread (also called
write-ahead log thread). So should we stop saying "network thread"
and say "iproto thread" instead?

Say no because:
(1) We have called it a "network thread" for years, not merely in the
manual but in talks.
(2) The word network might be a word that people have heard before.

Say yes because:
(1) When you use ps you see the thread name iproto, for example:
$ ps -T -p 15147
    PID    SPID TTY          TIME CMD
  15147   15147 pts/1    00:00:00 tarantool
  15147   15149 pts/1    00:00:00 iproto
  15147   15150 pts/1    00:00:00 wal
  15147   15151 pts/1    00:00:00 coio
(2) When you use box.cfg{} the new option name is iproto_threads.

I will assume that you say no. Therefore the pages to change are:

Submodule box.cfg
https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_cfg/
... add an item in the list

Configuration reference, "Networking" section
https://www.tarantool.io/en/doc/latest/reference/configuration/#networking
... Most of the suggested addition goes here

[...]

PR #2094, opened 2021-04-22
Fixes gh-2059 implement ability to run multiple iproto threads

pgulutzan:


NickVolynkin:
@EvgenyMekhanik please have a look.

NickVolynkin (approved):

doc/reference/configuration/cfg_networking.rst
```diff
@@ -69,3 +69,21 @@
     | Type: integer
     | Default: 16320
     | Dynamic: **yes**
+
+.. _cfg_networking-iproto_threads:
+
+.. confval:: iproto_threads
```
NickVolynkin, line 75:
```suggestion
..   confval:: iproto_threads
```

This was indented with 2 spaces and the text below with 4. As a result, the text is formatted as a blockquote. See https://tarantool-develop.herokuapp.com/en/doc/pgulutzan-gh-2059/reference/configuration/#confval-iproto_threads.

pgulutzan: I did make a mistake but it doesn't seem to be with confval, which is done the same way that other items in the file are done. However, the text below was was indented with 5 spaces.

NickVolynkin: Oh, it's 5 spaces, indeed. Should be 4 as well.

PR #2115, opened 2021-04-30
add ddl/migrations section

artur-barsegyan:
**TODO:**

- [x] Mention this article at Reference pages: `create_space` and `format`


Totktonada:
Hmm, it is in Russian. Are you plan to translate it to English?

The description of tarantool/ddl gives impression that it is just `require('yaml').decode(io.open('/path/to/schema.yaml'):read('*all'))`. I think you should elaborate the set of problems the module solves.

I would shift accents a bit: tarantool/ddl is the dependency of cartridge, not part of cartridge. Key point: can be used per se.

The migrations description should be illustrated by examples: the format should be like Mons's article about queues. How to code a simple migration: description, code example. How to change it to handle some problem. How to change it to handle another problem. To illustrate why this format is necessary: a reader may not know what you assume by 'lock managenent'. Aside of this, if you'll write the relevant code youself, it'll make your suggestions more precise and in-depth.

It also worth to think around the most logical way to split 'pure tarantool', tarantool/ddl, tarantool/ddl + tarantool/migrations, tarantool/moonwalker, cartridge data schema, cartridge + migrations descriptions. It is about the website structure and a best way represent various knowledge around tarantool and its modules. Quite interesting topic.

artur-barsegyan:
> Hmm, it is in Russian. Are you plan to translate it to English?

Yes, we do. Maybe as a separate ticket.
 
> The description of tarantool/ddl gives impression that it is just `require('yaml').decode(io.open('/path/to/schema.yaml'):read('*all'))`. I think you should elaborate the set of problems the module solves.

I see no reason to elaborate on such details. Сan you explain?

> I would shift accents a bit: tarantool/ddl is the dependency of cartridge, not part of cartridge. Key point: can be used per see.

Agree with you. Will change it. 

> The migrations description should be illustrated by examples: the format should be like Mons's article about queues. How to code a simple migration: description, code example. How to change it to handle some problem. How to change it to handle another problem. To illustrate why this format is necessary: a reader may not know what you assume by 'lock managenent'. Aside of this, if you'll write the relevant code youself, it'll make your suggestions more precise and in-depth.

About migrations and code examples: I don't post any examples of migration (except for two safety) because any kind of migration without context is useless in general. My point is: if you will post a simple generic migration code, a user will copy-paste it and get some trouble after that. It was discussed with @Mons. 

The main goal of this PR: introduce the basics to newbies. At now we don't have an article with that information. In future, we can detail more caveats and describe best practices.

Totktonada:
> > The description of tarantool/ddl gives impression that it is just `require('yaml').decode(io.open('/path/to/schema.yaml'):read('*all'))`. I think you should elaborate the set of problems the module solves.
> 
> I see no reason to elaborate on such details. Сan you explain?

It is not about details, it is about problems the module solves. It is unclear from the description in the new section for me, so unlikely will be clear for the target audience.

> > The migrations description should be illustrated by examples: the format should be like Mons's article about queues. How to code a simple migration: description, code example. How to change it to handle some problem. How to change it to handle another problem. To illustrate why this format is necessary: a reader may not know what you assume by 'lock managenent'. Aside of this, if you'll write the relevant code youself, it'll make your suggestions more precise and in-depth.
> 
> About migrations and code examples: I don't post any examples of migration (except for two safety) because any kind of migration without context is useless in general. My point is: if you will post a simple generic migration code, a user will copy-paste it and get some trouble after that. It was discussed with @Mons.

You show several general points and state that the target audience is newcomers. I'll take 'lock management' for example:

> - Фоновый процесс мигрирует данные, и менеджит локи на чтение определенных записей в старом спейсе

Should I use fiber.cond / fiber.channel? Or something else? In conjuction with box.atomic()? How exaclty? How should I integrate it with accesses to the old space?

Do you think this suggestion will be useful in this form?

I disagree here. General words without any examples raise more questions than solves, at least here and at least for me.

> The main goal of this PR: introduce the basics to newbies. At now we don't have an article with that information. In the future, we can detail more caveats and describe best practices.

I don't understand: should a developer be able to write a migration after reading the section or not?

Totktonada:
> It is not about details, it is about problems the module solves. It is unclear from the description in the new section for me, so unlikely will be clear for the target audience.

We discussed it offline with Artur. I took this point back. Okay, tarantool/ddl solves several information schema related problems: it is distinct from a particular instance, provides the single access point, stores sharding information and likely will offer more in a future (relations, integration with a configuration storage client, more sharding related information). However we can describe it here as the helper to represent a database schema in a declarative way. But we should give a snippet how to use it this way, like so:

```lua
local fh = io.open('<...>', 'r')
local schema = yaml.decode(fh:read('*all'))
fh:close()
local ok, err = ddl.check_schema(schema)
if not ok then
    <..handle error..>
end
local ok, err = ddl.set_schema(schema)
if not ok then
    <..handle error..>
end
```

Maybe it worth to place something like this into the readme of the module.

doc/book/box/data_model.rst
```diff
@@ -892,3 +892,229 @@ resource usage of each function.
...
+
+--------------------------------------------------------------------------------
+Migrations
+--------------------------------------------------------------------------------
+
+Миграцией называется любое изменение схемы данных: добавление/удаления поля,
+добавление/удаления индекса, изменение формата поля и.т.д.
+
+В Tarantool есть только два вида миграции, после которых данные в спейсе
+останутся гарантированно консистентными:
```
sharonovd, line 1018: непонятная формулировка. это же про то что они позволительны без миграции данных.

doc/book/box/data_model.rst
```diff
@@ -892,3 +892,229 @@ resource usage of each function.
...
+
+Эти проблемы могут быть или не быть релевантными в зависимости от вашего приложения и требований к его доступности.
+
+**********************************************
+Что нужно знать при написании сложных миграций
+**********************************************
+
+В Tarantool есть механизм транзакций. Он полезен при написании миграции т.к. позволяет
+атомарно работать с данными. Но перед использованием механизма транзакций нужно знать об ограничениях.
+Подробнее о них описано :ref:`тут <atomic-atomic_execution>`.
```
sharonovd, line 1078: И про йилды тоже надо знать, видимо

doc/book/box/data_model.rst
```diff
@@ -892,3 +892,229 @@ resource usage of each function.
...
+Код миграций исполняется на запущенном инстансе Tarantool.
+Важно: ни один способ не гарантирует вам транзакционное применение миграций на всем кластере.
+
+**Способ 0**: написать миграции в коде основного приложения
+
+Тут все просто: при перезагрузке кода произойдет миграция данных в нужный момент и схема данных БД обновится.
+Однако, такой способ может подойти не всем. У вас может не быть возможности перезапустить Tarantool или
+обновить код через механизм hot-reload.
+
+**Cпособ 1**: утилита `tarantoolctl`
```
sharonovd, line 1093: А мы __правда__ хотим рекомендовать способы 1 и 2?

sharonovd: я бы их описал внизу и с пометкой что мы их не очень рекомендуем

doc/book/box/data_model.rst
artur-barsegyan: Ну это уже написано, когда ты перейдешь в раздел про транзакции

doc/book/box/data_model.rst
```diff
@@ -892,3 +892,260 @@ resource usage of each function.
...
+***************************
+
+Other types of migrations are also allowed but it would be more difficult to
+maintain data consistency.
+
+Migrations are possible in two cases:
+
+-   When Tarantool starts, and no client uses the database yet
+
+-   During request proccessing, when active clients already use the database
```
NickVolynkin, line 1068:
```suggestion
-   During request processing, when active clients already use the database
```

doc/book/box/data_model.rst
```diff
@@ -892,3 +892,260 @@ resource usage of each function.
...
+******************
+
+Index creation is described in the
+:doc:`/reference/reference_lua/box_space/create_index` method.
+
+***************************
+Other types of migrations
+***************************
+
+Other types of migrations are also allowed but it would be more difficult to
```
NickVolynkin, line 1061:
```suggestion
Other types of migrations are also allowed, but it would be more difficult to
```

PR #2116, opened 2021-04-30
Onvember/fix doc warnings

Onvember:
fix all possible sphinx warnings

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -70,7 +70,10 @@ exclude_patterns = [
     'images',
     'book/cartridge/cartridge_overview.rst',
     'book/cartridge/CONTRIBUTING.rst',
+    'book/cartridge/topics',
+    'book/cartridge/cartridge_api/modules/cartridge.test-helpers.rst',
     'reference/reference_rock/luatest/README.rst',
+    'reference/reference_rock/luatest/modules/luatest.rst',
     '**/_includes/*'
 ]
 
--- a/doc/getting_started/getting_started_cartridge.rst
+++ b/doc/getting_started/getting_started_cartridge.rst
@@ -17,7 +17,7 @@ Next, create an application named ``myapp``. Say:
 
 This will create a Tarantool Cartridge application in the ``./myapp`` directory,
 with a handful of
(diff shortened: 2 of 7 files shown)


doc/getting_started/getting_started_cartridge.rst
```diff
@@ -17,7 +17,7 @@ Next, create an application named ``myapp``. Say:
 
 This will create a Tarantool Cartridge application in the ``./myapp`` directory,
 with a handful of
-:ref:`template files and directories <creating-an-application-from-template>`
+:ref:`template files and directories <cartridge_cli_creating_an_application_from_template>`
```
NickVolynkin, line 20: Sometimes it's `cartridge_cli_`, other times it's `cartridge-cli-`. What's the logic behind the label naming?

conf.py
```diff
@@ -70,7 +70,10 @@
     'images',
     'book/cartridge/cartridge_overview.rst',
     'book/cartridge/CONTRIBUTING.rst',
+    'book/cartridge/topics',
+    'book/cartridge/cartridge_api/modules/cartridge.test-helpers.rst',
```
NickVolynkin, line 74: What is wrong with this page? https://www.tarantool.io/en/doc/latest/book/cartridge/cartridge_api/modules/cartridge.test-helpers/

Onvember: this page represents only 'see also' lines to other pages (cartridge.test-helpers.server, cartridge.test-helpers.cluster, cartridge.test-helpers.ectd, cartridge.test-helpers.stateboard) but doesn't really links to them. We have these pages in table of contents so cartridge.test-helpers is redundant here.

Onvember: But it causes several lines of warnings. By removing it, I reduced them to 1

doc/getting_started/getting_started_cartridge.rst
Onvember: Actually, more common usage is cartridge-cli-.... so I will just change this tag. 
We have some tag naming rules but here with cli I don't know how to follow them because it contains a dash in the name

Issue #1975, opened 2021-03-22
[3pt] Write about changing space schema


artur-barsegyan:
I don't found any articles about removing fields from space. Find a decision after 15 minutes of searching in Telegram community chat's history.

We need to write about adding/removing fields from space with data. Some answers are placed at StackOverflow. Some places at Telegram chat.

**Product:** Tarantool
**Since:** all versions
**Audience/target:** all users, especially "first 5-minutes noobies"
**Root document:** maybe at [`space:format()` page ](https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_space/format/#box-space-format)
**SME:** Mons, dsharonov
**Peer reviewer:** Me

# Details
*   Examples of code for removing a field from space
```lua
box.cfg{}

test = box.schema.create_space('test', { format = {
    {name='id', type = 'number'},
    {name = 'fullname', type = 'string'},
    {name = 'age', type = 'number'}
}, if_not_exists = true})

test:create_index('pk', {if_not_exists = true})
test:insert({1, 'aaa', 34})

test:format({})
test:replace({1, 34})

test = box.schema.create_space('test', { format = {
    {name='id', type = 'number'},
    {name = 'age', type = 'number'}
}, if_not_exists = true})
```

PR #2171, opened 2021-06-03
Add info about changing space schema

xuniq:
fixes gh-1975


artur-barsegyan:
Thanks for your PR.

There is already a section on migrations that describes the migration options. Check it out [here.](https://www.tarantool.io/ru/doc/latest/book/box/data_model/#migrations)
There is the link to that page at the top of the `format` method page.

We don't say to users about migrations like _' deleting field'_ or another one. These migrations have difficulty implementing so that customers of the service on Tarantool do not notice this. 

I suggest closing that PR and close the linked issue.

PR #2179, opened 2021-06-08
Proofread and translate Replication Leader Election

NickVolynkin:
https://www.tarantool.io/ru/doc/latest/book/replication/repl_leader_elect/

Part of #2178

PR #2180, opened 2021-06-09
Update repl_leader_elect.rst

patiencedaur:
"actual" in English means something completely different
@NickVolynkin


NickVolynkin:
Will commit this change directly to #2179

Issue #799, opened 2019-06-06
[3pt] box.space._vcollation


TarantoolBot:
_vcollation is a system space that represents a virtual view.
The structure of its tuples is identical to that of _collation.
Tuples of this sysview is always readable, except when the user
doesn't have "public" role.
Requested by @romanhabibov in https://github.com/tarantool/tarantool/commit/3e3ef182fbe4a2428a4d1c86039d171bbf478fa0.

Totktonada:
> [about _vcollation system space] permissions for certain tuples are limited in accordance with user privileges.

AFAIU a user usually able to access to a collation record using `_vcollation` even while (s)he has no access to _collation. The cited phrase may be understood as '_vcollation is more restrictive then _collation' due to the word 'limited'. I found that explicit

> Tuples of this sysview is always readable, except when the user doesn't have "public" role.

gives more information and easier to understand.

I'll reopen the issue to track the discussion.

pgulutzan:
Waiting for a reply in email thread = The public role and system views.

Totktonada:
There is https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_space/_vcollation/, which says:

> [...] permissions for certain tuples are limited in accordance with user privileges.

I guess some changes were missed. @NickVolynkin, can you, please, investigate what is going wrong here (and whether there are other missed changes like this one)?

pgulutzan:
I sent this set of copies of emails to the public dev mailing list, but might as well repeat here.


-------- Forwarded Message --------
Subject: 	The public role and system views
Date: 	Fri, 14 Jun 2019 17:20:03 -0600
From: 	Peter Gulutzan <pgulutzan@ocelot.ca>
To: 	Alexander Turenko <alexander.turenko@tarantool.org>
CC: 	dev@tarantool.org


Hi,

You suggested that I change _vollation description
https://github.com/tarantool/doc/issues/799#issuecomment-502238385
because the "public" role matters.

I did not document this because I did not know all of it.

() 'sysview' = 'system view' = any space whose name starts with _v and
   corresponds to a 'system space' with the same name except
   for the 'v', e.g. _vspace is the system view of _space.
   In _space we say that engine='sysview'.
() role 'public' initially has:
   'read' privilege for _vcollation
   'read' privilege for _vspace
   'read' privilege for _vsequence
   'read' privilege for _vindex
   'read' privilege for _vfunc
   'read' privilege for _vuser
   'read' privilege for _vpriv
[...]

PR #2261, opened 2021-07-29
Fixes gh-799 box.space._vcollation

pgulutzan:
There was a commit for issue#799 https://github.com/tarantool/doc/issues/799 in 2019
https://github.com/tarantool/doc/commit/80bcbd3dd83d3d53b45dedfd8c9cea2b27c4ebdd
but, as can be seen from the comments on issue#799, for some reason it disappeared.
So this does again what was in that commit,, but there are a few differences.
() Sometime between 2019 and 2021, somebody put some of the _v* (sysviews) descriptions
in separate files. This is not really necessary now because redundancy was removed,
but I didn't remove the files.
() There was description in _priv re how privileges work etc. I replaced it with
"See Access control for details about user privileges."

One of the people involved with the 2019 correspondence was @Totktonada.


patiencedaur:
Crowdin link for translation: https://crowdin.com/project/tarantool-docs/ru#/pgulutzan-gh-799/en/reference/reference_lua

patiencedaur:
@Totktonada 
We found that the term "system view" is not quite clear. It is a read-only copy of a system space, which has limited user permissions, correct? If so, we suggest renaming it to "system space view". "System view" can be read as "view of the system" or "view used by the system", while it is really a "view of a system space". What do you think?
In Russian, "system space" and "system space view" will be termed системный спейс and представление системного спейса, accordingly.

Totktonada:
Both 'system view' and 'system space view' sound ok for me. Technically the latter is more precise. I don't mind. However 'system views' terminology was introduced together with the feature: https://github.com/tarantool/tarantool/commit/4c15968178be0d0ade2517c54f0fb56e4973116d. I don't know, whether it worth to change it. Can we ask @locker as well?

> It is a read-only copy of a system space, which has limited user permissions, correct?

More or less, but it is a bit inaccurate description. We can say that a view is an object, which works like the original space, but with a set of automatically applied filters. If it would be a copy, its data would not be updated at updating of the original space. The idea of our views is very similar to [views in SQL](https://www.geeksforgeeks.org/sql-views/).

patiencedaur:
Thank you! There are only 4 matches for "system view" in the current doc. So if @locker approves and we decide to fix the English term, I expect it to be easy and relatively quick.

locker:
> Both 'system view' and 'system space view' sound ok for me. Technically the latter is more precise. I don't mind. However 'system views' terminology was introduced together with the feature: [tarantool/tarantool@4c15968](https://github.com/tarantool/tarantool/commit/4c15968178be0d0ade2517c54f0fb56e4973116d). I don't know, whether it worth to change it. Can we ask @locker as well?

@Totktonada @patiencedaur I'm fine either way. If there are just a few occurrences in the doc, I think it wouldn't hurt to update them.

Issue #2250, opened 2021-07-21
[3pt] New field - IPROTO_BALLOT_CAN_LEAD


TarantoolBot:
**Product:** Tarantool
**Root document:** https://www.tarantool.io/en/doc/latest/dev_guide/internals/box_protocol/
**SME:** @ Gerold103

# Details

It is sent as a part of `IPROTO_BALLOT (0x29)`. The field is a
boolean flag which is true if the sender has `election_mode` in
its config `'manual'` or `'candidate'`.

During bootstrap the nodes able to become a leader are preferred
over the nodes configured as `'voter'`.

The field key is `0x07`.
Requested by @Gerold103 in https://github.com/tarantool/tarantool/commit/583253cc192543a0ed818e2b556b14dcc0d6b220.

Commit message: https://github.com/tarantool/tarantool/commit/583253cc192543a0ed818e2b556b14dcc0d6b220

Issue #2088, opened 2021-04-21
[2pt] describe election_mode='manual'


TarantoolBot:
## Details

Manual election mode is introduced. It may be used when the user wants to
control which instance is the leader explicitly instead of relying on
Raft election algorithm.

When an instance is configured with `election_mode='manual'`, it behaves
as follows:
 1) By default, the instance acts like a voter: it is read-only and may
    vote for other instances that are candidates.
 2) Once `box.ctl.promote()` is called, the instance becomes a
    candidate and starts a new election round. If the instance wins the
    elections, it remains leader, but won't participate in any new elections.

https://github.com/tarantool/tarantool/commit/9f33de449516dcc8f8dcf63324c739b2bcc48b9c.

tbd after #2089

**root document:** https://www.tarantool.io/ru/doc/latest/reference/configuration/#cfg-replication-election-mode
**SME:** @sergepetrenko

[2pt]

PR #2328, opened 2021-09-08
Fixes gh-2250 New field - IPROTO_BALLOT_CAN_LEAD

pgulutzan:
The change includes a reference to election_mode='manual', which will be described for a different issue: #2088.
I have marked the version as 2.10, although I see that the change has also happened in 2.7 and following. And I did not say in the text that this was added in version x.xx, because there is no release document to point to. Perhaps I should say it is added in 2.10?


Gerold103 (approved):

patiencedaur:
> So you're suggesting: At the end of the IPROTO_BALLOT_CAN_LEAD, a new sentence: IPROTO_BALLOT_CAN_LEAD support was added simultaneously in version :doc:`2.6.3 </release/2.6.3>`, version :doc:`2.7.2 </release/2.7.2>`, and version :doc:`2.8.1 </release/2.8.1>`. ? If so, I'll add a new commit.

Please add that `IPROTO_BALLOT_CAN_LEAD` is supported since version 2.7.3.

2.6.3 does not contain it, this is a mistake.

pgulutzan:
Okay, I pushed a commit with those version numbers.

Issue #2206, opened 2021-06-24
[2pt] feedback: Monitoring: getting started | Tarantool


TarantoolBot:
**Product:**  tools (Monitoring)
**Root document:** https://www.tarantool.io/en/doc/latest/book/monitoring/getting_started/
**SME:** @ZolotarevIgor-mail-ru 

# Details

> **|To change metrics HTTP path in runtime, you may use the following configuration
> (to learn more about Cartridge configuration, see
> this|).**
> We don’t recommend to use it to set up metrics role, use set_<…>

https://www.tarantool.io/en/doc/latest/book/monitoring/getting_started/

HTTP 400 

https://www.tarantool.io/en/doc/latest/book/cartridge/topics/clusterwide-config/#managing-role-specific-data

yngvar-antonsson:
Closed by @patiencedaur in https://github.com/tarantool/metrics/pull/310

Issue #2298, opened 2021-08-19
[2pt] QUOTE() and DOUBLE argument


TarantoolBot:
# Update description of the QUOTE() function

Product: Tarantool
Version: 2.10.0-beta1
Root document: https://www.tarantool.io/en/doc/latest/reference/reference_sql/sql_statements_and_clauses/#quote
SME: @ ImeevMA

# Details

After this patch function QUOTE() will return argument in case it
receives DOUBLE value as an argument. The same for all other numeric
types. In case it was given value of non-numeric type, it will return
string representation of the argument.

Requested by @ImeevMA in https://github.com/tarantool/tarantool/commit/62d622f8ef83bc93ca2585d3dc000156d6d0c094.

Can be done via #2296 

---

Resolved by #2350

pgulutzan:
I think I understand.
quote(1.1) returned '1.1' in version 2.8, quote(1.1) returns 1.1 in 2.10.
quote(any number type) returns the same value as the input and the same type.
quote(false) and quote(uuid()) and quote(x'00') return the same value as the input and not the same type
(it as like saying cast(false as string) etc.)
I agree that this can be done as part of 2296 which I have assigned to myself.

ImeevMA:
Actually, I plan to make QUOTE() accept only STRING argument. Or drop this function. Not sure if either of this will be accepted. I will write a document.

Issue #2296, opened 2021-08-19
Behavior changes in SQL


pgulutzan:
After discussion in https://github.com/tarantool/tarantool/issues/5956
I am adding this list of things that must change in the Tarantool manual
due to changes that have appeared in version 2.9, mostly due to
pushes from @ImeevMA. I will assign to myself if nobody objects
within a few days.

Updated comment: I originally wrote "2.9.0" because that's where the
changes are now. But now I see
https://github.com/tarantool/tarantool/releases/tag/2.10.0-beta1
A comment in isssue #2316 acknowledges this issue but release-note changes described here will only be read.
I continue to assume the rest of the description is necessary. 

In release/2.10.0.rst:

This file has not been created yet.
When it appears, this should be in the SQL section:

**Breaking change**:  NUMBER data type
Arithmetic (+ * - / % & | ~) operations and bit-shift operations (>> <<) are now illegal.
Example:
SELECT number_column + 1 was legal in version 2.8; now it will cause an error.

**Breaking change**: SCALAR data type
Arithmetic (+ * - / % & | ~) operation and bit-shift operations (>> <<) are now illegal.
Concatenation (||) operations are now illegal.
Values in SCALAR columns now have data type 'SCALAR', not the value's data type.
Example:
TYPEOF(CAST(1 AS SCALAR)) was 'integer' in version 2.8; now it will be be 'scalar'.

**Breaking change**: Explicit cast
CAST() between BOOLEAN or VARBINARY and INTEGER or UNSIGNED or NUMBER data types is now illegal.
Example:
CAST(FALSE AS INTEGER) was 0 in version 2.8; now it will cause an error.

**Breaking change**: TYPEOF()
TYPEOF(NULL) will now return 'NULL' if type cannot be determined from context.
TYPEOF(-NaN) will now return 'double'.
TYPEOF(map_column) will now return 'varbinary' instead of 'map'.
Example:
SELECT TYPEOF(NULL); was 'boolean' in version 2.8; now it will be 'NULL'.

**Breaking change**: Implicit conversions
Arithmetic operators must now have numeric operands, string operands are illegal.
Example:
SELECT 1 + '1'; was 2 in version 2.8; now it will cause an error.

**Breaking change**: Operations on SCALAR columns
Since the type of a value no longer determines whether an operation is valid, comparisons and functions which require a specific type will no longer work.
Example: if table T has only one row with a scalar column containing 'a':
[...]

PR #2350, opened 2021-09-26
Fixes gh-2206 Behavior changes in SQL

### LINKED ISSUE #2295: [3pt] SCALAR and NUMBER values in SQL
# Scalar and number values in SQL

Product: Tarantool
Since: 2.10.0-beta1
Root document: ? https://www.tarantool.io/en/doc/latest/reference/reference_sql/
SME: @ ImeevMA
Peer reviewer: @

# Details

SCALAR values can now be compared with values of any other scalar type,
but cannot be implicitly cast to any other scalar type. This means that
SCALAR values cannot participate in arithmetic, bitwise operations,
concatenation, or functions that, by definition, do not accept SCALAR
values.

NUMBER values now also cannot be implicitly cast to any other numeric
type, which means that NUMBER values cannot participate in arithmetic
and bitwise operations, or in functions that, by definition, do not
accept NUMBER values.
Requested by @ImeevMA in https://github.com/tarantool/tarantool/commit/bd320d3d267e097a9fa176a93c7e69187dd5bbae.

pgulutzan:
@ImeevMA: I am suggesting you as a reviewer, but if you think it is more appropriate for other people it can be changed to whatever you think is reasonable. Since it is a large set of changes, it might be proper to have more than one dev reviewer. Since it is for 2.10 and the "release" page is not officially done yet, there is no rush. (That is why I have not yet added @NickVolynkin as a reviewer, but it will happen someday.) Of course this is related to issue #2296.

Some new comments:

* We will say "numeric" rather than "number" in some places. It's clumsy, but otherwise people will associate "number" with "NUMBER".
  The alternative is to say, repeatedly, "number (but not NUMBER)".

* I made a mistake saying "always round down". I see now that rounding is toward zero.

* In reference_sql/sql_features.rst: I decided to make no change. Although version 2.10 is a backward step with less standard compatibility than before, the examples for E011-04 and E011-06 still work. I noticed by chance that ALTER TABLE ... ADD COLUMN has been working for a while, so I changed that.

Some more new comments: regarding fourth commit:
Default function parameter types are at the end of SQL Statements and Clauses.
There are replies for all comments, although a few haven't been marked yet as resolved.

Resolves #2295, #2296, #2298


doc/reference/reference_sql/sql_user_guide.rst
```diff
@@ -921,18 +946,18 @@ Assignments and operations involving NULL cause NULL or UNKNOWN results. |br|
...
+    ~                To BOOLEAN | To numeric | To STRING | To VARBINARY | To UUID
     ---------------  ----------   ----------   ---------   ------------   -------
-    From BOOLEAN   | AAA        | S--        | A--       | ---          | ---
-    From number    | A--        | SSA        | A-A       | ---          | ---
-    From STRING    | S--        | S-S        | AAA       | A--          | SS-
+    From BOOLEAN   | AAA        | ---        | A--       | ---          | ---
+    From numeric   | ---        | SSA        | A--       | ---          | --
+    From STRING    | S--        | S--        | AAA       | A--          | SS-
     From VARBINARY | ---        | ---        | A--       | AAA          | SS-
     From UUID      | ---        | ---        | AA-       | AA-          | AAA
```
ImeevMA, line 962: Not sure if it should be mentioned here, but UUID cannot be implicitly cast to anything except UUID. The same for STRING/VARBINARY to UUID.

doc/reference/reference_sql/sql_user_guide.rst
```diff
@@ -969,26 +993,33 @@ But comparisons of values of different types are allowed if the definition is SC
...
 column is ``A--`` and the first letter of ``A--`` is for explicit cast and A means Always Allowed.
-The result is 1.
+The result is 'TRUE'.
 
 ``UPDATE ... SET varbinary_column = 'A'`` is illegal because the intersection of the "From STRING" row with the "To VARBINARY"
 column is ``A--`` and the second letter of ``A--`` is for implicit cast (assignment) and - means not allowed.
 The result is an error message.
 
-``1.7E-1 > 0`` is legal because the intersection of the "From number" row with the "To number"
+``1.7E-1 > 0`` is legal because the intersection of the "From nueric" row with the "To numeric"
```
ImeevMA, line 1013: I think here should be `numeric` instead of `nueric`.

ImeevMA (commented):
Hi! I believe all of the problems mentioned in #2296 have been documented here. See 11 comments below. I also want to say that some changes will be changed again in the next release.

doc/reference/reference_sql/sql_statements_and_clauses.rst
```diff
@@ -316,24 +318,22 @@ Two column values in a SCALAR column can have two different primitive data types
...
    is defined as SCALAR -- values which happen to be INTEGER will be changed
-   so their data type is STRING.
+   so their data type is SCALAR.
 #. There is no literal syntax which implies data type SCALAR.
-#. TYPEOF(x) is never SCALAR, it is always the underlying data type.
-   This is true even if ``x`` is NULL (in that case the data type is BOOLEAN).
-   In fact there is no function that is guaranteed to return the defined data type.
-   For example, ``TYPEOF(CAST(1 AS SCALAR));`` returns INTEGER, not SCALAR.
+#. TYPEOF(x) is always SCALAR, it is never the underlying data type.
+   This is true even if ``x`` is NULL (in that case the underlying data type is BOOLEAN).
```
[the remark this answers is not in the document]
pgulutzan: Okay, I'm removing the sentence and changing the previous one to
TYPEOF(x) is always 'scalar' or 'NULL', it is never the underlying data type.

doc/reference/reference_sql/sql_statements_and_clauses.rst
```diff
@@ -3339,9 +3352,9 @@ LIKELIHOOD
 
 Syntax:
 
-:samp:`LIKELIHOOD({expression}, {number literal})`
+:samp:`LIKELIHOOD({expression}, {numeric literal})`
```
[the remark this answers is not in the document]
pgulutzan: Okay, changing to DOUBLE.

doc/reference/reference_sql/sql_statements_and_clauses.rst
```diff
@@ -3339,9 +3352,9 @@ LIKELIHOOD
...
 
-Return the result of the expression, provided that the number literal is between 0.0 and 1.0.
+Return the result of the expression, provided that the numeric literal is between 0.0 and 1.0.
```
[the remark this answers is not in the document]
pgulutzan: Okay, changing to "Return the expression without change ..."

doc/reference/reference_sql/sql_statements_and_clauses.rst
```diff
@@ -3465,7 +3474,10 @@ which are part of SQL statements, because of SQL's rules that
 string literals are enclosed by single quotes, and single quotes
 inside such strings are shown as two single quotes in a row.
 
-Example: ``QUOTE('a')`` is ``'a'``.
+Starting with Tarantool version 2.10, arguments with non-STRING
+data types are returned without change.
```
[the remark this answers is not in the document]
pgulutzan: Okay, changing "non-STRING" to "numeric".

doc/reference/reference_sql/sql_statements_and_clauses.rst
```diff
@@ -3672,15 +3684,20 @@ Syntax:
 
 :samp:`TYPEOF({expression})`
 
-Return the :ref:`data type <sql_column_def_data_type>` of the expression.
+Return the :ref:`data type <sql_column_def_data_type>` of the expression,
+or, if expression is the name of a column defined as SCALAR, return 'scalar',
+or, if expression is NULL, return 'NULL'.
```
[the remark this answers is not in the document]
pgulutzan: If column a contains NULL then typeof(a) will return 'NULL' although NULL is not a data type.
I'm changing to
Return 'NULL' if the expression is NULL,
or return 'scalar' if the expression is the name of a column defined as SCALAR,
or return the :ref:`data type <sql_column_def_data_type>` of the expression.

doc/reference/reference_sql/sql_user_guide.rst
```diff
@@ -575,91 +592,97 @@ promotion to a broader type may occur to avoid overflow.
 Arithmetic with NULL operands will result in a NULL operand.
 
 In the following list of operators, the tag "(arithmetic)" indicates
-that all operands are expected to be numbers and should result in a number;
+that all operands are expected to be numerics (other than NUMBER) and should result in a nuericr;
```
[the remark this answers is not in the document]
pgulutzan: Oops. Yes.

doc/reference/reference_sql/sql_user_guide.rst
```diff
@@ -575,91 +592,97 @@ promotion to a broader type may occur to avoid overflow.
...
 the tag "(comparison)" indicates that operands are expected to have similar
 data types and should result in a BOOLEAN; the tag "(logic)"
 indicates that operands are expected to be BOOLEAN and should result in a BOOLEAN.
 Exceptions may occur where operations are not possible, but see the "special situations"
 which are described after this list.
 Although all examples show literals, they could just as easily show column identifiers.
 
+Starting with Tarantool version 2.10.1, arithmetic operands must usually cannot be NUMBERs.
```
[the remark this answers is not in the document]
pgulutzan: My mistake. I'm removing the words "must usually".

doc/reference/reference_sql/sql_user_guide.rst
```diff
@@ -575,91 +592,97 @@ promotion to a broader type may occur to avoid overflow.
...
 non-integer results.
 Example: ``5 / 2``, result = 2.
 
 ``%`` modulus (arithmetic)
-Divide second number into first number according to standard arithmetic rules.
+Divide second numeric into first numeric according to standard arithmetic rules.
 The result is the remainder.
 Example: ``17 % 5``, result = 2.
 
 ``<<`` shift left (arithmetic)
```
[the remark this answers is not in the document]
pgulutzan: Okay for %. And for the bit-wise operations, I'm changing "unsigned" to "non-negative INTEGER or UNSIGNED".

doc/reference/reference_sql/sql_user_guide.rst
pgulutzan: Currently typeof(quote(uuid())) is 'string' but I expect that will change.
Currently least(uuid(),'',x'00') is '' but I won't consider that to be implicit cast for comparison.
I thought that I had tested for some other implicit casts, but now I see that they don't work.
So, okay, I'm changing to S--.

doc/reference/reference_sql/sql_user_guide.rst
pgulutzan: Oops. Yes.

doc/reference/reference_sql/sql_statements_and_clauses.rst
```diff
@@ -3405,26 +3417,22 @@ Syntax:
...
 result is 'D09441'. If you now execute
 ``SELECT POSITION('A', 'ДA');``
 the result will be 2,
 because 'A' is the second character in the string.
-However, if you now execute
-``SELECT POSITION(X'41', X'D09441');``
-the result will be 3,
-because X'41' is the third byte in the byte sequence.
+
+Starting with Tarantool version 2.10, arguments with data type VARBINARY are illegal.
```
ImeevMA, line 3435: Actually, now I'm working on bringing back to the POSITION() an ability to work with VARBINARY. Though I'm not sure if it's worth mentioning this in the doc right now.

doc/reference/reference_sql/sql_user_guide.rst
```diff
@@ -473,34 +476,47 @@ FALSE is less than TRUE.
...
 between 0 and +2^64, or NULL.
 
 .. _sql_data_type_double:
 
-DOUBLE values are numbers that do contain decimal points (for example 0.5) or
+DOUBLE values are numeric that do contain decimal points (for example 0.5) or
 are expressed with exponential notation (for example 5E-1).
 The range of possible values is the same as for the IEEE 754 floating-point
-standard, or NULL. Numbers outside the range of DOUBLE literals may be displayed
+standard, or NULL. Nuumerics outside the range of DOUBLE literals may be displayed
```
ImeevMA, line 494: I believe here should be `Numerics` instead of `Nuumerics`.

doc/reference/reference_sql/sql_user_guide.rst
```diff
@@ -473,34 +476,47 @@ FALSE is less than TRUE.
...
 But NUMBER values may also also be integers, and, if so,
 arithmetic operation results will be exact rather than approximate.
-For example, if a NUMBER column ``X`` contains 5, then ``X / 2`` is 2, an integer.
+For example, if a NUMBER column ``X`` contains 5, then ``X / 2`` would be 2 if division is legal.
 There is no literal format for NUMBER (literals like ``1.5`` or ``1E555``
 are considered to be DOUBLEs), so use :ref:`CAST <sql_function_cast>`
-to insist that a number has data type NUMBER, but that is rarely necessary.
+to insist that a numeric has data type NUMBER, but that is rarely necessary.
 See the description of NoSQL type :ref:`'number' <index-box_number>`.
+Support for arithmetic and built-in functions with NUMBERs was removed in Tarantool version 2.10.1.
```
ImeevMA, line 507: Should we mention, that this is about "arithmetic" built-in functions? Some functions, such as typeof(), max(), least() and so on can work with NUMBERs.

doc/reference/reference_sql/sql_user_guide.rst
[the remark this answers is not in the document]
ImeevMA: For % INTEGER could be of any sign, for example:
```
tarantool> box.execute([[SELECT -123 % 4;]])
---
- metadata:
  - name: COLUMN_1
    type: integer
  rows:
  - [-3]
...

tarantool> box.execute([[SELECT 123 % -4;]])
---
- metadata:
  - name: COLUMN_1
    type: integer
  rows:
  - [3]
...
```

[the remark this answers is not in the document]
pgulutzan: Okay, I added  Example: -123 % 4, result = -3.

doc/reference/reference_sql/sql_user_guide.rst
pgulutzan: Okay, I changed to  built-in arithmetic functions

pgulutzan:
@ImeevMA: Okay, I did a new commit.
I accepted 3 of your 4 suggestions, but am not sure what to do re POSITION, see my question.
I marked earlier comments "resolved" since you didn't say there were problems with how I handled them.
If you decide that we should continue to say that POSITION will not work with VARBINARY, then
I believe that your acceptance will take care of issue #2295 and issue #2296, but not issue #2298.

pgulutzan:
@ImeevMA:
I am hoping that my last three commits address remaining possible troubles.
The one marked "seventh commit" is rather large, because I went through the SQL sections
and fixed many things, and most of them were not specifically due to version 2.10.
There were bug fixes and improvements that I didn't know about,
and some cases where I realized people would not understand well.
I apologize if this causes extra difficulty but I think you will see why I
needed to make the changes once you see them.
Regarding the fact that you didn't notice the earlier fixes -- I did them,
I am not sure why you didn't notice them. 
You've probably noticed that I also submitted four new issues on github.

veod32:
@ImeevMA Do we need another round of your review for the latest changes Peter did?

ImeevMA (approved):
Hi! Thank you for the fixes! LGTM.

pgulutzan:
@igormunkin: You self-requested a review two months ago. Are you still interested?

doc/reference/reference_sql/sql-features.rst
```diff
@@ -38,7 +38,7 @@ marked "Okay" will probably be balanced by tests which are unfairly marked "Fail
     |            |                                               |                                                          | columns (created from Lua) from FLOAT/DOUBLE/REAL ones  |
     |            |                                               |                                                          | (created from SQL).                                     |
     +------------+-----------------------------------------------+----------------------------------------------------------+---------------------------------------------------------+
-    | E011-03    | DECIMAL and NUMERIC data types                | ``CREATE TABLE td (s1 NUMERIC PRIMARY KEY);``            | Fail, DECIMAL and NUMERIC data types are not supported  |
+    | E011-03    | DECIMAL and NUMERIC data types                | ``CREATE TABLE td (s1 NUMERIC PRIMARY KEY);``            | Fail, NUMERIC data types are not supported              |
```
igormunkin, line 41: Minor: Is DECIMAL left intentionally in the second column?

igormunkin (approved):
@pgulutzan, sorry for such delay: PR has been sunk in GitHub notifications traffic. The changes LGTM, except a couple of tiny nits. Thanks a lot!

doc/reference/reference_sql/sql_user_guide.rst
```diff
@@ -535,24 +547,25 @@ UUID support in SQL was added in Tarantool version 2.9.1.
...
 it is BOOLEAN.
 
-All the SQL data types correspond to
+Most of the SQL data types correspond to
 :ref:`Tarantool/NoSQL types <details_about_index_field_types>` with the same name.
 There are also some Tarantool/NoSQL data types which have no corresponding SQL data types.
 If Tarantool/SQL reads a Tarantool/NoSQL value which has a type which has no SQL equivalent,
 Tarantool/SQL may treat it as NULL or INTEGER or VARBINARY.
-For example, ``SELECT "flags" FROM "_space";`` will return a column whose data type is ``'map'``.
+For example, ``SELECT "flags" FROM "_vspace";`` will return a column whose Lua type is ``'map'``.
```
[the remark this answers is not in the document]
pgulutzan: Thanks, and I did one more commit, I removed the word "Lua". Actually, as you know, this whole
paragraph will be obsolete because MAP support was recently added, but
we can handle that in a separate issue.

doc/reference/reference_sql/sql-features.rst
pgulutzan: Yes, that is the official Feature Name for feature E011-03 in the SQL:2016/Foundation document, so we have to say it that way.

doc/reference/reference_sql/sql_user_guide.rst
```diff
@@ -214,7 +214,7 @@ A literal has :ref:`data type = VARBINARY <sql_data_type_varbinary>`
 
 Here are four ways to put non-ASCII characters,such as the Greek letter α alpha,  in string literals: |br|
```
alexandra-mara, line 215:
```suggestion
Here are four ways to put non-ASCII characters, such as the Greek letter α alpha, in string literals: |br|
```

doc/reference/reference_sql/sql_user_guide.rst
```diff
@@ -214,7 +214,7 @@ A literal has :ref:`data type = VARBINARY <sql_data_type_varbinary>`
...
 First make sure that your shell program is set to accept characters as UTF-8. A simple way to check is |br|
-``SELECT hex('α');``
+``SELECT hex(cast('α' as VARBINARY));``
 If the result is CEB1 -- which is the hexadecimal value for the UTF-8 representation of α -- it is good. |br|
 (1) Simply enclose the character inside ``'...'``, |br|
```
alexandra-mara, line 219: @pgulutzan, this looks like a list, should we maybe format it like one? I mean, in ReStructuredText formatting.

doc/reference/reference_sql/sql_user_guide.rst
```diff
@@ -575,91 +588,100 @@ promotion to a broader type may occur to avoid overflow.
...
+Example for numerics: ``0 >= 0``, result = TRUE. Example for strings: ``'Z' >= 'Γ'``, result = FALSE.
 
 .. _sql_equal:
 
 ``=`` equal (assignment or comparison)
 After the word SET, "=" means the first operand gets the value from the second operand.
 In other contexts, "=" returns TRUE if operands are equal.
 Example for assignment: ``... SET column1 = 'a';``
-Example for numbers: ``0 = 0``, result = TRUE. Example for strings:  ``'1' = '2 '``, result = FALSE.
+Example for numerics: ``0 = 0``, result = TRUE. Example for strings:  ``'1' = '2 '``, result = FALSE.
```
alexandra-mara, line 684:
```suggestion
Example for numerics: ``0 = 0``, result = TRUE. Example for strings: ``'1' = '2 '``, result = FALSE.
```

doc/reference/reference_sql/sql_user_guide.rst
```diff
@@ -945,50 +968,54 @@ So AAA = Always for explicit, Always for Implicit (assignment), Always for Impli
...
 (8 hexadecimal digits) hyphen (4 hexadecimal digits) hyphen (4 hexadecimal digits) hyphen (4 hexadecimal digits) hyphen (12 hexadecimal digits),
 such as ``'8e3b281b-78ad-4410-bfe9-54806a586a90'``. |br|
 From VARBINARY to UUID is allowed only if the value is
 16 bytes long,
 as in ``X'8e3b281b78ad4410bfe954806a586a90'``. |br|
 The chart does not show To|From SCALAR because the conversions depend on the type of the value,
-not the type of the column definition.
-Explicit cast to SCALAR is allowed but has no effect, the result data type is always the same as the original data type.
-But comparisons of values of different types are allowed if the definition is SCALAR.
+not the type of the column definition.  Explicit cast to SCALAR is always allowed.
```
alexandra-mara, line 984:
```suggestion
not the type of the column definition. Explicit cast to SCALAR is always allowed.
```

doc/reference/reference_sql/sql_user_guide.rst
```diff
@@ -945,50 +968,54 @@ So AAA = Always for explicit, Always for Implicit (assignment), Always for Impli
...
+    Explicit cast from numeric to BOOLEAN, |br|
+    Explicit cast from BOOLEAN to numeric, |br|
+    Implicit cast from NUMBER to other numeric types for arithmetic or built-in functions. |br|
+    Implicit cast from numeric to STRING. |br|
+    Implicit cast from STRING to numeric.
+
 Examples of casts, illustrating the situations in the chart:
 
-``CAST(TRUE AS INTEGER)`` is legal because the intersection of the  "From BOOLEAN" row with the "To number"
+``CAST(TRUE AS STRING)`` is legal because the intersection of the  "From BOOLEAN" row with the "To STRING"
```
alexandra-mara, line 1001:
```suggestion
``CAST(TRUE AS STRING)`` is legal because the intersection of the "From BOOLEAN" row with the "To STRING"
```

alexandra-mara: @pgulutzan also it sounds now like this cast is legal because of the contents of the mentioned table, which is not actually true. :) Should we maybe rephrase it to make it clear that this cast is allowed because of the rules of casting or underlying algorithms in Tarantool? Or just leave this part out at all, and rephrase it like "``CAST(TRUE AS STRING)`` is legal; you can see that in the intersection of  "From BOOLEAN" row with the "To STRING" column" and so on.

alexandra-mara (commented):
Hello @pgulutzan,
During the translation I encountered a few punctuation moments, as well as a couple of other suggestions, I would be happy if you could take a look at them

doc/reference/reference_sql/sql_user_guide.rst
pgulutzan: @alexandra-mara: You are right, but "CAST(TRUE AS STRING) is legal; you can see that in the intersection of the  From BOOLEAN row with the To STRING column is A--" etc. would not be perfectly grammatical, so I would suggest "CAST(TRUE AS STRING) is legal. The intersection of the From BOOLEAN row with the To STRING column is A--" etc. I can make that change in each example, and fix the spaces, and push again. Okay?

alexandra-mara: @pgulutzan, yeah, sure

doc/reference/reference_sql/sql_user_guide.rst
pgulutzan: Okay, changed in latest commit.

doc/reference/reference_sql/sql_user_guide.rst
pgulutzan: That might be a good idea, but I think it should be considered for a different pull request. As you might have noticed, there are now some more data types, so after this is merged I will probably need to make a new pull request for "Even more behavior changes in SQL".

doc/reference/reference_sql/sql_user_guide.rst
pgulutzan: Okay, changed in latest commit.

doc/reference/reference_sql/sql_user_guide.rst
pgulutzan: Okay, changed in latest commit.

doc/reference/reference_sql/sql_user_guide.rst
pgulutzan: Okay, changed in latest commit.

doc/reference/reference_sql/sql_user_guide.rst
alexandra-mara: @patiencedaur should we make a task out of it and put it in backlog for now?

patiencedaur: @alexandra-mara Yes please. Can you create an issue so we don't forget about it?

doc/reference/reference_sql/sql_statements_and_clauses.rst
```diff
@@ -346,33 +345,37 @@ Two column values in a SCALAR column can have two different primitive data types
...
 
    That is only possible with Tarantool/NoSQL scalar rules, but ``SELECT SUM(s2)``
    would not be legal because addition would in this case require implicit casting
-   from VARBINARY to a number, which is not sensible.
-#. The result data type of a primitive combination is never SCALAR because we
-   in effect use TYPEOF(item) not the defined data type.
+   from VARBINARY to a numeric, which is not sensible.
+#. The result data type of a primitive combination is sometimes SCALAR although we
+   in effect use the primitive data type not the defined data type.
    (Here we use the word "combination" in the way that the standard document
```
patiencedaur, line 366: Hello @pgulutzan,
Is [this](https://www.sis.se/api/document/preview/916885/) the standard document you are talking about?

doc/reference/reference_sql/sql_user_guide.rst
```diff
@@ -755,14 +777,16 @@ for example ``SELECT 1e318 - 1e318;`` is NULL and ``SELECT 1e318 * 0;`` is NULL.
...
+In older Tarantool versions,
+a string would be converted to a numeric if it was used with an arithmetic operator and conversion was possible,
 for example ``'7' + '7'`` = 14.
 And for comparison, ``'7'`` = 7.
-This is called implicit casting. It is applicable for STRINGs and all numeric data types.
+This is called implicit casting. It was applicable for STRINGs and all numeric data types.
+Starting with Tarantool version 2.10, it is no longer supported.
 
 Limitations: (`Issue#2346 <https://github.com/tarantool/tarantool/issues/2346>`_) |br|
 * Some words, for example MATCH and REGEXP, are reserved but are not necessary for current or planned Tarantool versions |br|
```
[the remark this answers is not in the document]
pgulutzan: Since these words are reserved, they cannot be used except within quote marks, so yes their use would cause an error. Although it is a good idea to reserve a word if it is going to be used in a future Tarantool version, that is not the case here, or at least it wasn't the case when I wrote that and I know of no plan now. By the way, there are a few words that will be added to the reserved list soon.

doc/reference/reference_sql/sql_statements_and_clauses.rst
pgulutzan: Not exactly, that's the technical corrigendum i.e. the list of corrections for SQL:2011. We care about SQL:2016 as mentioned earlier in the manual, and we care about the actual document not a corrigendum. However, there wasn't significant change between 2011 and 2016 for this or indeed for most of the features that Tarantool tries to support, so we can just say "standard document".

doc/reference/reference_sql/sql_user_guide.rst
```diff
@@ -192,8 +192,8 @@ DOUBLE literals: |br|
 Examples: .0, 1.0, 1E5, 1.1E5. |br|
 A literal has :ref:`data type = DOUBLE <sql_data_type_double>` if it contains a period, or contains "E".
 DOUBLE literals are also known as floating-point literals or approximate-numeric literals.
-To represent "Inf" (infinity), write a real number outside the double-precision number range, for example 1E309.
-To represent "nan" (not a number), write an expression that does not result in a real number,
+To represent "Inf" (infinity), write a real numeric outside the double-precision numeric range, for example 1E309.
+To represent "nan" (not a number), write an expression that does not result in a real numeric,
```
patiencedaur, line 196: @pgulutzan By "real numeric", do you mean a number belonging to [this mathematical set](https://en.wikipedia.org/wiki/Real_number)? That is, to represent "nan", you have to enter square root of minus one, for example?

pgulutzan: True enough, although since we're talking about a restriction of the set, I prefer to think of the Wikipedia article about https://en.wikipedia.org/wiki/NaN. As described earlier, we avoid the English word "number" now except for the 'number' type, but perhaps in Russian translation there would be no such problem. Is this preventing you from approving the pull request? ~~~~

patiencedaur: No, but this is preventing me from finishing the translation.

> As described earlier, we avoid the English word "number" now except for the 'number' type
Could you please give me a link to this discussion? I did not know of it.

doc/reference/reference_sql/sql_user_guide.rst
pgulutzan: @patiencedaur: In the pull request https://github.com/tarantool/doc/pull/2523 you will see that I formatted the list with bullets.

doc/reference/reference_sql/sql_user_guide.rst
pgulutzan: What made the matter more necessary now was that the NUMBER data type no longer had the typical characteristics that one associates with numbers. So in the initial comment I said
"
We will say "numeric" rather than "number" in some places. It's clumsy, but otherwise people will associate "number" with "NUMBER".
The alternative is to say, repeatedly, "number (but not NUMBER)".
"
Since no reviewer objected, I think it's considered okay.

PR #2495, opened 2021-12-13
Remove webpagemap 110

NickVolynkin:


artembo (commented):
Would you please check does the translation work on the local build?

NickVolynkin:
@patiencedaur why?

patiencedaur:
I don't want to distract the DevX team to look into this, so I quickly checked with artembo if I could close this. The reasons are: I don't see how it makes the doc better, nobody is working on this PR, and it looks stale.

PR #2511, opened 2021-12-21
Remove caption headerlinks from images/figures

NickVolynkin:


artembo:
Already in `latest`  
Closing PR

Issue #2516, opened 2021-12-22
[8pt] New SQL data types and function changes


pgulutzan:
The new data types in SQL -- MAP, ARRAY, ANY -- should be done together because they affect the same charts and subsections of the SQL Reference. Also there might be an effect on SQL functions so the tentative plan is to work on those at the same time if all the changes are in the same .rst file.
Thus fixing this issue may be fixing:
Issue #2406 Literals for INTEGER, DECIMAL and DOUBLE
Issue #2439 SUBSTR() function
Issue #2440 Field type ANY in SQL
Issue #2455 Default types of SQL built-in functions
Issue #2461 Field type ARRAY in SQL
Issue #2462 Field type MAP in SQL
Issue #2475 Syntax for ARRAY in SQL

patiencedaur:
Dear Peter @pgulutzan,

I am now the Tarantool documentation team lead instead of @NickVolynkin.

Once you're finished working on an issue, please tag me in a comment to the PR so I know that the task is ready. Then we can review it and pass it to the translation team.

It's best if you do it for smaller chunks of work rather than large ones. For example, you described this issue as an epic overarching several other issues. In this case, tag me when you're finished with each smaller issue, not with the whole epic. This will simplify and accelerate our work.

Thank you!

P. S. We're going to update our style guide, making it more detailed. Just keeping you informed. I'll send you a link once it's finished. Here's the work in progress if you're interested: https://github.com/tarantool/doc/pull/2539.

patiencedaur:
@pgulutzan
Yes, please add me to the reviewer list. Here is my work email, which you can CC the reports to: d.daur@corp.mail.ru.

I understand that this epic is about making small changes to many documents. So you're right---it would be best if you notify when the whole thing is done.

How do you usually get the next assignment to work on---should I just assign issues to you so you pick them up?

pgulutzan:
Thanks for understanding. You've probably noticed that I used to be involved with all documentation (because K. Osipov hired me first), but nowadays I only get a few assignments related to internals or (especially) to SQL. Sometimes I self-assign what looks interesting, and always I look for flaws; you can check with the developers that I've bothered recently (e.g. M. Imeev, M. Siomkin, T. Safin, A. Turenko) and they can tell you whether they have any problems with this. And yes you can assign issues to me whenever you wish.

patiencedaur:
Thank you, Peter. We will definitely have something about SQL to document quite soon.

Here are the style guide updates, as promised:
[...]

Issue #2475, opened 2021-12-01
 Syntax for ARRAY in SQL


TarantoolBot:
The syntax for creating ARRAY values is available in SQL. You can use
`[` and `]` to create an ARRAY value - all values in those brackets will
be part of ARRAY. The position of the values will be translated to the
same positions in ARRAY.

Examples:
```
tarantool> box.execute("SELECT [1, 'a', 1.5];")
---
- metadata:
  - name: COLUMN_1
    type: array
  rows:
  - [[1, 'a', 1.5]]
...
```

```
tarantool> box.execute("SELECT [1, 'a', ['abc', 321], 1.5];")
---
- metadata:
  - name: COLUMN_1
    type: array
  rows:
  - [[1, 'a', ['abc', 321], 1.5]]
...
```
Requested by @ImeevMA in https://github.com/tarantool/tarantool/commit/8dab66b0b53a8dc9f3b2bd9de1bfe8bb47409e9b.

pgulutzan:
This will require changes to the railroad diagrams and the charts showing data type definitions and comparisons and assignments. I have added a server issue  6668 "Problems with arrays".

pgulutzan:
Earlier I had missed that there was an issue #2461 Field type ARRAY in SQL. That issue and this issue can be handled together.
I have now seen explanations from @ImeevMA for server issue 6668 "Problems with arrays", but there might be some more small questions that can be answered there.

This should wait until merge occurs for the pull request of issue #2350, which is still waiting
for approval from @NickVolynkin. But when we get to it, it seems that we'll be saying:

In the section about Literals:
Change "five kinds of literals" to "six kinds of literals".
Add
ARRAY literals:
[left square bracket] [comma-separated list of values] [right square bracket]
Examples: [1,2,3,4], [1,[2,3],4], ['a',X'22',uuid()]
A literal has data type = ARRAY if it is a sequence of zero or more values
enclosed in square brackets ([ and ]).
Values may be of any type; values may be expressions; arrays may be nested.
[[ Perhaps "elements" is a better term, this is undecided. ]]

In the section Reserved Words:
Add ARRAY.

In the section Operand data types:
Add in the chart: SQL type=ARRAY, NoSQL type=array, no aliases, minimum=[], Maximum=?
[[I don't think there's a maximum depth or a maximum number of values]]
+
ARRAY values are any sequence of zero or more values, or NULL.
ARRAY literal values are expressed as a comma-separated list within matching square brackets.
[...]

Issue #2462, opened 2021-11-24
 Field type MAP in SQL


TarantoolBot:
Properties of type MAP in SQL:
1) a value of type MAP can be implicitly and explicitly cast only to
ANY;
2) only a value of type ANY with primitive type MAP can be explicitly
cast to MAP;
3) a value of any other type cannot be implicitly cast to MAP;
4) a value of type MAP cannot participate in arithmetic, bitwise,
comparison, and concatination operations.
Requested by @ImeevMA in https://github.com/tarantool/tarantool/commit/1364629ecf3c939d16d3f51e221854f4bbf7ba61.

pgulutzan:
@ImeevMA: Have I understood this correctly?
MAP exists only so that  there's correspondence with Lua 'map'.
SQL does not have MAP literals and there is no plan that it will ever have MAP literals.
But MAP values can be retrieved with Lua, for example
box.execute([[SELECT TYPEOF(lua('return {group_id = 1}'));]])
returns 'map'.
The functions that can accept MAP values are (though I know you don't call CAST a function):
CAST COALESCE IFNULL PRINTF QUOTE TYPEOF.

Although CAST(map-value TO STRING) is impossible, I can convert maps to strings with
PRINTF('%s',map-value) or QUOTE(map-value).  Should this fact be documented?
They both return the same thing, for example
'{"group_id": 1}'
The interesting thing here is that "group_id" is inside ""s not inside ''s,
as if it's an identifier. But if I use Lua e.g. in box.space._vspace:select() I see
{'group_id': 1}
Is this an inconsistency that somebody might care about?

And of course SELECT DISTINCT map-column FROM table-name; causes a segmentation fault,
but you already know that, it is server issue 6668.

ImeevMA:
@pgulutzan I believe you are a bit wrong.
We will have the syntax for creating MAP values, it will be as follows: {key : value}. It is currently in review.

> The functions that can accept MAP values are (though I know you don't call CAST a function):
CAST COALESCE IFNULL PRINTF QUOTE TYPEOF.

Functions that accept ARRAY also accept MAP. I think there is a bit more functions, I will check later.

> Although CAST(map-value TO STRING) is impossible, I can convert maps to strings with
PRINTF('%s',map-value) or QUOTE(map-value). Should this fact be documented?

Yes.

>  They both return the same thing, for example
'{"group_id": 1}'
The interesting thing here is that "group_id" is inside ""s not inside ''s,
[...]

Issue #2461, opened 2021-11-24
 Field type ARRAY in SQL


TarantoolBot:
Properties of type ARRAY in SQL:
1) a value of type ARRAY can be implicitly and explicitly cast only
   to ANY;
2) only a value of type ANY with primitive type ARRAY can be explicitly
   cast to ARRAY;
3) a value of any other type cannot be implicitly cast to ARRAY;
4) a value of type ARRAY cannot participate in arithmetic, bitwise,
   comparison, and concatination operations.
Requested by @ImeevMA in https://github.com/tarantool/tarantool/commit/e2102a15fbf7da9189db5a6f2b133b23bf5967bf.

PR #2523, opened 2021-12-24
Fixes gh-2516 New SQL data types and function changes

pgulutzan:
Issue #2516 is intended as a fix for
Issue #2406 Literals for INTEGER, DECIMAL and DOUBLE
Issue #2439 SUBSTR() function
Issue #2440 Field type ANY in SQL
Issue #2455 Default types of SQL built-in functions
Issue #2461 Field type ARRAY in SQL
Issue #2462 Field type MAP in SQL
Issue #2475 Syntax for ARRAY in SQL
Probably reviewers will include @ImeevMA and @NickVolynkin.
Probably there will be several commits so consider this a first draft.
A possible controversial point  is that I introduced arrays and maps in a section about literals rather than a later section about expressions -- putting them later would require some reorganizing.
UPDATE (2022-02-07): This is no longer a draft, so reviewer @ImeevMA can look at it whenever it's convenient.


ImeevMA:
Hi, @pgulutzan! I will review this PR. I also want to say that there will be a few more patches related to MAP and ARRAY. There will be a patch that adds a syntax for MAP values, patches that fix an error when MAP or ARRAY is used in ORDER BY/GROUP BY, and patches that fix some problems with using MAP and ARRAY in C user-defined functions. Also, it's possible that I will find more problems.

pgulutzan:
@ImeevMA: "A possible controversial point is that I introduced arrays and maps in a section about literals rather than a later section about expressions -- putting them later would require some reorganizing." What is your opinion?
By the way, I will make small style changes too, for example now we should if possible avoid saying "we" in he manual.

pgulutzan:
@patiencedaur:

In sql_plus_lua.rst, examples that work in earlier versions will fail in Tarantool version 2.10.0.
So I'm adding comments in the code to tell users what to do if they have earlier versions.
For example:

.. code-block:: lua

    box.schema.func.create('_DECODE',
       {language = 'LUA',
        returns = 'string',
        body = [[function (field, key)
                 -- If Tarantool version < 2.10.0, replace next line with
                 -- return require('msgpack').decode(field)[key]
                 return field[key]
                 end]],
        is_sandboxed = false,
        -- If Tarantool version < 2.10.0, replace next line with
        -- param_list = {'string', 'string'},
        param_list = {'map', 'string'},
        exports = {'LUA', 'SQL'},
        is_deterministic = true})

The comments should be translated for the Russian manual.
Alternative 1: Duplicate every example so there's an old and a new version.
Alternative 2: Put the instructions outside, and refer to line numbers.
If you like one of the alternatives,
or if you have a better idea,
we'll do whatever you prefer.

patiencedaur:
@pgulutzan Yes, let's do what you suggest. Thanks for letting me know.

doc/reference/reference_sql/sql_user_guide.rst
```diff
@@ -432,40 +449,81 @@ this manual will not use that term and will instead say "byte sequence".
...
+     - ffffffff-ffff-ffff- |br| dfff-ffffffffffff
+   * - SCALAR
+     - (varies)
+     - (none)
+     - FALSE
+     - maximum UUID value
+   * - MAP
+     - map
+     - (none)
+     - ``{'':''}``
```
ImeevMA: I think `{}` can be considered smaller than `{'':''}`.

doc/reference/reference_sql/sql_user_guide.rst
```diff
@@ -555,17 +613,45 @@ Prior to Tarantool version 2.10.1, individual column values had
...
+:ref:`QUOTE <sql_function_quote>`,
+:ref:`TYPEOF <sql_function_typeof>`, and functions involving NULL comparisons.
+
+.. _sql_data_type_any:
+
+ANY can be used for
+:ref:`column definitions <sql_column_def_data_type>` and the individual column values have
+type ANY.
+The difference between SCALAR and ANY is:
+SCALAR columns may not contain MAP or ARRAY values, but ANY columns may contain them.
```
ImeevMA: I think we should add that ANY values are not comparable (SCALAR values are comparable).

doc/reference/reference_sql/sql_user_guide.rst
```diff
@@ -808,10 +895,44 @@ value IS [NOT] NULL |br|
 CASE ... WHEN ... THEN ... ELSE ... END |br|
   ... for setting a series of conditions.
 
+.. _sql_map_expression:
+
+{ key : value } |br|
+... for MAP expressions. |br|
+Literal Examples: ``{'a':1}``, ``{ "column_1" : X'1234' }`` |br|
+Non-literal Examples: ``{"a":"a"}``, ``{UUID(), (SELECT 1) + 1}`` |br|
```
ImeevMA: I think here should be `{UUID(): (SELECT 1) + 1}`. Also, I think we should add some examples of MAP values where there is more than one key-value pairs, for example: `{1:'a123', 'two':uuid()}`.

doc/reference/reference_sql/sql_user_guide.rst
```diff
@@ -808,10 +895,44 @@ value IS [NOT] NULL |br|
...
+An expression has data type = MAP if it is enclosed in curly brackets
+(also called braces) ``{`` and ``}`` and contains a key for identification,
+then a colon ``:``, then a value for what the key identifies.
+The key data type must be INTEGER or STRING or UUID.
+The value data type may be anything.
```
ImeevMA: I think we should add that the value can also be a MAP, so MAPs can be nested (as was done for ARRAY).

doc/reference/reference_sql/sql_user_guide.rst
```diff
@@ -808,10 +895,44 @@ value IS [NOT] NULL |br|
...
+An expression has data type = ARRAY if it is a sequence of zero or more values
+enclosed in square brackets (``[`` and ``]``).
+Often the values in the sequence are called "elements".
+The element data type may be anything, including ARRAY -- that is, ARRAYs may be nested.
+Different elements may have different types.
+The Lua equivalent type is `'array' <https://www.lua.org/pil/11.1.html>`_.
+
+.. _sql_array_index_expression:
+
+ARRAY index expression: |br|
```
ImeevMA, line 928: I think we should expand this part to include `[]` for MAP. Or it should be dropped, since the patch is currently in review: https://github.com/tarantool/tarantool/pull/7140

ImeevMA (approved):
Hi! Thanks for the patch-set. I have a few comments below, but nothing major. I only reviewed the part about MAP and ARRAY.

Issue #2496, opened 2021-12-14
[8pt] Translate SQL reference


patiencedaur:
## Info

**Page:** https://www.tarantool.io/ru/doc/latest/reference/reference_sql/
**Crowdin:** TBA
**Words:** about 10,000
**Reviewer:** TBA

Only a couple of small sections of this reference have been translated. (see https://github.com/tarantool/doc/pull/2350)

- [ ] What Tarantool’s SQL product delivers
- [ ] Beginners' guide
- [ ] User guide
- [ ] Statements and clauses
- [ ] SQL plus Lua
- [x] SQL features (https://github.com/tarantool/doc/pull/2537)

patiencedaur:
This issue is not a bug fix request, it was created solely for translation purposes. Now that Russian translations are frozen, this issue has become irrelevant. Closed.

PR #2537, opened 2022-01-10
Translate SQL features

patiencedaur:
Part of https://github.com/tarantool/doc/issues/2496
Page: https://www.tarantool.io/ru/doc/latest/reference/reference_sql/sql-features/
Crowdin: https://crowdin.com/translate/tarantool-docs/84117/en-ru?filter=basic&value=2
Deployment: https://develop.tarantool.io/ru/doc/translate-sql-features/reference/reference_sql/sql-features/
Words: 2466

Refactor the table of SQL features to make it maintainable and translatable.


doc/reference/reference_sql/sql-features.rst
```diff
@@ -14,443 +13,1032 @@ else we will mark it "Fail".
...
+            -   Tests 
+        *   -   T631   
+            -   IN predicate with one list element
+            -   ``SELECT * FROM t WHERE 1 IN (1);``   
+            -   Okay.   
+
+
+Total number of items marked "Fail": 67
+
+Total number of items marked "Okay": 79
```
NickVolynkin: Please add trailing newline

patiencedaur: There is a trailing newline, its number is 1045.

doc/reference/reference_sql/sql-features.rst
```diff
@@ -14,443 +13,1032 @@ else we will mark it "Fail".
...
+        *   -   Feature ID
+            -   Feature
+            -   Example
+            -   Tests 
+        *   -   F471
+            -   Scalar subquery values
+            -   ``SELECT s1 FROM t WHERE s1 = (SELECT count(*) FROM t);``   
+            -   Okay.
+        *   -   F481   
+            -   Expanded NULL зredicate
```
igorkozloff:
```suggestion
            -   Expanded NULL predicate
```

alexandra-mara:
```suggestion
            -   Expanded NULL predicate
```

doc/reference/reference_sql/sql-features.rst
```diff
@@ -14,443 +13,1032 @@ else we will mark it "Fail".
...
+        :header-rows: 1
+
+        *   -   Feature ID
+            -   Feature
+            -   Example
+            -   Tests
+        *   -   E153   
+            -   Updatable queries with subqueries   
+            -
+            -
```
alexandra-mara: Hello @pgulutzan,
There's no example and no test results for E153, is that correct? This table contains tests results for all other items, even cursors, though Tarantool doesn't support them and we have no examples for them.

alexandra-mara: Hello @pgulutzan,
There's no example and no test results for E153, is that correct? This table contains tests results for all other items, even cursors, though Tarantool doesn't support them and we have no examples for them.

doc/reference/reference_sql/sql-features.rst
```diff
@@ -14,443 +13,1032 @@ else we will mark it "Fail".
...
+        *   -   E121-01   
+            -   DECLARE CURSOR   
+            -
+            -   Fail. Tarantool doesn't support cursors.   
+        *   -   E121-02   
+            -   ORDER BY columns need not be in select list
+            -   ``SELECT s1 FROM t ORDER BY s2;``
+            -   :ref:`Okay <sql_order_by>`.   
+        *   -   E121-03   
+            -   Value expressions in select list
```
alexandra-mara: @pgulutzan, is this description correct? Shouldn't there be something like "Value expressions in ORDER BY clause"?

doc/reference/reference_sql/sql-features.rst
pgulutzan: It is not correct, I should have given an example.
I think UPDATE "view_containing_subquery" SET column1=0; would be good enough, and the result would be Fail.
I can add it in the English manual if nobody objects.
I am changing sql-features.rst anyway because of the style guide, see
https://github.com/tarantool/doc/commit/b63189cbf4dcc955f6b0f2bd4e14c4817811c26d

doc/reference/reference_sql/sql-features.rst
pgulutzan: @alexandra-mara: You are right, I erred, it should be "Value expressions in ORDER BY clause". Again, I can fix that in the English manual if there are no objections.

doc/reference/reference_sql/sql-features.rst
alexandra-mara: @pgulutzan this change should be performed in this PR then, as the table was refactored according to our style guides. If you can add this change to this PR, please do, if it's not convenient I can do it.
And can you please wait for us to merge this PR before finishing your work on gh-2516? This PR's almost ready, translation is completed, and I'm finishing the review

doc/reference/reference_sql/sql-features.rst
alexandra-mara: @pgulutzan If you can add this change to this PR, please do, if it's not convenient I can do it

doc/reference/reference_sql/sql-features.rst
pgulutzan: Whoops, I see now that we've both made the same change. Okay, you have priority because I must wait for a dev person to approve  all the other changes that I made in the SQL sections. I can cancel what I did in sql-features.rst, and you can make this change in "this PR" (i.e. 2537).

doc/reference/reference_sql/sql-features.rst
pgulutzan: @alexandra-mara: I believe it will be better to leave it to you, I am trying to avoid causing conflict. I see that someone has assigned me as a reviewer of this task, so please let me know when it is ready.

alexandra-mara: @pgulutzan that was an automatic assignment, as you commented on this PR you're considered its reviewer :)

doc/reference/reference_sql/sql-features.rst
```diff
@@ -14,443 +13,1032 @@ else we will mark it "Fail".
...
+            -   Fail. Tarantool doesn't have CREATE TABLE inside CREATE SCHEMA.   
+        *   -   F311-03   
+            -   CREATE VIEW   
+            -   Fail. Tarantool doesn't have CREATE VIEW inside CREATE SCHEMA.   
+        *   -   F311-04   
+            -   CREATE VIEW: WITH CHECK OPTION   
+            -   Fail. Tarantool doesn't have CREATE VIEW inside CREATE SCHEMA.   
+        *   -   F311-05   
+            -   GRANT statement   
+            -   Fail. Tarantool doesn't have GRANT inside CREATE SCHEMA.
```
alexandra-mara: @pgulutzan, since you're here, maybe you can help me out with this section, F311. It seems a little confusing that we first state that Tarantool doesn't have schemas, so we can't perform CREATE SCHEMA, but then we say that “Tarantool doesn't have CREATE TABLE inside CREATE SCHEMA”. If we can't perform CREATE SCHEMA, isn't it the main reason we can't comply with other items in F311 section?

pgulutzan: I don't see how that is confusing. Suppose someday we allow CREATE SCHEMA, but without any GRANT clauses within it. Then "Tarantool doesn't have GRANT inside CREATE SCHEMA" is still true. So CREATE SCHEMA is an essential prerequisite, but the specific thing that is missing is GRANT.

alexandra-mara: @pgulutzan okay, got it :)

alexandra-mara (approved):
Hello @pgulutzan, @patiencedaur
I made the changes we've discussed above, and I believe this PR is ready to merge

pgulutzan (approved):
Since I'm only marked as reviewer because of my comment, I only looked for the changes that we discussed.
I didn't see the change to E153 but see it is going to be done for issue #2659, and it will be marked Fail. So I initially thought that the count at the end, Total number of items marked “Fail”: 67, would need to be changed to 68. But now I counted the number of times that I see "Fail." or "Fail," not counting F021, it seems it would be 65. But I'm not good at counting. 
So I will say that it looks good to me.

alexandra-mara:
@pgulutzan I've updated E153 with the example you provided via patch that was merged into this branch, you should be able to see it here https://github.com/tarantool/doc/pull/2537/commits/751b017fa305c4a2f4a20a2823b40508210394e4
Yes, of course we'll have to recount test results, I haven't thought of that. I actually searched through this page for "Fail" occurrences and there were 55, excluding those not from table and F021. And 80 "Okay"s. I was checking our current test deployment with these updates, it is not live yet

pgulutzan:
@alexandra-mara: the number 55 surprised me, and now I have realized: all the privilege features are gone. This isn't right, because the list is supposed to be the same as the list in the standard. I suggest that I can put them back as part of the SQL changes for issue #2523. Also I might try changing the "widths:" to see whether it looks better if there is more space in the Example column.

alexandra-mara:
@pgulutzan, I'm sorry I missed your message, I was away for a week. Yeah, surely we need to add privilege features, as part of #2523, or we can create a new issue

pgulutzan:
@alexandra-mara: in my last commit for issue #2523 https://github.com/tarantool/doc/pull/2523/commits/ba284f0b86c9e677d66ce75754841618c6d9f10d  I copied your version of sql-features.rst and added the privilege features and removed a reference to "we" due to another change in the style guide. If I remember correctly, I didn't make other significant changes.

Issue #2028, opened 2021-04-05
[2pt] Style guide for list and table items


NickVolynkin:
## Style guide for list and table items

**Root document:** https://www.tarantool.io/en/doc/latest/contributing/docs/style/ 

Add the guidelines for writing lists and tables to the styleguide.

- capitalization?
- commas, semicolons, dots?
- no conjunctions like in #2029

patiencedaur:
Closed via #2539

PR #2539, opened 2022-01-11
Add more style-related instructions on writing docs

patiencedaur:
- Style recommendations
- Advice on table formatting and code snippets
- Altered structure

This is to help our external contributors.


doc/contributing/docs/style.rst
```diff
@@ -56,8 +71,90 @@ Simple sentences are easier to read, understand and translate.
...
+
+Bad example:
+
+..  image:: images/dont.png
+    :width: 400
+    :alt: Don't "draw" tables with ASCII characters
+
+Good example:
+
+..  image:: images/do.png
```
arctic-dreamer, line 122: заметила, что строки в таблице не отбиты пустой строкой — вроде в нашем гайдлайне в примере есть пустая строка + помню, коля говорил, что лучше отбивать: https://www.tarantool.io/en/doc/latest/contributing/docs/markup/tables/

doc/contributing/docs/style.rst
```diff
@@ -73,22 +170,28 @@ aren't familiar with the
...
+Consistent ending punctuation
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-    Use full :doc:`link names </contributing/docs/markup/links>`.
-    
\ No newline at end of file
+Make sure that your lists and tables have consistent punctuation---either
+all similar list/table items end with a period or they all don't.
+In the example below, *all* items in the second column don't have
+ending punctuation. Meanwhile, *all* items in the fourth column end with a period:
```
Totktonada, line 194: IFAIU, we should prefer to end a list/table item with a pediod if most of items are a kind of sentence (or even several sentences), not just noun group. Similar to puntuation of lists in Russian. (It is just common sense, I don't know, whether there are grammar rules around this for English.)

What is interesting is when we have a list/table of noun groups (or short phrases). Am I understand correct: we should NOT end an items with a comma, just leave it without punctuation marks? It is different from Russian, but that's what I usually see in English texts. I don't know whether it is normative or informal 'internet style'.

doc/contributing/docs/style.rst
patiencedaur: не знала, что это работает. спасибо!

doc/contributing/docs/style.rst
patiencedaur: In my experience, punctuation rules differ across languages.

For English docs, we've been using recommendations from the Microsoft Style Guide, which I'll cite below. I see no reason to stop relying on them now.

[For lists](https://docs.microsoft.com/en-us/style-guide/scannable-content/lists#punctuation):

In bulleted and numbered lists, end each list item with a period if:
- Any item forms a complete sentence when combined with the list introduction that precedes the colon.
(Exception) Don’t use periods if all items have three or fewer words or if the items are UI labels, headings, subheadings, or strings.
- Any item by itself is a complete sentence.

Don’t use semicolons, commas, or conjunctions (like and or or) at the end of list items.

[For tables](https://docs.microsoft.com/en-us/style-guide/scannable-content/tables#punctuation):

For the text in cells, use periods or other end punctuation only if the cells contain complete sentences or a mixture of fragments and sentences.

Totktonada: Thanks! I would like to see an explicit recommendation in the guide, because 1) it is not obvious for me what is normative; 2) we have the section that looks quite relevant.

arctic-dreamer (approved):

doc/contributing/docs/style.rst
patiencedaur: We're planning to do it in https://github.com/tarantool/doc/issues/2028. Is that OK?

Totktonada: That's nice, thanks again!

Totktonada: Hm. But the updated version states the following:

> For the text in cells, use periods or other end punctuation
> only if the cells contain complete sentences or a mixture of fragments and sentences.

And it is not very clear for me:

1. Whether it is about Russian, English or both (I guess English)?
2. Whether I can use a comma for complete sentences (I guess, no)?
3. If we're about English, I would state explicitly 'don't use a comma or a period when all items are not sentences'. It can be deduced from current wording, but requires some mental effort. Explicit is better than implicit.
4. If we're about Russian, I would state explicitly that formatting (as a list) does not change punctuation rules and the comma should be used between homogenous parts of a sentence.

Sorry for nits.

patiencedaur: Nits are good when you're making instructions. :)

I tried applying what you suggested in Item 3, but it made me confused. Please take a look at the updated version.

doc/contributing/docs/style.rst
```diff
@@ -73,22 +166,67 @@ aren't familiar with the
...
+    -   The items are UI labels, headings, subheadings, or strings.
+
+*   Any item by itself is a complete sentence.
+
+The above rules come from the
+`Microsoft style guide <https://docs.microsoft.com/en-us/style-guide/scannable-content/lists>`__.
+
+Please don't use commas or semicolons at the end of list items---it's
+either a period or nothing.
+Don't add redundant conjunctions like "and"/"or" before the last list item.
```
Totktonada, line 210: Hm. I do in https://github.com/tarantool/tarantool/pull/6644/files (line 22), because for me it reads a bit more friendly. Should I remove 'and' here?

doc/contributing/docs/style.rst
```diff
@@ -73,22 +166,67 @@ aren't familiar with the
...
+^^^^^^
+
+For the text in cells, use periods or other end punctuation
+only if the cells contain complete sentences or a mixture of fragments and sentences.
+(This is also a
+`Microsoft guideline <https://docs.microsoft.com/en-us/style-guide/scannable-content/tables#punctuation>`__
+for the English language.)
+
+Besides, make sure that your table punctuation is consistent---either
+all similar list/table items end with a period or they all don't.
```
Totktonada, line 223: 'Your table punctuation' assumes a column here? Can we say that a punctuation is a column is the same as for a list?

doc/contributing/docs/style.rst
Totktonada: Agreed to use 'more'.

Totktonada: @patiencedaur said that I should. Replaced so.

doc/contributing/docs/style.rst
patiencedaur: What about the case where an item by itself isn't a complete sentence, but it forms a complete sentence when combined with the sentence preceding the list? Looks like it's missing here.

Totktonada: We discussed it privately and agreed that both cases looks the same, so there is no sense to make the instruction more complex. @patiencedaur updated the instruction, it looks okay for me (in the sense that I can understand it and follow). So I'll resolve the discussion.

PR #2540, opened 2022-01-11
Fixing "space_object.index" broken link

### LINKED ISSUE #2387: [1pt] Fix a link
https://www.tarantool.io/ru/doc/1.10/reference/reference_lua/box_space/

изображение

arctic-dreamer:
Making minor updates in this article just to see if the link fixes itself, as everything works fine in the local autobuild.

Fixes #2387


arctic-dreamer:
The problem is in 1.10, will create another PR

PR #2623, opened 2022-01-24
Add a guideline for using dashes

patiencedaur:
- Types of dashes
- Dash formatting in RST

### DIFF
--- a/doc/contributing/docs/style.rst
+++ b/doc/contributing/docs/style.rst
@@ -174,12 +174,38 @@ Spelling and punctuation
 US vs British spelling
 ~~~~~~~~~~~~~~~~~~~~~~
 
-We use the US English spelling.
+Use the US English spelling.
 
 Check your spelling and punctuation
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Consider checking spelling, grammar, and punctuation with special tools.
+
+Typographic characters
+~~~~~~~~~~~~~~~~~~~~~~
+
+Use these rules (dashes):
+
+There is no need to insert special dashes, quotation marks, or apostrophes. 
+The documentation :doc:`builder </contributing/docs/build/>`
+renders specific character sequences in the source into correct typographic characters.
+
+Use these sequences:
+
+..  container:: table
+
+    ..  list-table::
+        :header-rows: 1
+
+        *   -   Source
+            -   Replaced by 
+
+        *   -   ``---``
+            -   Em dash
+        
+
+Sometimes it is necessary to escape characters
+
 
 Ending punctuation in lists and tables
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/doc/contributing/docs/style.rst
+++ b/doc/contributing/docs/style.rst
@@ -1,4 +1,3 @@
-
 Language and style
 ==================
 
@@ -193,13 +192,14 @@ Type three hyphens (``---``) to insert an em dash (---)
 and two hyphens (``--``) for an en dash (--).
 The following rules apply:
 
-*   The shorter **en dash** ``--`` is used to mark ranges (for example, 4--16 GB).
 *   The longer **em dash** ``---`` is used to separate extra information or mark a break in a sentence.
+*   The shorter **en dash** ``--`` is used to mark ranges (for example, 4--16 GB or Dover--Calais crossing).
 
 Don't use a single hyphen as a dash. Don't add spaces on either side of a dash. 
 
 When indicating a range like ``code element 1``\--``code element 2``, escape the series of hyphens using
-`character-level inline markup <https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#character-level-inline-markup-1>`_:
+`character-level inline markup <https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#character-level-inline-markup-1>`_. 
+Otherwise, the RST interpreter will perceive the dash as part of the RST syntax:
 
 ..  code-block:: rst
 
--- a/doc/contributing/docs/style.rst
+++ b/doc/contributing/docs/style.rst
@@ -184,16 +184,16 @@ Consider checking spelling, grammar, and punctuation with special tools.
 Dashes
 ~~~~~~
 
-There is no need to insert special symbols like dashes, quotation marks, or apostrophes.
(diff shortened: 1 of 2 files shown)


doc/contributing/docs/style.rst
```diff
@@ -174,13 +174,38 @@ Spelling and punctuation
...
+There is no need to insert special symbols like dashes, quotation marks, or apostrophes.
+The documentation :doc:`builder </contributing/docs/build/>`
+renders specific character sequences in the source into correct typographic characters. 
+
+Type three hyphens (``---``) to insert an em dash (---)
+and two hyphens (``--``) for an en dash (--).
+The following rules apply:
+
+*   The shorter **en dash** ``--`` is used to mark ranges and with the meaning "to"
+    in phrases like "Dover--Calais crossing".
```
NickVolynkin, line 196: We hardly ever have such cases in our docs, but we can have number ranges, like 4–16 gigabytes.

doc/contributing/docs/style.rst
```diff
@@ -174,13 +174,38 @@ Spelling and punctuation
...
+*   The longer **em dash** ``---`` is used to separate extra information or mark a break in a sentence.
+
+Don't use a single hyphen as a dash. Don't add spaces on either side of a dash. 
+
+When indicating a range like ``code element 1``\--``code element 2``, escape the series of hyphens using
+`character-level inline markup <https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#character-level-inline-markup-1>`_:
+
+..  code-block:: rst
+
+    ``box.begin()``\--``box.commit()``
```
NickVolynkin, line 206: what's the use case for this syntax? range between two functions?

doc/contributing/docs/style.rst
```diff
@@ -174,13 +174,38 @@ Spelling and punctuation
...
 
 Check your spelling and punctuation
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Consider checking spelling, grammar, and punctuation with special tools.
 
+Dashes
+~~~~~~
+
+There is no need to insert special symbols like dashes, quotation marks, or apostrophes.
```
NickVolynkin, line 187: It's less about need and more about consistency in the sources and in the translation memory.

NickVolynkin: And we can say the same about quotes

doc/contributing/docs/style.rst
patiencedaur: This is a real example:
image

doc/contributing/docs/style.rst
```diff
@@ -174,13 +174,38 @@ Spelling and punctuation
...
+renders specific character sequences in the source into correct typographic characters.
+
+Type three hyphens (``---``) to insert an em dash (---)
+and two hyphens (``--``) for an en dash (--).
+The following rules apply:
+
+*   The shorter **en dash** ``--`` is used to mark ranges (for example, 4--16 GB).
+*   The longer **em dash** ``---`` is used to separate extra information or mark a break in a sentence.
+
+Don't use a single hyphen as a dash. Don't add spaces on either side of a dash.
```
Totktonada, line 199: As I see, both styles are in use in English (with and without surrounding whitespaces). In my opinion, em dash without surrounding spaces looks odd. So I'm curious, what is the reason behind choosing the style without spaces.

doc/contributing/docs/style.rst
```diff
@@ -174,13 +174,38 @@ Spelling and punctuation
...
+across all Tarantool documentation in a single language.
+This is because the documentation :doc:`builder </contributing/docs/build/>`
+renders specific character sequences in the source into correct typographic characters.
+
+Type three hyphens (``---``) to insert an em dash (---)
+and two hyphens (``--``) for an en dash (--).
+The following rules apply:
+
+*   The shorter **en dash** ``--`` is used to mark ranges (for example, 4--16 GB).
+*   The longer **em dash** ``---`` is used to separate extra information or mark a break in a sentence.
```
Totktonada, line 197: We describe it as em-en in the paragraph as en-em in the list (I'm about the order), it may be a bit confusing.

doc/contributing/docs/style.rst
```diff
@@ -174,13 +174,38 @@ Spelling and punctuation
...
+Special symbols like dashes, quotation marks, and apostrophes look the same
+across all Tarantool documentation in a single language.
+This is because the documentation :doc:`builder </contributing/docs/build/>`
+renders specific character sequences in the source into correct typographic characters.
+
+Type three hyphens (``---``) to insert an em dash (---)
+and two hyphens (``--``) for an en dash (--).
+The following rules apply:
+
+*   The shorter **en dash** ``--`` is used to mark ranges (for example, 4--16 GB).
```
Totktonada, line 196: Numerical ranges or ranges of any nature? As I see from the example below, the latter, but it is not stated explicitly.

doc/contributing/docs/style.rst
```diff
@@ -174,13 +174,38 @@ Spelling and punctuation
...
+*   The longer **em dash** ``---`` is used to separate extra information or mark a break in a sentence.
+
+Don't use a single hyphen as a dash. Don't add spaces on either side of a dash. 
+
+When indicating a range like ``code element 1``\--``code element 2``, escape the series of hyphens using
+`character-level inline markup <https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#character-level-inline-markup-1>`_:
```
Totktonada, line 202: Is it to prevent separating with a whitespace from the blocks? I would clarify it in the text.

Totktonada (approved):
I have no real objections, but you can consider some thoughts below.

doc/contributing/docs/style.rst
patiencedaur: Originally, the em dash was surrounded by typographic marks known as "thin spaces". But teaching our users more about types of spaces (so many of them) is just cruel IMO. We can either drop the spaces or make them thicker. I chose the former option because this way the usage is similar to the en dash, so the user has less to think about.

doc/contributing/docs/style.rst
patiencedaur: Good point

doc/contributing/docs/style.rst
Totktonada: Couldn't it be solved automatically?

@NickVolynkin ?

PR #2641, opened 2022-01-28
Update the Lua style guide

### LINKED ISSUE #753: [2pt] Update Lua style guide
https://www.tarantool.io/en/doc/2.1/dev_guide/lua_style_guide/#error-handling

This example will likely result in troubles with box.NULL:
```
local data, err = foo()
if not data
    return nil, err
end
```

To replace with this good practice:

Never do 
```
if x then
```
and 
```
if not x then
```
as you'll have troubles with box.NULL then.
Always do 
```
if x~= nil then
```
and 
```
if x == nil then
```
instead.

patiencedaur:
Resolves #753 

This is not a primary task for the doc team. To save time, I only proofread parts where I understood the meaning.

### DIFF
--- a/doc/dev_guide/lua_style_guide.rst
+++ b/doc/dev_guide/lua_style_guide.rst
@@ -258,7 +258,7 @@ Zen of Python is good; understand it and use wisely:
 
     .. code-block:: lua
 
-        if thing then
+        if thing ~= nil then
             -- ...stuff...
         end
         function derp()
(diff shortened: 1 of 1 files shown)


Totktonada (approved):
I'm okay regarding the suggestion to avoid implicit casting to boolean in an if condition. I would prefer to don't look over all changes in this PR.

PR #2676, opened 2022-02-10
Only run push-translation-sources on click

patiencedaur:
Will it work?


artembo (approved):
It should work. Looks good to me.

PR #2687, opened 2022-02-16
Document ability of passing several URIs in different ways

veod32:


doc/reference/configuration/index.rst
```diff
@@ -97,10 +99,39 @@ If username:password is omitted, then 'guest' is assumed. Some examples:
...
+Starting from version 2.10.0, a user can open several listening iproto sockets on a Tarantool instance
+and, consequently, can specify several URIs in the configuration parameters such as :ref:`box.cfg.listen <cfg_basic-listen>` and :ref:`box.cfg.replication <cfg_replication-replication>`.
+
+In addition to specifying URI as a number or a string, a number of other ways to pass URI values is available starting from Tarantool 2.10.0:
+
+*   As a string with one or several URIs separated by commas (provides backward compatibility)
+
+    code-block:: lua
+
+    box.cfg { listen = "127.0.0.1:3301, /unix.sock, 3302" }
```
EvgenyMekhanik: I would write here about all the methods with an indication of the recommended ones.

In addition to specifying URI as a number or a string, a number of other ways to pass URI values with parameters are available starting from Tarantool 2.10.0:
```lua
-- Single URI, passed as before
box.cfg({ listen = "/tmp/unix.sock" })
-- Single URI, with query paramters
box.cfg({ listen = "/tmp/unix.sock?q1=v1&q2=v2" })
-- Several URIs with parameters in one string, separated by commas
box.cfg({ listen = "/tmp/unix.sock_1?q=v, /tmp/unix.sock_2?q=v" })
-- Single URI passed in table, with additional parameters, passed
-- in "params" table. This parameters overwrite parameters from
-- URI string (q1 = "v2" in example below).
box.cfg({ listen = {"/tmp/unix.sock?q1=v1", params = {q1 = "v2"}} })
-- Several URIs passed in table with default parameters, passed
-- in "default_params" table, which are used for parameters, which
-- not specified for URI (q3 parameter with "v3" value corresponds
-- to all URIs, and used if there is no such parameter in URI).
box.cfg({ listen = {
        "/tmp/unix.sock_1?q1=v1",
        { uri = "/tmp/unix.sock_2", params = { q2 = "v2" } },
        default_params = { q3 = "v3" }
    }
}) 
```
Recommended ways to pass URIs with parameters:
```lua
-- Common way for several URIs with parameters
box.cfg ({  listen = {
        {uri = "127.0.0.1:3301", params = {q1 = "v1"}},
        {uri = "/unix.sock", params = {q2 = "v2"}},
        {uri = 3302, params = {q3 = "v3"}}
    }
})
-- For single URI with parameter is also ok
box.cfg ({ listen = {
    uri = 'localhost:3301',
    params = {
        transport = "ssl",
        ssl_ca_file = "/path_to_ca_file",
        ssl_cert_file = "/path_to_cert_file",
        ssl_key_file = "/path_to_key_file"
    }
}})
```

doc/reference/configuration/index.rst
```diff
@@ -99,40 +104,76 @@ If ``username:password`` is omitted, then 'guest' is assumed. Some examples:
...
-            {uri = "/unix.sock"},
-            {uri = 3302}
+        box.cfg { listen = {
+                {uri = "127.0.0.1:3301"},
+                {uri = "/unix.sock"},
+                {uri = 3302}
+            }
         }
+
+Also, you can specify additional URI parameters. The recommended syntax:
```
EvgenyMekhanik, line 155: Here i would indicate all the methods as in the previous comment with the recommended ones

EvgenyMekhanik:
I tarantool commit message letght shoud be less that 80 symbols. I don't know is it problem for doc PRs?

veod32:
As for commit message length, there's no problem.

doc/reference/configuration/index.rst
veod32: Examples are added.

EvgenyMekhanik (approved):

doc/reference/configuration/cfg_replication.rst
```diff
@@ -32,9 +32,13 @@
...
     :extsamp:`box.cfg{ replication = { {*{'uri1'}*}, {*{'uri2'}*} } }`
 
-    If one of the URIs is "self" -- that is, if one of the URIs is for the
-    instance where ``box.cfg{}`` is being executed on -- then it is ignored.
-    Thus it is possible to use the same ``replication`` specification on
+    ..  note::
+
+        Starting from version 2.10.0, there is a number of other ways for specifying several URIs. See :ref:`syntax examples <index-uri-several>`.
+
+    If one of the URIs is "self"---that is, if one of the URIs is for the
```
EvgenyMekhanik, line 39: I don't really understand this phrase

veod32: Can be fixed in a separate ticket because it's from some old descritption.

PR #2716, opened 2022-02-23
Document setting box.cfg parameters via environment variables

veod32:


Totktonada:
I suggest to add environment variable names into box.cfg parameters description explicitly for user's convenience. Looks at the etcd documentation for inspiration: https://etcd.io/docs/v3.4/op-guide/configuration/.

veod32:
> I suggest to add environment variable names into box.cfg parameters description explicitly for user's convenience. Looks at the etcd documentation for inspiration: https://etcd.io/docs/v3.4/op-guide/configuration/.

Good suggestion. Currently, there're around 75 parameters in the config reference. If we're going to do this update, I think it's better to do it as a separate task.

Totktonada:
Okay. Let's split it as it is convenient for you.

veod32:
> Okay. Let's split it as it is convenient for you.

#2722

doc/reference/configuration/index.rst
```diff
@@ -296,66 +300,106 @@ dynamic, that is, they can be changed at runtime by calling ``box.cfg{}``
...
+
+..  _box-cfg-params-prior:
+
+Methods of setting and priorities
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Tarantool configuration parameters can be specified in different ways.
+The priority of parameter sources is the following, from higher to lower:
+
+*   ``box.cfg{}`` call
```
Totktonada, line 313: `box.cfg{}` call options, maybe? To make it clear and strictly separated from defaults.

doc/reference/configuration/index.rst
```diff
@@ -296,66 +300,106 @@ dynamic, that is, they can be changed at runtime by calling ``box.cfg{}``
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Starting from version :doc:`2.8.1 </release/2.8.1>`, you can specify configuration parameters via special environment variables.
+The name of a variable should have the following pattern: ``TT_<NAME>``,
+where ``<NAME>`` is the uppercase name of the corresponding :ref:`box.cfg parameter <box-cfg-params-ref>`.
+
+For example:
+
+* ``TT_LISTEN``---corresponds to the ``box.cfg.listen`` option.
+* ``TT_REPLICATION``---corresponds to the ``box.cfg.replication`` option.
```
Totktonada, line 330: It worth to add one example of an option, whose name is constructed from several words.

Totktonada (approved):
LGTM. Thanks!

doc/reference/configuration/index.rst
veod32: Corrected

doc/reference/configuration/index.rst
veod32: Added ``memtx_dir`` as an example of such options.

PR #2732, opened 2022-03-02
Choose a more fitting instance file name

### LINKED ISSUE #1080: feedback: Instance configuration
> For example, **|/etc/tarantool/instances.enabled/my_app.lua|** (here we load
> my_app.lua module and make a call to start() func<…>

https://www.tarantool.io/en/doc/2.2/book/admin/instance_config/

Example instance file name, my_app.lua, is the same as the application file name that can be confusing. Better to rename the instance file in the example: say, into init.lua

patiencedaur:
Resolves #1080


Totktonada:
Answered in the linked issue.

PR #2767, opened 2022-03-21
Remove `code` formatting in headers

lastoCHka42:
Resolves tarantool/tarantool.io#964


patiencedaur (approved):
It turned out that code formatting in headers breaks navigation in footers.

PR #2788, opened 2022-04-01
Test whether external .md's are easy to integrate in the doc

patiencedaur:
Tarantool developers write documentation in .md. We need to research whether it's easy to integrate in the documentation website, where all the content is in .rst.


patiencedaur:
@NickVolynkin, what else do I have to take into account except for `.rst` files? Will any of the new modules be integrated in singlehtml, pdf, or other more exotic representations?

PR #2800, opened 2022-04-11
Add an explicit requirement to capitalize Tarantool

patiencedaur:
We have a "Writing about code" section, but the user may not be sure when to consult it. This is an important product-related requirement, so I decided to make it explicit and provide a link to the section.

### DIFF
--- a/doc/contributing/docs/style.rst
+++ b/doc/contributing/docs/style.rst
@@ -170,6 +170,14 @@ For this reason, it's best to avoid using them.
 Spelling and punctuation
 ------------------------
 
+Tarantool capitalization
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+The word "Tarantool" is capitalized because it's a product name.
+The only context where it can start with a lowercase "t" is code.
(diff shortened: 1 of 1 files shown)


doc/contributing/docs/style.rst
```diff
@@ -170,6 +170,14 @@ For this reason, it's best to avoid using them.
 Spelling and punctuation
 ------------------------
 
+Tarantool capitalization
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+The word "Tarantool" is capitalized because it's a product name.
+The only context where it can start with a lowercase "t" is code.
```
Totktonada, line 177: Technically it makes legal to use the lowercased variant in a comment in the code. I would say, when a shell commmand / executable assumed.

patiencedaur: If the comment becomes part of the documentation (say, via `ldoc`), then this rule is applied. Comments that don't end up in the documentation are covered by the code style guide, not by this guide.

PR #2856, opened 2022-05-02
chore: Coherent naming

puria:


NickVolynkin (approved):
This is now consistent with function names in other code fragments.

patiencedaur:
Thanks for the patch!

PR #2897, opened 2022-05-27
Democratize the use of dashes

patiencedaur:
Previously, we distinguished between `--` and `---` in various cases. Now we decided to stick with `--` – it's simpler. (But we still deprecate the use of a hyphen as a dash.)

The overall tendency to simplify dash usage in documentation may be connected to the diversity of the audiences. The Tarantool docs are not read by British academia, so it's a good decision to simplify the rules of dash usage.


locale/ru/LC_MESSAGES/reference/reference_lua/net_box.po
```diff
@@ -54,7 +54,7 @@ msgid ""
 "network socket, but each fiber gets back a correct response. Reducing the "
 "number of active sockets lowers the overhead of system calls and increases "
 "the overall server performance. However for some cases a single connection "
-"is not enough---for example, when it is necessary to prioritize requests or "
+"is not enough -- for example, when it is necessary to prioritize requests or "
```
Totktonada, line 57: Are we start to add spaces around a dash in this case? Should we give certain rule in the style guide?

patiencedaur: The style guide provides for it. See line 200 in `style.rst`: "Add spaces on both sides of the dash."

PR #2901, opened 2022-05-30
Rewrite the example for http 1.1.1 API

### LINKED ISSUE #2673: feedback: Cookbook recipes | HTTP v2 is gone
The example uses http v2 API, which was dropped. Correct the example so it's based on the 1.1.1 API.

---

> <…>erver').new(nil, 8080) -- listen *:8080
> local router = **require('|http.router|').new({charset** = "utf8"})
> server:set_router(router)
> router:rout<…>

https://www.tarantool.io/en/doc/latest/book/app_server/cookbook/#http-server-lua

http v2 is gone.

patiencedaur:
Resolves #2673


doc/book/app_server/cookbook.rst
```diff
@@ -776,10 +776,8 @@ to turn Tarantool into a web server.
         return self:render{ json = { ['Your-IP-Is'] = self:peer().host } }
```
Totktonada: `self:peer().host` should be replaced with `self.peer.host` for http v1. Let's start the script and test it with curl:

```shell
$ curl http://localhost:8080
```

You'll see the error with current variant of the code.

Totktonada: Seems resolved.

PR #2902, opened 2022-05-31
Update version and url in configs

p7nov:


patiencedaur (approved):
Thank you! Could you please tell about this piece of config and why it is important at our educational meeting? This would be insightful for the team.

PR #2916, opened 2022-06-02
Explain what to do with optional parameters

patiencedaur:
A slight change to the rules for documenting API function parameters.

The `(optional)` mark for each parameter looks redundant, because we already see in the function declaration which arguments are optional and which are required.


Totktonada (approved):
I see the only usage of the `(optional)` in the parameter description [here](https://www.tarantool.io/en/doc/latest/reference/reference_lua/swim/#lua-function.swim_member_object.on_member_event). So it was not used in fact. I'm okay to make the guide actual.

PR #2931, opened 2022-06-08
Replace mentions of betas with 2.10 release

### LINKED ISSUE #2930: Remove pre-release version from docs after release
Descriptions of new features added in pre-release versions (alpha, beta, rc) often contain introductions like "since 2.10.0-beta1". They help us highlight new features, promote them, and stimulate users to try new versions.
After the release version is out, such intros become outdated (we don't expect users to use pre-release versions anymore) and even confusing (users may think that the release version is not available yet).

Taking into account that we have a single doc for all versions of a major release (1.*, 2.*), it's better to leave such intros and just remove the pre-release version postfix. For example:
> Since 2.10.1-beta1 >>> Since 2.10

This will let users know what features are available in their version and can also stimulate upgrades.

**Reviewer:** @ documentarian name

# Definition of done

- [ ] replace mentionings of pre-release versions and accompanying links with the release version.

patiencedaur:
Resolves #2930


doc/book/replication/repl_leader_elect.rst
```diff
@@ -83,11 +83,6 @@ between each node pair so as it would be the full mesh topology. This is needed
 because election messages for voting and other internal things need direct
 connection between the nodes.
 
-Also, if election is enabled on the node, it won't replicate from any nodes except
-the newest leader. This is done to avoid the issue when a new leader is elected,
-but the old leader has somehow survived and tries to send more changes
-to the other nodes.
-
```
p7nov: Is this deletion intended? And the other one in this file too.
I'm asking because it's not related to the issue.

p7nov (approved):
LGTM, just check the question about repl_leader_elect.rst.

doc/book/replication/repl_leader_elect.rst
patiencedaur: Oops, my mistake. Will revert this change.

PR #2969, opened 2022-06-22
[Raft] Document the new fencing feature

### LINKED ISSUE #2597: [3pt] Raft: Document the fencing mechanism
**Product:** Tarantool
**Since:** 2.10
**Dev issue:** https://github.com/tarantool/tarantool/issues/6661
**Root document:** https://www.tarantool.io/en/doc/latest/book/replication/repl_leader_elect/#leader-election-process
**SME:** @ sergepetrenko 

# Details

[input from dev issue]
A leader doesn't track its connectivity to the rest of the cluster.
Once it is elected, it stays in the leader position until receives a newer term from some remote node.
In case the leader is partitioned from the rest of the cluster, it'll remain to be a leader as long as the cluster is split up. Even if the rest of the cluster have already elected a new leader.

After implementing the fencing mechanism, the leader will track follower connections and step off to being a follower as soon as it loses a quorum of connections.

**TDB**  details of the fencing mechanism implementation in Tarantool

# ToDo

- Update the description of the [leader election process](https://www.tarantool.io/en/doc/latest/book/replication/repl_leader_elect/) -- the process will be different after implementation of the fencing functionality.
- Possibly, update the description of the [Raft properties](https://www.tarantool.io/en/doc/latest/book/replication/repl_leader_elect/#configuration) if something is going to change. The current view is that the already existing options will be used without any modification.

veod32:


doc/book/replication/repl_leader_elect.rst
```diff
@@ -96,6 +96,18 @@ between each node pair so as it would be the full mesh topology. This is needed
 because election messages for voting and other internal things need direct
 connection between the nodes.
 
+.. _repl_leader_elect_fencing:
+
+In the classic Raft algorithm, a leader doesn't track its connectivity to the rest of the cluster.
+Once the leader is elected, it considers itself in the leader position until receiving a new term from some of the cluster nodes.
```
grafin: "from another cluster node" sounds better to me.

doc/book/replication/repl_leader_elect.rst
```diff
@@ -96,6 +96,18 @@ between each node pair so as it would be the full mesh topology. This is needed
...
+This can lead to the split situation if the other nodes elect a new leader upon losing the connectivity to the previous one.
+
+The issue is resolved in Tarantool version :doc:`2.10.0 </release/2.10.0>` by introducing the leader *fencing* mode.
+The mode can be switched on and off by the :ref:`election_fencing_enabled <repl_leader_elect_config>` configuration parameter.
+When the fencing is on, the leader will resign its leadership if it has less than the :ref:`replication_synchro_quorum <repl_leader_elect_config>`
+of alive connections to the cluster nodes. The resigning leader receives the follower status in the current election term and becomes a read-only node.
```
grafin: "becomes read-only." sounds better to me.

doc/reference/configuration/cfg_replication.rst
```diff
@@ -518,11 +519,30 @@
...
+
+..  _cfg_replication-election_fencing_enabled:
+
+..  confval:: election_fencing_enabled
+
+    Since version :doc:`2.10.0 </release/2.10.0>`.
+
+    Switches on and off the :ref:`leader fencing mode <repl_leader_elect_fencing>` that affects the leader election process.
+    When the parameter is set to ``true``, the leader will resign its leadership if it has less than the :ref:`replication_synchro_quorum <cfg_replication-replication_synchro_quorum>`
+    of alive connections to the cluster nodes. The resigning leader receives the follower status in the current election term and becomes a read-only node.
```
grafin: Same as above.

doc/reference/configuration/cfg_replication.rst
```diff
@@ -518,11 +519,30 @@
...
+    Fencing applies to the instances that have :ref:`election_mode <election_mode>` set to "candidate" or "manual".
+
+    If the fencing is enabled on the current leader and it doesn't have the :ref:`replication_synchro_quorum <cfg_replication-replication_synchro_quorum>` of alive connections,
+    the leader resigns its leadership immediately.
```
grafin: I'm not sure on "immediately". It will resign as soon as it will notice that those connections are dead (it's not the same as them being dead).

grafin (changes requested):
Thanks for the patch! The only real problem is with using "immediately" when describing fencing. Other comments are just my opinion and you may leave them as is.

doc/reference/configuration/cfg_replication.rst
veod32: Decided to remove the phrase because the leader behaviour is not different from the description above.

grafin (approved):
Thanks for very quick fixes. LGTM!

PR #2983, opened 2022-06-27
Update the leader election article with the info on resolving the split-brain

### LINKED ISSUE #2954: new error type: ER_SPLIT_BRAIN
**Related dev. issue(s):** https://github.com/tarantool/tarantool/issues/5295

Part of the [Errors](https://github.com/tarantool/doc/issues/2960) **epic**

**Product:** Tarantool
**Since:** 2.10
**Audience/target:** 
**Root document:** https://www.tarantool.io/en/doc/latest/book/replication/repl_leader_elect/
**SME:** @ sergepetrenko, @ grafin 

---

If for some reason the cluster had 2 leaders working independently (for
example, user has mistakenly lovered the quorum below N / 2 + 1), then
once such leaders and their followers try connecting to each other, they
will receive the ER_SPLIT_BRAIN error, and the connection will be
aborted. This is done to preserve data integrity. Once the user notices
such an error he or she has to manually inspect the data on both the
split halves, choose a way to restore the data, and rebootstrap one of
the halves from the other.
Requested by @cyrillos in https://github.com/tarantool/tarantool/commit/af7d703fd6cec8d3d7637ea2c6e7782fc4975fb0.

veod32:


doc/book/replication/repl_leader_elect.rst
```diff
@@ -108,6 +108,18 @@ When the fencing is on, the leader resigns its leadership if it has less than th
...
 Fencing applies to the instances that have the :ref:`election_mode <repl_leader_elect_config>` set to "candidate" or "manual".
 
+.. _repl_leader_elect_splitbrain:
+
+There can still be a situation when a replica set has two leaders working independently (so called *split-brain*).
+It can happen, for example, if a user mistakenly lowered the :ref:`replication_synchro_quorum <repl_leader_elect_config>` below ``N / 2 + 1``.
+In this situation, to preserve the data integrity, if an instance detects the split-brain anomaly in the incoming replication data,
+it breaks the connection with the instance sending the data and writes the ``ER_SPLIT_BRAIN`` error in the log.
+
+Once noticing this error, a user has to manually inspect the data on both instances having the leader status,
```
sergepetrenko: It's not necessary that any nodes in the split halves will be leaders.

The user will see that there are two sets of nodes, which do not intersect, and any node from one set is disconnected from any node from the other set with ER_SPLIT_BRAIN error.

Then the user might choose any representative from each of the sets, and inspect the data on it.

The sets essentially represent an old leader and all the nodes which followed that leader and believe that it's the source of truth.

veod32: Re-written. Please take a look.

sergepetrenko (approved):
Thanks! Looks good to me.

PR #2991, opened 2022-06-30
Add txn_isolation reference

### LINKED ISSUE #2793: Document transaction isolation settings
Root document: https://www.tarantool.io/en/doc/latest/book/box/atomic/
https://www.tarantool.io/en/doc/latest/reference/
SME: @ Mons @ alyapunov 

This ticket may require adding a new page. 
Also, we may need to split the Transactions page into separate sections. Related issue: #2812

# Details
This may require making graphs, as the topic is hard to understand from mere writing.

---

All transactions in tarantool see only committed changes, but is
a loose concept since the commit of a transaction is long-term
act and there's a dilemma of choosing a timepoint at which the
changes must be seen by other transactions.

When enabled, mvcc engine allows some options on this field.
To be  specific, we introduced two levels of isolation:
* read-committed: all transaction that have started committing
(box.commit() was called) are visible.
* read-confirmed: all transaction that have finished committing
(box.commit() returnd) are visible.

The first level is good for RW transactions since it allows to
avoid some conflicts, while the second is good for RO transactions
since it allows to get tuples that a definitely was persistent on
disk.

There's also a default option that allows mvcc engine to decide:
* best-effort: do the best depending on which actions the
transaction has made.

For autocommit transaction (one-statement actions without explicit
box.begin/commit calls) an obvious rule is applied: RO (select,
get, count etc) transactions are done with read-confirmed; all
other (replace, delete, etc) - read-committed.

If a transaction has an explicit box.begin call, the level can be
specified in the following way:
box.begin({tnx_isolation = 'default'})
box.begin({txn_isolation = 'read-committed'})
box.begin({txn_isolation = 'read-confirmed'})
box.begin({tnx_isolation = 'best-effort'})

One can similarly set the level in net.box stream begin() method.

If not specified (or if 'default' was passed) the level is taken
from configuration. One can set default level in box.cfg:
box.cfg({default_txn_isolation = 'read-committed'})
box.cfg({default_txn_isolation = 'read-confirmed'})
box.cfg({default_tnx_isolation = 'best-effort'})

The default option is 'best-effort'.

In iproto the level is specified by IPROTO_TXN_ISOLATION key in
the body of IPROTO_BEGIN request. The value is the one of the
following enum values:
enum txn_isolation_level {
	/** Take isolation level from global default level. */
	TXN_ISOLATION_DEFAULT = 0,
	/** Allow to read committed, but not confirmed changes. */
	TXN_ISOLATION_READ_COMMITTED = 1,
	/** Allow to read only confirmed changes. */
	TXN_ISOLATION_READ_CONFIRMED = 2,
	/** Determine isolation level automatically. */
	TXN_ISOLATION_BEST_EFFORT = 3,
};

NO_TEST=no changes
Closes #6930
Requested by @alyapunov in https://github.com/tarantool/tarantool/commit/0b1f9b0f731d9c04efb9aa9e7272f637728d1653.

p7nov:
Resolves #2793

### DIFF
--- a/doc/book/box/atomic/txn_mode_mvcc.rst
+++ b/doc/book/box/atomic/txn_mode_mvcc.rst
@@ -60,13 +60,6 @@ option to enable it via ``box.cfg``.
 Setting the transaction isolation level
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-..  note::
-
-    For autocommit transactions (actions with a statement without explicit ``box.begin/commit`` calls) 
-    there is an obvious rule: read-only transactions (for example, ``select``) are performed with ``read-confirmed``; 
-    all others (for example, ``replace``) with ``read-committed``.
-
-
 The transaction manager has three options for the :ref:`transaction isolation level <transaction_model_levels>`:
 
 *   ``best-effort`` (default)
@@ -79,8 +72,7 @@ Using ``best-effort`` as the default option allows MVCC to consider the actions 
 independently and determine the best :ref:`isolation level <transaction_model_levels>` for them.
 It increases the probability of successful completion of the transaction and helps to avoid possible conflicts.
 
-To set the default isolation level with the other option, for example, 
-to ``read-committed``, use the following command:
+To set another default isolation level, for example, ``read-committed``, use the following command:
 
 ..  code-block:: lua
 
@@ -97,8 +89,12 @@ to the one set in ``box.cfg``.
 
 ..  note::
 
-    You can set the isolation level in the net.box :ref:`stream:begin() <net_box-stream_begin>` method
-    and :ref:`IPROTO_BEGIN <box_protocol-begin>` binary protocol request.
+    For autocommit transactions (actions with a statement without explicit ``box.begin/commit`` calls)
+    there is an obvious rule: read-only transactions (for example, ``select``) are performed with ``read-confirmed``;
+    all others (for example, ``replace``) -- with ``read-committed``.
+
+You can also set the isolation level in the net.box :ref:`stream:begin() <net_box-stream_begin>` method
+and :ref:`IPROTO_BEGIN <box_protocol-begin>` binary protocol request.
 
 
 Choosing the better option depends on whether you have conflicts or not. 
--- a/doc/book/box/atomic/txn_mode_mvcc.rst
+++ b/doc/book/box/atomic/txn_mode_mvcc.rst
@@ -86,7 +86,7 @@ to ``read-committed``, use the following command:
 
     box.cfg{default_txn_isolation = 'read-committed'}
 
-You can also set an isolation level for specific transactions in their ``box.begin()`` calls.
+You can also set an isolation level for specific transactions in their ``box.begin()`` calls:
 
 ..  code-block:: lua
 
@@ -98,7 +98,7 @@ to the one that is set in ``box.cfg``.
 ..  note::
 
     You can set the isolation level in the net.box :ref:`stream:begin() <net_box-stream_begin>` method
-    and :ref:`IPRORO_BEGIN <box_protocol-begin>` binary protocol request.
+    and :ref:`IPROTO_BEGIN <box_protocol-begin>` binary protocol request.
 
 
 Choosing the better option depends on whether you have conflicts or not. 
--- a/doc/book/box/atomic/txn_mode_mvcc.rst
+++ b/doc/book/box/atomic/txn_mode_mvcc.rst
@@ -57,8 +57,8 @@ option to enable it via ``box.cfg``.
 
 ..  _txn_mode_mvcc-options:
 
-Transaction manager options
-~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Setting the transaction isolation level
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 ..  note::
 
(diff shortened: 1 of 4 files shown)


doc/book/box/atomic/txn_mode_mvcc.rst
```diff
@@ -69,13 +69,13 @@ Transaction manager options
 
 The transaction manager has four options for the transaction isolation level that you can set in ``box-cfg``:
 
-*   ``default``.
+*   ``default``
```
p7nov: @rindblack15 I took the liberty to remove periods according to the [list punctuation guidelines](https://www.tarantool.io/en/doc/latest/contributing/docs/style/#lists). Hope you don't mind :)

doc/book/box/atomic/txn_mode_mvcc.rst
```diff
@@ -69,13 +69,13 @@ Transaction manager options
...
 
-*   ``best-effort``.
+*   ``best-effort``
 
-*   ``read-committed``.
+*   ``read-committed``
 
-*   ``read-confirmed``.
+*   ``read-confirmed``
```
rindblack15, line 70: Let's describe a little bit all the options here, not just the best-effort. Otherwise, it looks like you'll have to guess the rest or read about it on the previous pages.

doc/dev_guide/internals/box_protocol.rst
```diff
@@ -739,12 +739,8 @@ The body is optional and can contain two items:
...
 the transaction will be rolled back automatically.
 
-.. // TODO: add link to transaction isolation docs once they're ready
-
-IPROTO_TXN_ISOLATION is the transaction isolation level. It can take
-the following values:
-
-.. // TODO: provide links to level descriptions
+IPROTO_TXN_ISOLATION is the :ref:`transaction isolation level <txn_mode_mvcc-options>`.
+It can take the following values:
```
rindblack15, line 744: It's not very clear what a global default level is? Or why should I choose the automatic level detection?

doc/book/box/atomic/txn_mode_mvcc.rst
```diff
@@ -85,15 +85,15 @@ of successful completion of the transaction and helps to avoid possible conflict
 To set the default isolation level with the other option, for example, 
 to ``read-committed``, use the following command:
 
-..  code-block:: 
+..  code-block:: lua
 
     box.cfg{default_txn_isolation = 'read-committed'}
```
rindblack15, line 80: It would be nice if we could learn from this section that the isolation levels can also be set via IPROTO.

rindblack15 (changes requested):

doc/dev_guide/internals/box_protocol.rst
p7nov: I'd wouldn't expand these brief descriptions because it's just a reference and is already linked to the page with conceptual explanation.

doc/book/box/atomic/txn_mode_mvcc.rst
p7nov: Thanks.
I've rephrased this in such a way that here we only list options without the explanation (it's available under the link before the list). The sentence about `best-effort` explains why it was chosen as a default, not how it works. So this all seems to me pretty consistent.

doc/book/box/atomic/txn_mode_mvcc.rst
p7nov: Thanks, added this.

rindblack15 (approved):

doc/book/box/atomic/txn_mode_mvcc.rst
```diff
@@ -57,8 +57,8 @@ option to enable it via ``box.cfg``.
 
 ..  _txn_mode_mvcc-options:
 
-Transaction manager options
-~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Setting the transaction isolation level
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 ..  note::
```
patiencedaur: The note mentions terms that haven't yet been introduced. Consider moving the note lower.

doc/book/box/atomic/txn_mode_mvcc.rst
```diff
@@ -67,44 +67,43 @@ Transaction manager options
...
-..  code-block:: 
+..  code-block:: lua
 
-    box.cfg{default_txn_isolation = 'read-committed'}
- 
+    box.cfg{txn_isolation = 'read-committed'}
 
-If a transaction has an explicit ``box.begin()`` call, the level can be
-specified as follows:
+You can also set an isolation level for specific transactions in their ``box.begin()`` calls:
```
patiencedaur:
```suggestion
You can also set an isolation level for a specific transaction in its ``box.begin()`` call:
```

doc/book/box/atomic/txn_mode_mvcc.rst
```diff
@@ -67,44 +67,43 @@ Transaction manager options
...
 
-..  code-block:: 
+..  code-block:: lua
 
     box.begin({tnx_isolation = 'best-effort'})
 
+In this case, there is additionally the ``default`` option. It sets the transaction's isolation level
+to the one that is set in ``box.cfg``.
```
patiencedaur:
```suggestion
In this case, you can also use the ``default`` option. It sets the transaction's isolation level
to the one set in ``box.cfg``.
```

doc/book/box/atomic/txn_mode_mvcc.rst
```diff
@@ -67,44 +67,43 @@ Transaction manager options
...
+
 ..  note::
 
-    You can also do this in the net.box :ref:`stream:begin() <net_box-stream_begin>` method.
+    You can set the isolation level in the net.box :ref:`stream:begin() <net_box-stream_begin>` method
+    and :ref:`IPROTO_BEGIN <box_protocol-begin>` binary protocol request.
```
patiencedaur: I'm not sure that you can interfere with the IPROTO_BEGIN request per se. It is under the hood of `stream:begin()`.

doc/book/box/atomic/txn_mode_mvcc.rst
```diff
@@ -67,44 +67,43 @@ Transaction manager options
...
 
 
 Choosing the better option depends on whether you have conflicts or not. 
 If you have many conflicts, you should set the different options or use
```
patiencedaur:
```suggestion
If you have many conflicts, you should set a different option or use 
```

patiencedaur (changes requested):

doc/book/box/atomic/txn_mode_mvcc.rst
p7nov: If I read the [binary protocol intro](https://www.tarantool.io/en/doc/latest/dev_guide/internals/box_protocol/) right, then yes, it's possible.

doc/book/box/atomic/txn_mode_mvcc.rst
p7nov: Thanks, moved the note.

patiencedaur (approved):

Issue #1810, opened 2021-02-08
How to describe errors in documentation


Onvember:
When documenting a function, we have the "Possible errors" section. But now we don't have a unified style of describing them. I think, we should agree on this procedure and update our guidelines

Linked issues:
https://github.com/tarantool/doc/issues/1703#issuecomment-747071946
https://github.com/tarantool/doc/issues/1348

PR #3024, opened 2022-07-12
Actualize the API documentation style guide

patiencedaur:
Deployment: https://docs.d.tarantool.io/en/doc/api-style-guide/contributing/docs/api/

### DIFF
--- a/doc/contributing/docs/api.rst
+++ b/doc/contributing/docs/api.rst
@@ -1,12 +1,12 @@
-Examples and templates
-======================
+Documenting the API
+===================
 
 This document contains general guidelines for describing the Tarantool API,
 as well as examples and templates.
 
 Use this checklist for documenting a function or a method:
 
-*   General description
+*   General description (in imperative mood)
 *   :ref:`Parameters <documenting_parameters>`
 *   What this function returns (if nothing, write 'none')
 *   Return type (if exists)
@@ -82,7 +82,7 @@ Configuration parameters
 For every configuration parameter, list the following details:
 
 *   Since which Tarantool version
-*   General description
+*   General description (in imperative mood)
 *   Type
 *   Corresponding environment variable (if applicable)
 *   Default value
--- a/doc/contributing/docs/api.rst
+++ b/doc/contributing/docs/api.rst
@@ -18,27 +18,23 @@ For more style-related specifics, consult the :doc:`Language and style </contrib
 Indicating the version
 ----------------------
 
-For every new feature or parameter, please indicate the version it was introduced in.
+For every new module, function, or method, specify the version it first appears in.
 
-With Tarantool features and parameters, use one of the following Sphinx directives:
+For a new parameter, specify the version it first appears in if this parameter is a "feature"
+and the version it's been introduced in differs from
+the version introducing the function/method and all other parameters.
+
+To specify the version, use the following Sphinx directive:
 
 ..  code-block:: rst
-
-    Since :tarantool-release:`2.10.0`.
-    This is a link to the release notes on GitHub.
 
     Since :doc:`2.10.0 </release/2.10.0>`.
     This is a link to the release notes on the Tarantool documentation website.
 
 The result looks like this:
 
-    Since Tarantool :tarantool-release:`2.10.0`.
-    This is a link to the release notes on GitHub.
-
     Since Tarantool :doc:`2.10.0 </release/2.10.0>`.
     This is a link to the release notes on the Tarantool documentation website.
-
-See also the :doc:`guide on writing release notes </contributing/release_notes/>`.
 
 ..  _contributing-api-docs_general-description:
 
@@ -53,6 +49,8 @@ Use one of the two options:
 Checklist
 ---------
 
+Each list item is a characteristic to be described. Some items can be optional.
+
 Function or method
 ^^^^^^^^^^^^^^^^^^
 
@@ -61,8 +59,10 @@ Function or method
 *   :ref:`Parameters <documenting_parameters>`
 *   What this function returns (if nothing, write 'none')
 *   Return type (if exists)
-*   Possible errors (if exist)
-*   Complexity factors (if exist)
+*   :ref:`Possible errors <contributing-docs-possible_errors>` (if exist)
+*   :ref:`Complexity factors <index-box_complexity-factors>`
+    (for :doc:`CRUD operations </reference/reference_lua/box_space>` and
+    :doc:`index access functions </reference/reference_lua/box_index/>`)
 *   Usage with memtx and vinyl (if differs)
 *   Example(s)
 *   Extra information (if needed)
@@ -82,8 +82,8 @@ See :ref:`class data example <contributing-api-docs_class-example>`.
 
 ..  _documenting_parameters:
 
-Function and class parameters
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Function and method parameters
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 *   :ref:`Since which Tarantool version <contributing-docs-api_version>`
     (if added later)
@@ -93,9 +93,18 @@ Function and class parameters
 
 If the parameter is optional, make sure it is enclosed in square brackets
 in the function declaration (in the "heading").
+Do not mark parameters additionaly as "optional" or "required":
 
-In the "Possible errors" section of the function or class method,
-consider explaining what happens if the parameter hasn't been defined or has the wrong value.
+..  code-block:: rst
+
+    ..  function:: format(URI-components-table[, include-password])
+
+        Construct a URI from components.
+
+        :param URI-components-table: a series of ``name:value`` pairs, one for each component
+        :param include-password: boolean. If this is supplied and is ``true``, then
+                                 the password component is rendered in clear text,
+                                 otherwise it is omitted.
 
 ..  _documenting_confvals:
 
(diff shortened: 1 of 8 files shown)


Totktonada (approved):
LGTM. I would also add `CONTRIBUTING.md` file, because GitHub highlight it for first-time contributors. It is good entry point for new contributors.

doc/contributing/docs/api.rst
```diff
@@ -82,7 +88,7 @@ Configuration parameters
 For every configuration parameter, list the following details:
 
 *   Since which Tarantool version
-*   General description
+*   General description (in imperative mood)
```
p7nov: Minor: "imperative mood" seems unclear for parameter descriptions – they are mostly noun phrases.

p7nov (approved):

doc/contributing/docs/api.rst
veod32: It depends.
In many cases, we describe parameters starting with the verb and using the imperative mood. For example,

- https://www.tarantool.io/en/doc/latest/reference/configuration/#confval-background
- https://www.tarantool.io/en/doc/latest/reference/configuration/#confval-background

But in some cases the description starts with a noun, like
https://www.tarantool.io/en/doc/latest/reference/configuration/#confval-memtx_dir
https://www.tarantool.io/en/doc/latest/reference/configuration/#confval-memtx_dir

Perhaps, we need to recommend both options to writes and give examples of both.

doc/contributing/docs/api.rst
```diff
@@ -82,7 +88,7 @@ Configuration parameters
 For every configuration parameter, list the following details:
 
 *   Since which Tarantool version
```
veod32: We need to change the guideline and example by explaining how to write the version info -- use Sphinx role or reference to the corresponding release note (RN) document, for example,
```
Since :doc:`2.10.0 </release/2.10.0>`.
```

patiencedaur:
We might also consider explaining how the `..  function::` and `..  module::` directives work, specifically that they don't produce headings displayed in the right menu. This means that each new function section should start with an `.. _anchor:`, which should:
- either be linked to in the "List of all functions"/"Index" section on the module page
- or put in the hidden `.. toctree::` on the module page.

Let's do it in another PR.

doc/contributing/docs/api.rst
```diff
@@ -0,0 +1,170 @@
...
+Consider using `LanguageTool <https://languagetool.org/>`_ or `Grammarly <https://www.grammarly.com/>`_
+to check your English.
+For more style-related specifics, consult the :doc:`Language and style </contributing/docs/style>` section.
+
+..  _contributing-docs-api_version:
+
+Indicating the version
+----------------------
+
+For every new feature or parameter, please indicate the version it was introduced in.
```
veod32: (here and below in the text)

"Feature" sounds a bit unclear: actually, it can be anything -- a new module, or function, or method, or even a new parameter. Maybe we can mention them more specific?

Besides, I think we should be more specific when we need to mention the version and when we don't. 

For example, a new module is implemented in the 2.10.0 release. In the module description, we definetely should mention "Since 2.10.0". Should we do the same for each of the module functions and methods? Seemingly yes, because a user can navigate directly to a function/method page without looking at the module description. 
Should we mention "Since 2.10.0" for each parameter of the new module? I think no, because it's excessive -- this info is already here in the function/module description. We need to specify a version for a parameter if this parameter is "a feature" and the version it's introduced in differs from the version of the function/method and all other parameters.

doc/contributing/docs/api.rst
```diff
@@ -0,0 +1,170 @@
...
+
+With Tarantool features and parameters, use one of the following Sphinx directives:
+
+..  code-block:: rst
+
+    Since :tarantool-release:`2.10.0`.
+    This is a link to the release notes on GitHub.
+
+    Since :doc:`2.10.0 </release/2.10.0>`.
```
veod32, line 31: I think we need to use only this one in API description. Otherwise, we can get a kind of mess: someone will put a GitHub link and someone -- a link to documentation RNs. Better to unify this:
- in API section of documentation, we give links to RNs in our documentation
- in RNs, we give links to the original GutHub release notes (:tarantool-release:)

doc/contributing/docs/api.rst
```diff
@@ -0,0 +1,170 @@
...
+
+Language of the general description
+-----------------------------------
+
+Use one of the two options:
+
+*   Start with a verb in the imperative mood. Example: *Create a fiber.*
+*   Start with a noun. Example: *The directory where memtx stores snapshot files.*
+
+Checklist
```
veod32, line 49: I would add a short intro: that each list item is a point that should be described (taking into account that some items can be optional, of course).

doc/contributing/docs/api.rst
```diff
@@ -0,0 +1,170 @@
...
+
+Function or method
+^^^^^^^^^^^^^^^^^^
+
+*   :ref:`Since which Tarantool version <contributing-docs-api_version>`
+*   :ref:`General description <contributing-api-docs_general-description>`
+*   :ref:`Parameters <documenting_parameters>`
+*   What this function returns (if nothing, write 'none')
+*   Return type (if exists)
+*   Possible errors (if exist)
```
veod32: I would make a separate section for "Possible errors" in our guidelines similar to the "Function and class parameters" section.

doc/contributing/docs/api.rst
```diff
@@ -0,0 +1,170 @@
...
+*   Complexity factors (if exist)
```
veod32: We can make a separate section in the guidelines or mention directly here that complexity factors are relevant for a few functions only, and put a link here to https://www.tarantool.io/en/doc/latest/book/box/data_model/#complexity-factors

doc/contributing/docs/api.rst
```diff
@@ -0,0 +1,170 @@
...
+*   :ref:`Since which Tarantool version <contributing-docs-api_version>`
+*   :ref:`General description <contributing-api-docs_general-description>`
+*   Return type
+*   Example
+
+See :ref:`class data example <contributing-api-docs_class-example>`.
+
+..  _documenting_parameters:
+
+Function and class parameters
```
veod32: Maybe "Function and method parameters"? I don't remember if a class itself can have parameters. [Not sure about that]

doc/contributing/docs/api.rst
```diff
@@ -0,0 +1,170 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+*   :ref:`Since which Tarantool version <contributing-docs-api_version>`
+    (if added later)
+*   :ref:`General description <contributing-api-docs_general-description>`
+*   Type
+*   Default value (if optional), possible values
+
+If the parameter is optional, make sure it is enclosed in square brackets
+in the function declaration (in the "heading").
```
veod32, line 95: It's good that this guideline is in general description for all the parameters now!

Maybe we can give an example here? Because it's not clear what "heading" means in this case.

Also, we can think of being more direct: to tell the guys not to write "Optional" and "Required" in the paramter description because you can see that in the function declaration.

doc/contributing/docs/api.rst
```diff
@@ -0,0 +1,170 @@
...
+
+In the "Possible errors" section of the function or class method,
+consider explaining what happens if the parameter hasn't been defined or has the wrong value.
```
veod32: Can we put this in a separate section in the guidelines? Because possible errors can be related to the entire function/method, not necesserily to the misuse of some parameters. 
See, for example, 
- https://www.tarantool.io/en/doc/latest/book/connectors/cxx/tntcxx_api/#public-methods
- https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_space/pairs/

doc/contributing/docs/api.rst
```diff
@@ -0,0 +1,170 @@
...
+Configuration parameters are not to be confused with class and method parameters.
+Configuration parameters are passed to Tarantool via the command line or in an initialization file.
+You can find a list of Tarantool configuration parameters in the :doc:`configuration reference </reference/configuration/index>`.
+
+*   :ref:`Since which Tarantool version <contributing-docs-api_version>`
+*   :ref:`General description <contributing-api-docs_general-description>`
+*   Type
+*   Corresponding environment variable (if applicable)
+*   Default value
+*   Possible values
```
veod32: In the "Examples and templates > Configuration parameters" below, let's make it more clear where to describe possible values: in the general description or as a separate item like 
Type
Env variable
Default value
Dynamic

doc/contributing/docs/api.rst
```diff
@@ -0,0 +1,170 @@
...
+
+..  _contributing-api-docs_function-example:
+
+Module functions
+^^^^^^^^^^^^^^^^
+
+We use the Sphinx directives ``.. module::``
+and ``.. function::`` to describe functions of Tarantool modules:
+
+..  literalinclude:: ./_includes/function_template.rst
```
veod32, line 147: In this example we have
```
    :param function-arguments: what will be passed to function.
                               If the arguments are optional, put them in
                               square brackets in the function declaration.
```

I suggest to remove "If the arguments are optional, put them in square brackets in the function declaration."  
- We have this guideline already in the "Function and class parameters" section.
- It's now a mixture of the actual functional description of the parameter and the meta info (writing guidelines). And it's confusing.

doc/contributing/docs/api.rst
```diff
@@ -0,0 +1,170 @@
...
+..  include:: ./_includes/class_template.rst
+
+..  _contributing-api-docs_confval-example:
+
+Configuration parameters
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+Example:
+
+..  literalinclude:: ./_includes/confval_template.rst
```
veod32, line 182: Maybe we can give here another example? Of a parameter with a number of possible values -- to illustrate this checklist item.

doc/contributing/docs/api.rst
patiencedaur: There's a separate issue for this: https://github.com/tarantool/doc/issues/1810

doc/contributing/docs/api.rst
patiencedaur: Oops, that was definitely a typo.

doc/contributing/docs/api.rst
veod32: ОК

doc/contributing/docs/api.rst
patiencedaur: https://www.tarantool.io/en/doc/latest/reference/configuration/#confval-election_mode
This is the only one with "possible values" provided in a list. It has a huge description, so I wouldn't include it in this specific article.
I think that the possible values may, but need not, be a list.

veod32 (approved):

PR #3025, opened 2022-07-12
Add description of vshard.enable/disable()

### LINKED ISSUE #3012: vshard.router.enable/disable()
**Product:** vshard
**Since:** product version
**Root document:** https://www.tarantool.io/en/doc/latest/reference/reference_rock/vshard/vshard_router/
(source: https://github.com/tarantool/doc/blob/latest/doc/reference/reference_rock/vshard/vshard_router.rst)
**SME:** @ Serpentian

# Details

`vshard.router.disable()` makes most of the `vshard.router`
functions throw an error. As Lua exception, not via `nil, err`
pattern.

`vshard.router.enable()` reverts the disable.

`router_object:enable()/disable()`, where `router_object` is
the return value of `vshard.router.new()`, can also be used
for manual API access configuration for the specific non-static
router.

By default the router is enabled.

Additionally, the router is forcefully disabled automatically
until its configuration is finished and the instance finished
recovery (its `box.info.status` is `'running'`, for example).

Auto-disable protects from usage of vshard functions before the
router's global state is fully created.

Manual router's disabling helps to achieve the same for user's
application. For instance, a user might want to do some preparatory
work after `vshard.router.cfg` before the application is ready.
Then the flow would be:
```Lua
vshard.router.disable()
vshard.router.cfg(...)
-- Do your preparatory work here ...
vshard.router.enable()
```

The behavior of the router's API enabling/disabling is similar to the
storage's one.
Requested by @Serpentian in https://github.com/tarantool/vshard/commit/dd70cfb2c5ec36ab7d5355b0024e5f6d21bb8f9f.

Serpentian:
* Add description of the automatic and manual router API's
  enabling/disabling

Resolves #3012

### DIFF
--- a/doc/reference/reference_rock/vshard/vshard_router.rst
+++ b/doc/reference/reference_rock/vshard/vshard_router.rst
@@ -31,6 +31,8 @@ Router API
     |                                             | * :ref:`vshard.router.discovery_set() <router_api-discovery_set>`                                         |
     |                                             | * :ref:`vshard.router.info() <router_api-info>`                                                           |
     |                                             | * :ref:`vshard.router.buckets_info() <router_api-buckets_info>`                                           |
+    |                                             | * :ref:`vshard.router.enable() <router_api-enable>`                                                       |
+    |                                             | * :ref:`vshard.router.disable() <router_api-disable>`                                                     |
     |                                             | * :ref:`replicaset_object:call() <router_api-replicaset_call>`                                            |
     |                                             | * :ref:`replicaset_object:callro() <router_api-replicaset_callro>`                                        |
     |                                             | * :ref:`replicaset_object:callrw() <router_api-replicaset_callrw>`                                        |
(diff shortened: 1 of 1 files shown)


doc/reference/reference_rock/vshard/vshard_router.rst
```diff
@@ -768,6 +770,46 @@ Router public API
...
+
+..  function:: vshard.router.enable()
+
+    Manually allow access to the router API, revert
+    :ref:`vshard.router.disable() <router_api-disable>`.
+
+        ..  note::
+
+            ``vshard.router.enable()`` cannot be used for enabling a router
+            API that was automatically disabled due to a running configuration
```
Serpentian, line 783: Actually, `vshard.router.cfg` is a configuration process only for the static router. But vshard supports multiple routers in a single instance. These ones can be created via `vhard.router.new`. During configuration every router is blocked on its own,  and blocking API of the one uncofigured instance doesn't affect others. 

If we want to specify functions, used for configuration, explicitly, we should mention both of them. Another solution can be just stay on 'running configuration process'.

patiencedaur (approved):
Thanks!

PR #3031, opened 2022-07-14
Add tt commands help, version, completion

### LINKED ISSUE #2997: tt version, tt completion, tt help
**Epic: [Document TT CLI](https://github.com/tarantool/doc/issues/2553)**

**Related dev. document:** https://github.com/tarantool/tt#commands

**Product:** TT CLI
**Since:** 2.10
**Root document:** TBA
**SME:** @ LeonidVas

# Details

- [ ] `tt version`
- [ ] `tt completion`
- [ ] `tt help`

p7nov:
Resolves #2997


doc/reference/tt_cli/completion.rst
```diff
@@ -0,0 +1,25 @@
...
+
+``tt completion`` generates tab-based completion for ``tt`` commands
+in the specified shell.
+
+Details
+-------
+
+``tt`` completion can be generated for two shells: ``bash`` and ``zsh``.
+
+The generated completion includes both internal and external commands.
```
LeonidVas: Now it only works for internal modules.

LeonidVas (approved):
LGTM.

doc/reference/tt_cli/completion.rst
```diff
@@ -0,0 +1,19 @@
...
+
+
+Examples
+--------
+
+Generate ``tt`` completion for ``bash``:
+
+..  code-block:: bash
+
+    . <(tt completion bash)
```
patiencedaur: It's not immediately clear what "generating completion" means. Completion is a process, right?

This example might need just a little explanation about the Bash syntax used. Judging by the period at the beginning, the result of the command is immediately sourced. If so, does completion keep working if you open another console window?

patiencedaur (commented):
Minor comment.

doc/reference/tt_cli/completion.rst
p7nov: I'd say that completion is a function here (== "ability to suggest/complete commands upon pressing tab") and "generating bash completion" seems a commonly know term if we [ask Google](https://www.google.com/search?q=generate+bash+completion) - about six million results.


>  does completion keep working if you open another console window?

@LeonidVas could you help with the answer? And what do you think about the syntax explanation here? Do we need it?

LeonidVas: The command generates completion script that will work in the terminal where it was executed. For example, you can add the script execution to the `/home/vasya/.bashrc` file, and then the commands will work on all terminals started by the `vasya` user.
A simple example of how copletions work in bash and what copletions scripts are for a program; https://iridakos.com/programming/2018/03/01/bash-programmable-completion-tutorial

The dot in the example is an alias of the `source` command, and in this example copletions will only work in the terminal where the command was executed.

As for me, ''generatig completion'' sounds fine and It is not necessary to add explanations.

p7nov: Thanks, added a point about the current terminal.

patiencedaur: Maybe we should tell the user to add the generated script to `.bashrc` instead? So that it works in all terminals:
`tt completion bash >> ~/.bashrc && . ~/.bashrc` or something like that

LeonidVas: - As for me, I prefer to create a script in `home` and add in `.bashrc` just a call to this script.
- IMHO, we should allow the user to generate this script and how the user will use it is up to him. But I think we can give a recommendation (not a command) that he should add it to `.bashrc`, so that the script is automatically executed ... Because if we give users a command, they will start executing it without thinking and littering their system, `.bashrc`, something else.

patiencedaur: Good point!

patiencedaur (approved):

PR #3036, opened 2022-07-15
Describe datetime data and index types

### LINKED ISSUE #2509: Storage support for datetime values
It's now possible to store datetime values to spaces and to create
indexed datetime fields.
Requested by @tsafin in https://github.com/tarantool/tarantool/commit/3bd870261c462416c29226414fe0a2d79aba0c74.

Docbot didn't create request because that commit tried to invent its own docbot syntax by adding an empty line before `Title:` line. It does not work. So here I create it manually. For details about the feature itself contact T. Safin please.

patiencedaur:
Resolves #2509

### DIFF
--- a/doc/book/box/data_model.rst
+++ b/doc/book/box/data_model.rst
@@ -163,13 +163,17 @@ Lua versus MsgPack
             -   `cdata`_
             -   ``[!!binary 3t7e]``
         *   -   scalar
+            -   ext (for Tarantool ``decimal``)
+            -   `cdata`_
+            -   ``1.2``
+        *   -   scalar
             -   ext (for Tarantool ``datetime``)
             -   `cdata`_
             -   ``'2021-08-20T16:21:25.122999906 Europe/Berlin'``
         *   -   scalar
-            -   ext (for Tarantool ``decimal``)
+            -   ext (for Tarantool ``interval``)
             -   `cdata`_
-            -   ``1.2``
+            -   ``+1 months, 1 days``
         *   -   scalar
             -   ext (for Tarantool ``uuid``)
             -   `cdata`_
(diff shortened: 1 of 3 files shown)


doc/book/box/data_model.rst
```diff
@@ -119,67 +119,93 @@ Therefore a developer often deals with two type sets:
...
-    +===================+=========================+================================+==============================+
-    | scalar            | nil                     | "`nil`_"                       | ``nil``                      |
-    +-------------------+-------------------------+--------------------------------+------------------------------+
-    | scalar            | boolean                 | "`boolean`_"                   | ``true``                     |
-    +-------------------+-------------------------+--------------------------------+------------------------------+
-    | scalar            | string                  | "`string`_"                    | ``'A B C'``                  |
-    +-------------------+-------------------------+--------------------------------+------------------------------+
-    | scalar            | integer                 | "`number`_"                    | ``12345``                    |
-    +-------------------+-------------------------+--------------------------------+------------------------------+
-    | scalar            | float 64 (double)       | "`number`_"                    | ``1.2345``                   |
```
tsafin, line 145: I do see it's kind of offtopic for the `datetime` discussion, but still is bothering me: why there are 2 rows for double literals? (LuaJIT ffi cdata and standard Lua number). What we wanted to highlight here? (That we could, depending on value, may end up using either number or extended LuaJIT type? But similarly we would have for integer values, but there are no duplication.)

I'me very much confused here...

doc/book/box/data_model.rst
```diff
@@ -261,15 +287,53 @@ Here are examples of numbers using regular notation, exponential notation,
...
 Example: a value returned by a function in the :ref:`decimal <decimal>` module.
 
-.. _index-box_string:
+..  _index-box_datetime:
+
+**datetime**. Introduced in :tarantool-release:`2.10.0`.
+The Tarantool ``datetime`` type facilitates operations with date and time,
+accounting for leap years or the varying number of days in a month.
+It is stored as a MsgPack ext (Extension).
+This data type is based on the `c-dt <https://github.com/tarantool/c-dt>`_ module.
```
tsafin: Should be better sound like "Operations on this data type use code from  3rd party `c-dt <https://github.com/tarantool/c-dt>`_ library."

doc/book/box/data_model.rst
```diff
@@ -261,15 +287,53 @@ Here are examples of numbers using regular notation, exponential notation,
...
+
+To create a ``datetime`` object, pass either a table of time units (from nanoseconds to years) or a timestamp.
```
tsafin: Why we do have some basic datetime operations described here? While there are no such neither for `decimal` or for any other extended or standard type? 

It smells like storage data model doc is improper place for describing `datetime` API operations. Could we simply insert link here to datetime module documentation and be with it?

tsafin (changes requested):
I have no objections about markup refactorings here and there, but datetime doc in the messagepack datamodel descriptions seems like is mismatching their current neighbourhood. And should be moved elsewhere.

tsafin (approved):

doc/book/box/data_model.rst
patiencedaur: Thanks for noticing that! Tracked in https://github.com/tarantool/doc/issues/3040.

Issue #2445, opened 2021-11-16
[3pt] box.info.ro_reason


TarantoolBot:
**Product:** Tarantool
**Since:** 2.10.0-beta2
**Audience/target:** Tarantool users
**Root document:** https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_info/
**SME:** @ Gerold103

Related to https://github.com/tarantool/doc/issues/2444
Has a follow-up in https://github.com/tarantool/doc/issues/2953

## Details

The new `box.info` field - `ro_reason` - is `nil` on a writable
instance. On a read-only instance `box.info.ro == true` it reports
an error reason. Currently the list is
- `'election'` - `box.cfg.election_mode` is not `'off'` and this
  instance is not a leader. See `box.info.election` for details.
- `'synchro'` - the synchro queue is owned by some other instance.
  For details see `box.info.synchro`.
- `'config'` - `box.cfg.read_only` is true;
- `'orphan'` - the instance is in orphan state.
Requested by @Gerold103 in https://github.com/tarantool/tarantool/commit/48185e8641fe83be16aa2b93d250ab0482f8e5cb

## ToDo

- [x] - Mention the new field in the list on the main page https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_info/

- [x] Create a child page for the `box.info.ro_reason` and put the description of the reasons there

- [ ] Make cross-reference to the description of the `error_object.reason` (see #2444)

- [ ] Don't forget about the follow-up of the topic in #2953

PR #3080, opened 2022-08-10
Update the upgrade instructions

### LINKED ISSUE #836: [0pt] Clarify tarantool update process to a newer major version
Our guarantee is that Nth major version can be upgraded to (N+1)th one, but we does not guarantee the same for (N+2)th version.

Say, to update 1.7 to 2.1 we need to update it first to 1.9, then to 1.10 and only then to 2.1.

We should clarify also what is a major version of tarantool: 1.7, 1.9, 1.10, 2. The way to enumarate versions was changed since 2.*.

### LINKED ISSUE #1807: [3pt] Clarify documentation regarding upgrade from 1.6 to 2.x
Our doc states that there are no problems in upgrading 1.6 to 2.x. See [here](https://www.tarantool.io/en/doc/latest/book/admin/upgrades/#how-to-upgrade-from-tarantool-1-6-to-2-x)

However, the [paragraph above](https://www.tarantool.io/en/doc/latest/book/admin/upgrades/#upgrading-a-tarantool-instance) states that there are problems with upgrading from 1.6 to 2.x directly, and this is true.

Actually, there is no easy way to upgrade tarantool 1.6 to 2.x directly.
The recommended option is to perform a 2 step upgrade: first perform 1.6 -> 1.10 upgrade, and then 1.10 -> 2.x
Then there shouldn't be any problems.

Another option, when a direct upgrade is really needed, is to use an upgrade script. The script should be run on a 2.x instance before trying to boot the 1.6 data files (.xlog and .snap). The latest script version is in [this gist](https://gist.github.com/sergepetrenko/1e27197ca2d5995adda1a959f4627295). It'll be updated and fixed later, so do not rely on it right now.

### LINKED ISSUE #3042: Upgrade plan from 1.6 to 1.10
**Epic: [Tarantool installation and upgrades](https://github.com/tarantool/doc/issues/2572)**

**SME**: @ sergepetrenko

We have a cluster of 1.6 with the following configuration:
- 7 nodes
- full mesh
- one rw, others ro
- rich application with background fibers and network operations (net.box & http)
- rw have background fibers modify data and work with network
- balancer, that routes request traffic to ro or rw depending on instance status and role (in fact, balancers are 1.10)

We know, that application works on 1.10 without any modifications due to compatibility layer, based on `kit`, `config` and `spacer/v1` (from github.com/moonlibs).

We need a guide to perform **live** migration without downtime from 1.6 to 1.10, i.e. we can't stop writes or reads "for some time"

patiencedaur:
Resolves #3042 
Resolves #1807 
Resolves https://github.com/tarantool/doc/issues/836

Deploy: https://docs.d.tarantool.io/en/doc/gh-3042-caveats/book/admin/upgrades/

- Clarify the process of update in a replica set
- Move the version-specific instructions from tarantool/wiki:
  https://github.com/tarantool/tarantool/wiki/Caveats-when-upgrading-from-tarantool-1.6
  https://github.com/tarantool/tarantool/wiki/Fix-wrong-order-of-decimals-and-doubles-in-indices


doc/book/admin/_includes/upgrade_procedure.rst
```diff
@@ -0,0 +1,27 @@
...
+5. Make one of the updated replicas the new master.
+   Check that it continues following and being followed by all other replicas.
+
+6. :ref:`Upgrade <admin-upgrades_instance>` the former master.
+
+7. :ref:`Upgrade the database <admin-upgrades_db>` on the new master by running ``box.schema.upgrade()``.
+   Changes are propagated to other
+   nodes via the regular replication mechanism.
+
+8. Run ``box.snapshot()`` on every node in the replica set to take a snapshot of all the data.
```
sergepetrenko: I'd rather say
`to make sure replicas immediately see the upgraded database state in case of restart`

doc/book/admin/upgrades.rst
```diff
@@ -58,95 +50,78 @@ How to upgrade from Tarantool 1.7 to 2.x
 3. Update the Tarantool server. See installation instructions at Tarantool
    `download page <http://tarantool.org/download.html>`_.
 
-4. Launch the updated Tarantool server using ``tarantoolctl`` or ``systemctl``.
+4. Update the Tarantool database. Put the request ``box.schema.upgrade()``
+   inside a :doc:`box.once() </reference/reference_lua/box_once>` function in your Tarantool
```
sergepetrenko: You reference this section from "updating the cluster", I think the users might get confused and think that they should run `box.schema.upgrade()` on every node that they upgrade.
Let's mention that `box.schema.upgrade()` should be issued here only when the instance is not part of the cluster.

sergepetrenko: Maybe it would be better not to mention `box.schema.upgrade()` neither in "upgrading an instance" nor in "upgrading the cluster", and add a separate short paragraph "finishing the upgrade", which would describe `box.schema.upgrade()` and why it's needed.

doc/book/admin/upgrades.rst
```diff
@@ -58,95 +50,78 @@ How to upgrade from Tarantool 1.7 to 2.x
...
-`release notes for Tarantool 1.7 <https://github.com/tarantool/tarantool/releases>`_.
+..  important::
 
-1. Check with application developers whether application files need to be
-   updated due to incompatible changes (see
-   `1.7 release notes <https://github.com/tarantool/tarantool/releases>`_).
-   If yes, back up the old application files.
+    The only way to upgrade Tarantool from 1.6 to 2.x **without downtime** is
+    taking an intermediate step by upgrading from 1.6 to 1.10 and then to 2.x.
+    This also applies to upgrading Tarantool from 1.7 and 1.9.
```
sergepetrenko: Maybe say that "the only way to upgrade Tarantool from a version prior to 1.10 (like 1.6, 1.7 and 1.9) to 2.x without downtime is to take an intermediate step..." ?

doc/book/admin/upgrades.rst
```diff
@@ -58,95 +50,78 @@ How to upgrade from Tarantool 1.7 to 2.x
...
 
-2. Stop the Tarantool server.
+    Before upgrading Tarantool from 1.x to 2.y, please read about the associated
```
sergepetrenko: But this is only about upgrading 1.6 to 1.10, isn't it?

doc/book/admin/upgrades.rst
```diff
@@ -58,95 +50,78 @@ How to upgrade from Tarantool 1.7 to 2.x
...
-   num -> unsigned, str -> string) and options in Tarantool system spaces.
+#. Check the replica set health by running the following code on every replica:
 
-6. Update application files, if needed.
+   ..  code-block:: tarantoolsession
 
-7. Launch the updated Tarantool server using ``tarantoolctl`` or ``systemctl``.
+       box.info.ro -- shouldn't be "orphan"
+       box.info.ro_reason -- since 2.10, shouldn't be "orphan"
+       box.info.status -- should be "running"
```
sergepetrenko: Actually we should only check that `box.info.status` is "running" on all the instances and that at least one instance has `box.info.ro` == `false`.

In case all the instances have `box.info.ro == true`, one may check `box.info.ro_reason`, starting from version 2.10, to have a clue what the issue is

doc/book/admin/upgrades.rst
```diff
@@ -58,95 +50,78 @@ How to upgrade from Tarantool 1.7 to 2.x
...
 
-.. _admin-upgrades_replication_cluster:
+   If ``box.info.ro``, ``box.info.ro_reason``, or ``box.info.status`` have the value ``orphan``,
```
sergepetrenko: `box.info.status` and `box.info.ro_reason`.
`box.info.ro` only holds true or false

doc/book/admin/upgrades.rst
```diff
@@ -58,95 +50,78 @@ How to upgrade from Tarantool 1.7 to 2.x
...
-Upgrading Tarantool in a replication cluster
---------------------------------------------------------------------------------
+   If you're running Cartridge, you can also check node health in the UI.
 
-Tarantool 1.7 can work as a :ref:`replica <replication-architecture>`
-for Tarantool 1.6 and vice versa. Replicas
-perform capability negotiation on handshake, and new 1.7 replication features
-are not used with 1.6 replicas. This allows upgrading clustered configurations.
+#. Make sure each replica connected to the rest of the replica set.
+   Run the following, replacing the ``id`` with the replica ID (the output of ``box.info.id``):
```
sergepetrenko: I think it would be simpler to run
`box.info.replication`, and then check the output table (which is indexed by replica ids and contains upstream, downstream for each id). Each upstream and downstream should have status = follow. The only exceptions are upstream and downstream belonging to this particular instance (they are indexed by `box.info.id`)

doc/book/admin/upgrades.rst
```diff
@@ -58,95 +50,78 @@ How to upgrade from Tarantool 1.7 to 2.x
...
-   :ref:`Upgrading a Tarantool instance <admin-upgrades_instance>`.
+       box.info.replication[id].upstream
+       box.info.replication[id].downstream
 
-2. Verify installation on the replicas:
+   The ``status`` field in both outputs should have the value ``follow``.
+   This means that the replicas are connected and there are no errors in the data flow.
 
-   a. Start Tarantool.
+   The value of the ``lag`` field should be less than 1 second.
```
sergepetrenko: I think this needs some clarification.
The value of `lag` should be sane, for example, when `box.cfg.replication_timeout` is 1 second, it's generally ok to have lag ~ 10 seconds. I guess only the user knows which values of lag are fine for him.

It's always ok to have lag <= box.cfg.replication_timeout. 
When lag is > box.cfg.replication_timeout the user has to use common sense. A lag of several hours is surely not okay (>= thousands of seconds)

Sorry if I told you otherwise in chat.

doc/book/admin/upgrades/1.6-1.10.rst
```diff
@@ -0,0 +1,54 @@
+..  _admin-upgrades-1.6-1.10:
+
+Caveats when upgrading from Tarantool 1.6 to 1.10
+=================================================
+
+This page includes explanations and solutions to some common issues
+when upgrading the cluster from Tarantool 1.6 to 1.10.
+This will be of interest to anyone upgrading to the most recent Tarantool version,
+like 2.8.4, 2.10.0 and such, since taking an intermediate step with
```
sergepetrenko: Let's replace 2.10.0 with 2.10.1, since it's more up to date as of the time of writing

doc/book/admin/upgrades/1.6-1.10.rst
```diff
@@ -0,0 +1,54 @@
...
+upgrading 1.6 -> 1.10 -> 2.x is the only way to perform the upgrade without downtime.
+
+A direct upgrade from 1.6 to 2.x is also possible, but only with downtime.
```
sergepetrenko: I can't think of a place to put it in, but I think we should mention how is a direct upgrade from 1.6 (or 1.7 or 1.9) to 2.x with downtime possible:

1. Stop all instances in the cluster
2. Upgrade Tarantool version to 2.x on every instance
3. Upgrade the corresponding instance files and applications, if needed
4. Start all the instances with tarantool 2.x
5. Execute `box.schema.upgrade()` on master
6. Execute `box.snapshot()` on every node in the cluster.

sergepetrenko (changes requested):
Thanks for working on this!
It's good our upgrade page is getting up to date and more clear.
Please, find my comments below

doc/book/admin/_includes/upgrade_procedure.rst
patiencedaur: You mean, this is the reason we run `box.snapshot()` in this context?

sergepetrenko: Yep

patiencedaur:
Ro_reason: https://github.com/tarantool/doc/issues/2445

doc/book/admin/upgrades.rst
```diff
@@ -47,31 +47,32 @@ instructions for individual versions :ref:`in the list below <admin-upgrades_ver
...
-   The ``status`` field in both outputs should have the value ``follow``.
+   Run ``box.info.replication`` and check the output table.
+   For each instance ``id``, there are ``upstream`` and ``downstream`` values.
+   Both of them should have the value ``follow``, except on the instance where you run this code.
    This means that the replicas are connected and there are no errors in the data flow.
 
-   The value of the ``lag`` field should be less than 1 second.
+   The value of the ``lag`` field can be less or equal than ``box.cfg.replication_timeout``,
+   but it can also be moderately larger.
+   For example, if ``box.cfg.replication_timeout`` is 1 second, it's generally OK to have a lag of about 10 seconds.
```
sergepetrenko: > it's generally OK to have a lag of about 10 seconds

Let's append "... under high write load on master" there.

doc/book/admin/upgrades.rst
```diff
@@ -58,95 +25,113 @@ How to upgrade from Tarantool 1.7 to 2.x
...
-from 1.6.x to 1.7.x. Notice that this will **always imply a downtime**.
-To upgrade **without downtime**, you need several Tarantool servers running in a
-replication cluster (see :ref:`below <admin-upgrades_replication_cluster>`).
+    Before upgrading Tarantool from 1.6 to 2.x, please read about the associated
+    :ref:`caveats <admin-upgrades-1.6-1.10>`.
 
-Tarantool 1.7 has an incompatible :ref:`.snap <internals-snapshot>` and
-:ref:`.xlog <internals-wal>` file format: 1.6 files are
-supported during upgrade, but you won’t be able to return to 1.6 after running
-under 1.7 for a while. It also renames a few configuration parameters, but old
```
sergepetrenko, line 86: Please add this part to "caveats when upgrading 1.6 to 2.x"

Saying that tarantool 1.10 (and 2.x) have incompatible snapshot format compared to 1.6
So the user won't be able to return to 1.6 after running 1.10 or 2.x for a while

doc/book/admin/upgrades/1.6-1.10.rst
```diff
@@ -0,0 +1,51 @@
...
+
+    -- First find all spaces containing unsigned fields with non-unsigned values in them.
+    -- Say, we have one such space denoted problematic_space and the problem is in field problematic_field_no.
+    a = box.space.problematic_space:format()
+    a[problematic_field_no].type = 'number'
+    box.space.problematic_space:format(a)
+
+Once this is performed on the master, it's safe to proceed to step 8, making snapshot on every node.
+
+**Step 8:** The user might be concerned with snapshot size in 1.10 --
```
sergepetrenko, line 62: There are no steps 7 and 8 anymore, so these paragraphs should refer to step 6 (about finishing the upgrade)

sergepetrenko (changes requested):
Thanks for the fixes!
Some more comments from me

doc/book/admin/_includes/1.6-to-2.x-condition.rst
```diff
@@ -1,3 +1,8 @@
+Versions later that 1.6 have incompatible :ref:`.snap <internals-snapshot>` and
+:ref:`.xlog <internals-wal>` file formats: 1.6 files are
+supported during upgrade, but you won’t be able to return to 1.6 after running
+under 1.7 for a while. A few configuration parameters are also renamed.
```
sergepetrenko: "under 1.10 / 2.x"

That's up to you, but I think there's no reason to mention unsupported Tarantool versions

sergepetrenko (approved):
Thanks! Looks good to me.

doc/book/admin/upgrades.rst
```diff
@@ -58,95 +25,114 @@ How to upgrade from Tarantool 1.7 to 2.x
...
-   On startup, this will create new system spaces, update data type names (e.g.
-   num -> unsigned, str -> string) and options in Tarantool system spaces.
+#. Make sure each replica connected to the rest of the replica set.
+   Run ``box.info.replication`` and check the output table.
+   For each instance ``id``, there are ``upstream`` and ``downstream`` values.
+   Both of them should have the value ``follow``, except on the instance where you run this code.
+   This means that the replicas are connected and there are no errors in the data flow.
 
-6. Update application files, if needed.
+   The value of the ``lag`` field can be less or equal than ``box.cfg.replication_timeout``,
```
veod32: I would make this paragpaph to be the 3d item in the numbered list -- to make the check-out preparation points more clear
1. ro/rw instances
2. upstream/downtime
3. replication lag

doc/book/admin/_includes/upgrade_procedure.rst
```diff
@@ -0,0 +1,21 @@
+1. Pick any replica in the replica set.
```
veod32: Should it be an RO instance? If yes, let's make this clear.

doc/book/admin/upgrades.rst
```diff
@@ -58,95 +25,114 @@ How to upgrade from Tarantool 1.7 to 2.x
 3. Update the Tarantool server. See installation instructions at Tarantool
    `download page <http://tarantool.org/download.html>`_.
 
-4. Launch the updated Tarantool server using ``tarantoolctl`` or ``systemctl``.
+After that, make sure to :ref:`finish the upgrade properly <admin-upgrades_db>`.
```
veod32: Предлагаю сделать этот пункт снова явно номером #4 в нумерованном списке и привести здесь полный текст процедуры финализации -- или инклюдом включить, или просто текстом написать.

Мотивация:

В этой маленькой инструкции из 4х шагов, мы 3 раза отсылаем пользователя в другие места выполнять другие инструкции. Это всегда неудобно, и лучше сделать по максимуму, чтобы пользователь читал инструкцию на одной странице. Если пп. 2 и 3 еще более или менее норм отсылать, то п. 4 я бы лучше привел здесь в явном виде.

К тому же мы к этой инструкции отсылаем из следущей секции (Upgrading Tarantool in a replica set with no downtime) для апдейта каждого инстанса. Получается уж больно комплексная "матрешка" из редиректов по инструкциям. Это еще один плюс к тому, чтобы сделать максимально инструкцию в одном разделе тут.

Еще один плюс к этой мотивации -- в п.1 мы стоппим инстанс. По логике, в конце мы его должны стартануть. Но в текущем описании не понятно, где. Может в п.3, а может в п.4. 
Лучше это сделать как-то явным образом, чтобы читатель видел - в начале мы стоппанули инстанс, в конце запустили. Особенно учитывая то, что к процедуре апдейта standalone инстанс мы обращаемся из шагов инструкции по "Upgrading ... with no downtime" и той инструкции проверяем коннект инстанса к остальным репликам -- логично, что для этого инстанс должен быть запущен, и лучше явно написать в процедуре апгрейда для standalone.

doc/book/admin/_includes/upgrade_procedure.rst
```diff
@@ -0,0 +1,21 @@
...
+
+2. Upgrade this replica to the new Tarantool version. See details in
+   :ref:`Upgrading a Tarantool instance <admin-upgrades_instance>`.
```
veod32:
```suggestion
   :ref:`Upgrading Tarantool on a standalone instance <admin-upgrades_instance>`.
```

veod32: Учитывая, что мы отсылаем пользователя сделать процедуру "Upgrading Tarantool on a standalone instance", где инстанс
инстанс стоппится, т.е. у него есть явный даунтайм, может имеет смысл пояснить пользователю, почему мы называем процедуру в репликасете "Upgrading ... with no downtime" -- что репликасет продолжает консистентно писать и отдавать данные на оставшихся работающих репликах, а реплику, которую мы стоппанули и проапгрейдили, мы потом снова подключим к репликасету, и она "догонит" остальных  -- если я правильно понял идею этого апдегрейд with no downtime.

Может пользователю это и так очевидно, а может и нет.

doc/book/admin/_includes/upgrade_procedure.rst
```diff
@@ -0,0 +1,21 @@
...
+
+3. Make sure the replica connected to the rest of the replica set just fine:
```
veod32: Как писал в комменте к "Upgrading Tarantool on a standalone instance", лучше как-то явно написать, что в процедуре апгрейда одного инстанса мы его стоппим, а потом запускаем. 
В этом шаге проверяется коннект инстанса ко всем репликам в сете, инстанс явно должен быть в работающем состоянии.

doc/book/admin/_includes/upgrade_procedure.rst
```diff
@@ -0,0 +1,21 @@
...
+
+   ..  code-block:: tarantoolsession
+
+       box.info.replication[id].upstream
+       box.info.replication[id].downstream
+      
+   The ``status`` field in both outputs should have the value ``follow``.
+
+4. :ref:`Upgrade <admin-upgrades_instance>` all the replicas by repeating steps 1--3
```
veod32: > all the replicas

RO replicas?

doc/book/admin/_includes/upgrade_procedure.rst
```diff
@@ -0,0 +1,21 @@
...
+   until only the master keeps running the old Tarantool version.
+
+5. Make one of the updated replicas the new master.
```
veod32: >  Make one of the updated replicas the new master

How? What the procedure?

If a user needs to set `box.cfg{ read_only = false }` on this instance, let's make it clear. 
If yes:
- Can it be done on the working instance or not? It' better to make this clear as well.
- If we're setting this instance to be the new master (RW), what should we do with the former master and when? My point: if we set `box.cfg{ read_only = false }` on one of the RO instance making it the master (RW), there will be a period of time when there're 2 RW instances in the replica set -- this one (new master) and the current master until we make it RO. 

Sure we have explanation about the `read_only` option in the config reference https://www.tarantool.io/en/doc/latest/reference/configuration/#cfg-basic-read-only
but it's better to explain the main essential ToDo point here, and then to refer to the reference for other details if necessary.

doc/book/admin/_includes/upgrade_procedure.rst
```diff
@@ -0,0 +1,21 @@
...
+   Check that it continues following and being followed by all other replicas.
```
veod32: How? What the procedure?

If it's by checking

box.info.replication[id].upstream
box.info.replication[id].downstream

like in the step #3 above, let's make it clear by referencing to this step.

doc/book/admin/_includes/upgrade_procedure.rst
```diff
@@ -0,0 +1,21 @@
...
+
+6. :ref:`Upgrade <admin-upgrades_instance>` the former master.
```
veod32: See comment to the step #5 above.
Let's make it more clear if we need to make the former master to be RO -- how to do this and when.

doc/book/admin/upgrades.rst
```diff
@@ -58,95 +25,114 @@ How to upgrade from Tarantool 1.7 to 2.x
...
+If the replica set is healthy, proceed to the upgrade.
 
-.. _admin-upgrades_replication_cluster:
+Upgrade procedure
+~~~~~~~~~~~~~~~~~
+
+..  include:: ./_includes/upgrade_procedure.rst
+
+After upgrading the replica set, make sure to run ``box.schema.upgrade()`` on the new master
+as described below in the section ":ref:`Finishing the upgrade <admin-upgrades_db>`".
```
veod32:
```suggestion
as described below in the section ":ref:`Finalizing the upgrade <admin-upgrades_db>`".
```

doc/book/admin/upgrades.rst
```diff
@@ -58,95 +25,114 @@ How to upgrade from Tarantool 1.7 to 2.x
...
-7. Launch the updated Tarantool server using ``tarantoolctl`` or ``systemctl``.
+If the replica set is healthy, proceed to the upgrade.
 
-.. _admin-upgrades_replication_cluster:
+Upgrade procedure
+~~~~~~~~~~~~~~~~~
+
+..  include:: ./_includes/upgrade_procedure.rst
+
+After upgrading the replica set, make sure to run ``box.schema.upgrade()`` on the new master
```
veod32: I would make this paragraph to be the step #7 -- it's the part of the entire procedure.

doc/book/admin/upgrades.rst
```diff
@@ -58,95 +25,114 @@ How to upgrade from Tarantool 1.7 to 2.x
...
+as described below in the section ":ref:`Finishing the upgrade <admin-upgrades_db>`".
+There is no need  to run ``box.schema.upgrade()`` on every node:
+changes are propagated to other nodes via the regular replication mechanism.
 
---------------------------------------------------------------------------------
-Upgrading Tarantool in a replication cluster
---------------------------------------------------------------------------------
+Finally, run ``box.snapshot()`` on every node in the replica set
```
veod32: I would make this paragraph to be the step #8 -- it's the part of the entire procedure.

doc/book/admin/_includes/1.6-to-2.x-condition.rst
```diff
@@ -0,0 +1,12 @@
...
+:ref:`.xlog <internals-wal>` file formats: 1.6 files are
+supported during upgrade, but you won’t be able to return to 1.6 after running
+under 1.10 or 2.x for a while. A few configuration parameters are also renamed.
+
+To perform a **live** upgrade from Tarantool 1.6 to a more recent version,
+like :doc:`2.8.4 </release/2.8.4>`, :doc:`2.10.1 </release/2.10.1>` and such,
+it is necessary to take an intermediate step by upgrading 1.6 -> 1.10 -> 2.x.
+This is the only way to perform the upgrade without downtime.
+
+However, a direct upgrade of a replica set from 1.6 to 2.x is also possible, but only
```
veod32, line 11: > direct upgrade of a replica set from 1.6 to 2.x

Better to give a link to the description of this procedure.

doc/book/admin/upgrades/1.6-1.10.rst
```diff
@@ -0,0 +1,59 @@
+..  _admin-upgrades-1.6-1.10:
+
+Live upgrade from Tarantool 1.6 to 1.10
+=======================================
+
+This page includes explanations and solutions to some common issues
+when upgrading a replica set from Tarantool 1.6 to 1.10.
+
+..  include:: ../_includes/1.6-to-2.x-condition.rst
```
veod32, line 9: It's not really clear why we're talking about the 1.6 > 2.x upgrade conditions in the topic for the 1.6 > 1.10 upgrade.

doc/book/admin/upgrades/1.6-1.10.rst
```diff
@@ -0,0 +1,59 @@
...
+
+Let's first reiterate the upgrade procedure for any Tarantool version:
```
veod32: May be to rephrase this intro sentence somehow? 
The previous sentence says, "However, a direct upgrade of a replica set from 1.6 to 2.x is also possible, but only with downtime."
And then we have this intro and the procedure description, but it's not clear that this is the generic procedure regardless the TT version.

doc/book/admin/upgrades/1.6-1.10.rst
```diff
@@ -0,0 +1,59 @@
...
+    changes are propagated to other nodes via the regular replication mechanism.
+
+8.  Run ``box.snapshot()`` on every node in the replica set
+    to make sure that the replicas immediately see the upgraded database state in case of restart.
+
+What's different when upgrading from Tarantool 1.6:
+
+**Step 2:** Tarantool 1.10+ fails to recover from 1.6 xlogs, unless ``box.cfg{force_recovery = true}`` is set.
+There is some small difference between 1.6 and 1.10 xlogs, which makes 1.6 xlogs appear erroneous to 1.10+ instances.
+In order to work around this, start the instance in ``force_recovery`` mode.
```
veod32: > In order to work around this, start the instance in ``force_recovery`` mode.

When and where a user should set ``box.cfg{force_recovery = true}`` ? In the init.lua starting file before launching the instance after upgrade? Better to make it more clear.

doc/book/admin/upgrades/1.6-1.10.rst
```diff
@@ -0,0 +1,59 @@
...
+
+..  code-block:: lua
+
+    -- First find all spaces containing unsigned fields with non-unsigned values in them.
+    -- Say, we have one such space denoted problematic_space and the problem is in field problematic_field_no.
+    a = box.space.problematic_space:format()
+    a[problematic_field_no].type = 'number'
+    box.space.problematic_space:format(a)
+
+Once this is performed on the master, it's safe to proceed to step 8, making snapshot on every node.
```
veod32: Where a user should run this piece of Lua code
- in TT console 
- or put it in the init.lua starting file before launching the instance
- or as a separate script file that will be run after performing ``box.schema.upgrade()``

Better to make it more clear.

veod32: > to proceed to step 8

Be seeing that, I started to read the paragraph below **Step 8:** The user...
instead of going to the real step in the procedure above.

doc/book/admin/upgrades/1.6-2.0-downtime.rst
```diff
@@ -0,0 +1,20 @@
+..  _admin-upgrades-1.6-2.x_downtime:
+
+Upgrade from 1.6 directly to 2.x
```
veod32: May to  rename to 

Upgrade from 1.6 directly to 2.x with downtime

to make it clear what the case is (it this is the case, of course )))

veod32 (changes requested):

doc/book/admin/_includes/1.6-to-2.x-condition.rst
patiencedaur: This text is included both in the direct and stepwise instructions with the `include` directive. So I think the link won't be necessary.

doc/book/admin/upgrades.rst
patiencedaur: Тогда нам нужно обновить [доку по box.schema.upgrade](https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_schema/upgrade/) :)

doc/book/admin/upgrades.rst
patiencedaur: I think that it is unnecessary to divide the output of a single command into two separate steps.
Instead, I'll elaborate on the table output.

veod32 (approved):

PR #3102, opened 2022-08-17
[1.10] Put only document titles in the left navigation bar

NickVolynkin:
See https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html?highlight=toctree#directive-toctree

Part of tarantool/tarantool.io#1275

### DIFF
--- a/doc/toctree.rst
+++ b/doc/toctree.rst
@@ -1,5 +1,6 @@
 
-.. toctree::
+..  toctree::
+    :titlesonly:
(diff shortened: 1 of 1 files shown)


NickVolynkin:
There's no right navbar in the version 1.10, so it's too early to roll these changes.

NickVolynkin:
There's no right navbar in the version 1.10, so it's too early to roll these changes.

PR #3133, opened 2022-09-07
Support dynamic evaluation of synchronous replication quorum

### LINKED ISSUE #1730: Support dynamic evaluation of synchronous replication quorum
# Support dynamic evaluation of synchronous replication quorum

**Product:** Tarantool
**Since:** 2.5.3, 2.6.2, 2.7.1
**Audience/target:**
**Root document:**
- https://www.tarantool.io/en/doc/latest/reference/configuration/#cfg-replication-replication-synchro-quorum
- (possibly) https://www.tarantool.io/en/doc/latest/book/replication/repl_leader_elect/
- (possibly) https://www.tarantool.io/en/doc/latest/book/replication/repl_sync/
**SME:** @ cyrillos
**Peer reviewer:** @

# Details

Setting `replication_synchro_quorum` option to an explicit integer
value was introduced rather for simplicity sake mostly. For example
if the cluster's size is not a constant value and new replicas are
connected in dynamically then an administrator might need to increase
the option by hands or by some other external tool.

Instead one can use a dynamic evaluation of a quorum value via formal
representation using symbol `N` as a current number of registered replicas
in a cluster.

For example, the canonical definition for a quorum (that is, majority
of members in a set) of `N` replicas is `N/2+1`. For such configuration
define

```
box.cfg {replication_synchro_quorum = "N/2+1"}
```

The formal statement allows to provide a flexible configuration but keep
in mind that only canonical quorum (and bigger values, say `N` for all
replicas) guarantees data reliability and various weird forms such as
`N/3+1` while allowed may lead to unexpected results.
Requested by @cyrillos in https://github.com/tarantool/tarantool/commit/14fa5fd82422e2796e5de2e89eb8128ab5754465.

See other details in the issue comments below.

xuniq:
Fixes #1730 
Deploy:
https://docs.d.tarantool.io/en/doc/gh-1730-dynamic-evaluation/reference/configuration/#confval-replication_synchro_quorum
https://docs.d.tarantool.io/en/doc/gh-1730-dynamic-evaluation/book/replication/repl_leader_elect/#repl-leader-elect-splitbrain
https://docs.d.tarantool.io/en/doc/gh-1730-dynamic-evaluation/book/replication/repl_sync/#usage


doc/book/replication/repl_sync.rst
```diff
@@ -35,9 +35,16 @@ To control the behavior of synchronous transactions, there exist global
 
     box.cfg{replication_synchro_quorum = <number of instances>}
 
-This option tells how many replicas should confirm the receipt of a synchronous
-transaction before it is committed. Since version :doc:`2.10.0 </release/2.10.0>`,
-this option does not account for anonymous replicas. As a usage example, consider this:
+..  code-block:: lua
+
+    box.cfg{replication_synchro_quorum = "(cluster_size/2)+1"}
```
sergepetrenko: This example won't work. Better say
`box.cfg{replication_synchro_quorum = "N / 2 + 1"}`
like in other places.

doc/reference/configuration/cfg_replication.rst
```diff
@@ -428,12 +428,30 @@
...
+    Using a value less than the canonical one might lead to unexpected results,
+    including a :ref:`split-brain <repl_leader_elect_splitbrain>`.
+
     Since version :doc:`2.10.0 </release/2.10.0>`, this option
     does not account for anonymous replicas.
 
-    It is 1 by default, so synchronous transactions work like asynchronous when
-    not configured. 1 means successful WAL write on master is enough for
-    commit.
+    The default value for this parameter is `1`, so synchronous transactions work like asynchronous when not configured.
```
sergepetrenko: By the way, since 2.10.0 the default value is "N / 2 + 1". Let's reflect it here.
I couldn't find the corresponding doc request, must've forgotten to add it. But the corresponding commit is
https://github.com/tarantool/tarantool/commit/f0eb2ca3f9

sergepetrenko (changes requested):
Hi! Thanks for the patch!
Please find my comments below.

doc/reference/configuration/cfg_replication.rst
Totktonada: By the way (2), such behavior variations should be described like so:

> Since A.B.C: <...>.
>
> Since X.Y.Z: <...>.

sergepetrenko (approved):
Thanks!
Looks good.

PR #3157, opened 2022-09-20
Describe ``replication_threads`` configuration option

### LINKED ISSUE #2590: New configuration option - `replication_threads`
**Product:** Tarantool
**Since:** 2.10
**Root document:** https://www.tarantool.io/en/doc/latest/reference/configuration/#replication
**SME:** @ sergepetrenko

# Details

It's now possible to specify how many threads will be spawned to decode
the incoming replication stream.

The default value is '1', meaning a single thread will handle all
incoming replication streams.

You may set the value to anything from 1 to 1000. When there are
multiple replication threads, connections to serve are evenly
distributed between the threads.
Requested by @sergepetrenko in https://github.com/tarantool/tarantool/commit/dacbf708f4a4cf0a130a4733982b69955c611718.

xuniq:
Resolves #2590 
Deploy: https://docs.d.tarantool.io/en/doc/gh-2590-replication-threads/reference/configuration/#confval-replication_threads

### DIFF
--- a/doc/reference/configuration/cfg_replication.rst
+++ b/doc/reference/configuration/cfg_replication.rst
@@ -10,6 +10,7 @@
 * :ref:`instance_uuid <cfg_replication-instance_uuid>`
 * :ref:`replication_synchro_quorum <cfg_replication-replication_synchro_quorum>`
 * :ref:`replication_synchro_timeout <cfg_replication-replication_synchro_timeout>`
+* :ref:`replication_threads <cfg_replication-replication_replication_threads>`
 * :ref:`election_mode <cfg_replication-election_mode>`
 * :ref:`election_timeout <cfg_replication-election_timeout>`
 * :ref:`election_fencing_enabled <cfg_replication-election_fencing_enabled>`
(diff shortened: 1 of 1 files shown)


doc/reference/configuration/cfg_replication.rst
```diff
@@ -482,6 +482,23 @@
...
+    It means that a single separate thread handles all the incoming replication streams.
+
+    Possible values range from 1 to 1000.
+    If there are multiple replication threads, connections to serve are distributed evenly between the threads.
+
+    | Type: number
+    | Default: 1
+    | Possible values: from 1 to 1000
+    | Environment variable: TT_REPLICATION_THREADS
+    | Dynamic: **yes**
```
sergepetrenko: The option is not dynamic.

doc/reference/configuration/cfg_replication.rst
```diff
@@ -482,6 +482,23 @@
     | Environment variable: TT_REPLICATION_SYNCHRO_TIMEOUT
     | Dynamic: **yes**
 
+.. confval:: replication_threads
+
+    Since version :doc:`2.10.0 </release/2.10.0>`.
+    The number of threads spawned to decode the incoming replication data.
+
+    The default value is `1`.
+    It means that a single separate thread handles all the incoming replication streams.
```
sergepetrenko, line 494: Let's mention that a single thread for all the incoming data should be enough in almost all cases, so the user probably won't have to touch this config option.

sergepetrenko (changes requested):

sergepetrenko (approved):

PR #3338, opened 2023-02-16
Document the 'wal_queue_max_size' option

### LINKED ISSUE #1969: [2pt] new configuration option: 'wal_queue_max_size'
`wal_queue_max_size` puts a limit on the amount of concurrent write requests
submitted to WAL.
`wal_queue_max_size` is measured in number of bytes to be written (0
means unlimited, which was the default behaviour before).
The option only affects replica behaviour at the moment, and defaults
to 16 megabytes. The option limits the pace at which replica reads new
transactions from master.

Here's when the option comes in handy:

Before this option was introduced such a situation could be possible:
there are 2 servers, a master and a replica, and the replica is down for
some period of time. While the replica is down, master serves requests
at a reasonable pace, possibly close to its WAL throughput limit. Once the
replica reconnects, it has to receive all the data master has piled up and
there's no limit in speed at which master sends the data to replica, and,
without the option, there was no limit in speed at which replica submitted
corresponding write requests to WAL.

This lead to a situation when replica's WAL was never in time to serve the
requests and the amount of pending requests was constantly growing.
There was no limit for memory WAL write requests take, and this clogging
of WAL write queue could even lead to replica using up all the available
memory.

Now, when `wal_queue_max_size` is set, appliers will stop reading new
transactions once the limit is reached. This will let WAL process all the
requests that have piled up and free all the excess memory.
Requested by @sergepetrenko in https://github.com/tarantool/tarantool/commit/de93b44862818cc5ee78db999cdf0e80fc3492b5.

andreyaksenov:
Add the `wal_queue_max_size` option to the reference (https://github.com/tarantool/doc/issues/1969).

Resolves #1969

### DIFF
--- a/doc/concepts/engines/memtx.rst
+++ b/doc/concepts/engines/memtx.rst
@@ -55,7 +55,7 @@ For more details, see the ``box.slab`` module :doc:`reference </reference/refere
 .. image:: memtx/spaces_indexes.svg
 
 Also inside the TX thread, there is an event loop. Within the event loop, there are a number of :ref:`fibers <fiber-fibers>`.
-Fibers are cooperative primitives that allows interaction with spaces, that is, reading and writing the data.
+Fibers are cooperative primitives that allow interaction with spaces, that is, reading and writing the data.
 Fibers can interact with the event loop and between each other directly or by using special primitives called channels.
 Due to the usage of fibers and :ref:`cooperative multitasking <app-cooperative_multitasking>`, the ``memtx`` engine is lock-free in typical situations.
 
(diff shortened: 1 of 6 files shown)


sergepetrenko (approved):
Hi! Thanks for the patch!

LGTM.
I leave it for you to decide whether any explanations about raising or lowering `wal_queue_max_size` are needed.

doc/concepts/engines/memtx.rst
```diff
@@ -55,7 +55,7 @@ For more details, see the ``box.slab`` module :doc:`reference </reference/refere
 .. image:: memtx/spaces_indexes.svg
 
 Also inside the TX thread, there is an event loop. Within the event loop, there are a number of :ref:`fibers <fiber-fibers>`.
-Fibers are cooperative primitives that allows interaction with spaces, that is, reading and writting the data.
+Fibers are cooperative primitives that allows interaction with spaces, that is, reading and writing the data.
```
p7nov:
```suggestion
Fibers are cooperative primitives that allow interaction with spaces, that is, reading and writing the data.
```

p7nov (approved):
Great, just one minor fix (see the suggestion).

doc/concepts/engines/memtx.rst
andreyaksenov: thanks!

PR #3350, opened 2023-02-20
Document the 'wal_cleanup_delay' option

### LINKED ISSUE #2022: [3pt] Add wal_cleanup_delay configuration parameter
**Related dev. issue(s):** https://github.com/tarantool/tarantool/issues/5806

**Product:** Tarantool
**Since:** 2.8.1
**Root document:** https://www.tarantool.io/en/doc/latest/reference/configuration/#binary-logging-and-snapshots
**SME:** @ cyrillos

# Details

What should we write about? 
Describe the `wal_cleanup_delay` configuration option: 
- what it is doing
- what problem does it solve
- how to choose its value depending on the use case

---
The `wal_cleanup_delay` option defines a delay in second
before write ahead log files (`*.xlog`) are getting started
to prune upon a node restart.

This option is ignored in case if a node is running as
an anonymous replica (`replication_anon = true`). Similarly
if replication is unused or there are no plans to use
replication at all then this option should not be considered.

An initial problem to solve is the case where a node is operating
so fast that its replicas do not manage to reach the node state
and in case if the node is restarted at this moment (for various
reasons, for example due to power outage) then `*.xlog` files might
be pruned during restart. In result replicas will not find these
files on the main node and have to reread all data back which
is a very expensive procedure.

Since replicas are tracked via `_cluster` system space this we use
its content to count subscribed replicas and when all of them are
up and running the cleanup procedure is automatically enabled even
if `wal_cleanup_delay` is not expired.

The `wal_cleanup_delay` should be set to:

 - `0` to disable the cleanup delay;
 - `>= 0` to wait for specified number of seconds.

By default it is set to `14400` seconds (ie `4` hours).

In case if registered replica is lost forever and timeout is set to
infinity then a preferred way to enable cleanup procedure is not setting
up a small timeout value but rather to delete this replica from `_cluster`
space manually.

Note that the option does *not* prevent WAL engine from removing
old `*.xlog` files if there is no space left on a storage device,
WAL engine can remove them in a force way.

Current state of `*.xlog` garbage collector can be found in
`box.info.gc()` output. For example

``` Lua
 tarantool> box.info.gc()
 ---
   ...
   is_paused: false
```

The `is_paused` shows if cleanup fiber is paused or not.
Requested by @cyrillos in https://github.com/tarantool/tarantool/commit/2fd51aea271726fee8dc9380382d26ae48f4573a.

# Definition of done

- [ ] add an option description
- [ ] specify `Since version` with a link to release notes
- [ ] add links to related docs
- [ ] make sure all option properties (Type, Default, etc.) are specified
- [ ] add an option anchor to the top of the section

andreyaksenov:
Add the `wal_cleanup_delay` option to the reference.

Resolves #2022

### DIFF
--- a/doc/reference/configuration/cfg_binary_logging_snapshots.rst
+++ b/doc/reference/configuration/cfg_binary_logging_snapshots.rst
@@ -117,11 +117,13 @@
 
 .. confval:: wal_cleanup_delay
 
-    Since version :doc:`2.8.1 </release/2.8.1>`.
+    Since version :doc:`2.6.3 </release/2.6.3>`.
     The delay (in seconds) used to prevent the :ref:`Tarantool garbage collector <cfg_checkpoint_daemon-garbage-collector>`
     from immediately removing :ref:`write-ahead log<internals-wal>` files after a node restart.
-    This delay helps :ref:`replicas <replication-roles>` sync with a master faster after its restart and
-    eliminates the need to download all the data again.
+    This delay eliminates possible erroneous situations when the master deletes WALs
+    needed by :ref:`replicas <replication-roles>` after restart.
+    As a consequence, replicas sync with a master faster after its restart and
+    don't need to download all the data again.
 
     Once all the nodes in the replica set are up and running,
     automatic cleanup is started again even if ``wal_cleanup_delay`` has not expired.
--- a/doc/reference/configuration/cfg_binary_logging_snapshots.rst
+++ b/doc/reference/configuration/cfg_binary_logging_snapshots.rst
@@ -4,6 +4,7 @@
 * :ref:`wal_mode <cfg_binary_logging_snapshots-wal_mode>`
 * :ref:`wal_dir_rescan_delay <cfg_binary_logging_snapshots-wal_dir_rescan_delay>`
 * :ref:`wal_queue_max_size <cfg_binary_logging_snapshots-wal_queue_max_size>`
+* :ref:`wal_cleanup_delay <cfg_binary_logging_snapshots-wal_cleanup_delay>`
 
 .. _cfg_binary_logging_snapshots-force_recovery:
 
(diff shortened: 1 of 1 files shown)


doc/reference/configuration/cfg_binary_logging_snapshots.rst
```diff
@@ -111,3 +112,26 @@
     | Default: 16777216 bytes
     | Environment variable: TT_WAL_QUEUE_MAX_SIZE
     | Dynamic: **yes**
+
+.. _cfg_binary_logging_snapshots-wal_cleanup_delay:
+
+.. confval:: wal_cleanup_delay
+
+    Since version :doc:`2.8.1 </release/2.8.1>`.
```
sergepetrenko: The option already appeared in 2.6.3 (2.6.2-148-gd129729d9) and 2.7.2 (2.7.1-152-ge8c094644) (we had rolling releases back then, so 2.8.1 was released together with 2.7.2. and 2.6.3). This was considered a bugfix, so it was cherry-picked from 2.8.1 to the older versions.

doc/reference/configuration/cfg_binary_logging_snapshots.rst
```diff
@@ -111,3 +112,26 @@
...
+    The delay (in seconds) used to prevent the :ref:`Tarantool garbage collector <cfg_checkpoint_daemon-garbage-collector>`
+    from immediate removing :ref:`write-ahead log<internals-wal>` files after a node restart.
+    This delay helps :ref:`replicas <replication-roles>` sync with a master faster after its restart and
```
sergepetrenko: I wouldn't say it helps to sync faster: it eliminates the risk that master deletes WALs needed for replicas after a restart. It fixes a possible erroneous situation rather than improves anything.

Although, yes, this means replicas do not risk downloading the data all over again, and thus can sync faster.

sergepetrenko (changes requested):
Hi! Thanks for the patch!

Please see my comments below.

sergepetrenko (approved):
Thanks for the fixes! LGTM.

doc/reference/configuration/cfg_binary_logging_snapshots.rst
```diff
@@ -111,3 +112,28 @@
...
+.. _cfg_binary_logging_snapshots-wal_cleanup_delay:
+
+.. confval:: wal_cleanup_delay
+
+    Since version :doc:`2.6.3 </release/2.6.3>`.
+    The delay (in seconds) used to prevent the :ref:`Tarantool garbage collector <cfg_checkpoint_daemon-garbage-collector>`
+    from immediately removing :ref:`write-ahead log<internals-wal>` files after a node restart.
+    This delay eliminates possible erroneous situations when the master deletes WALs
+    needed by :ref:`replicas <replication-roles>` after restart.
+    As a consequence, replicas sync with a master faster after its restart and
```
p7nov: the master?

doc/reference/configuration/cfg_binary_logging_snapshots.rst
```diff
@@ -111,3 +112,28 @@
...
+    don't need to download all the data again.
+
+    Once all the nodes in the replica set are up and running,
+    automatic cleanup is started again even if ``wal_cleanup_delay`` has not expired.
+
+    .. NOTE::
+
+        The ``wal_cleanup_delay`` option is not in effect if a node is running as an
```
p7nov: If I get it right , this wording will be more precise:

```suggestion
        The ``wal_cleanup_delay`` option has no effect on nodes running as
```

I mean, it still has effect on other nodes in the replica set except anonymous replicas, right?

p7nov (approved):
Nice. Just one question

doc/reference/configuration/cfg_binary_logging_snapshots.rst
andreyaksenov: sure, thanks

doc/reference/configuration/cfg_binary_logging_snapshots.rst
andreyaksenov: Thanks, sounds better.

PR #3408, opened 2023-03-21
Document the PAP-SHA256 authentication protocol and new security enhancements

### LINKED ISSUE #3271: Document IPROTO_AUTH_TYPE - new key of IPROTO_ID
A new key was added to the `IPROTO_ID` response:
name `IPROTO_AUTH_TYPE`, code 0x5b, value type `MP_STR`.

Its value equals the name of the authentication method that is
currently used on the server for generating user authentication
data from a password `box.schema.passwd()`. It's configured with
`box.cfg.auth_type`.

In Community Edition, the only authentication method currently
available is 'chap-sha1'. Other authentication methods may be used
in Enterprise Edition.

A client can use the received value as the authentication method
that should be used by default, in case it isn't specified explicitly
along with the user credentials.
Requested by @locker in https://github.com/tarantool/tarantool/commit/13b0ce9c673d2ed9baea4de7714e8604e613ae94.

### LINKED ISSUE #3275: Document auth_history and last_modified _user space fields
Field name: auth_history. Field no: 6. Type: array.
Description: The field stores an array of previous authentication data:
when a user password is changed, the last value of the 'auth' field is
appended to 'auth_history'.  The length of the history is configured by
the `box.cfg.password_history_length` option, which is available only
in Tarantool EE, where it's used to prevent users from reusing old
passwords. In Tarantool CE, the array is always empty.

Field name: last_modified. Field no: 7. Type: unsigned.
Description: The field stores the timestamp (seconds since Unix epoch)
of the last user password update. It's never used in Tarantool CE.
In Tarantool EE, it's used to disable users that haven't changed the
password for more than `box.cfg.password_lifetime_days`.

`box.schema.upgrade()` sets the new field values to an empty array
and 0 for users that haven't updated them yet.
Requested by @locker in https://github.com/tarantool/tarantool/commit/1c33484d535ebe3ee53a681c29a719050b9ff30b.

andreyaksenov:
Added information about the PAP-SHA256 authentication protocol, which is available in Tarantool Enterprise:
- Added info about PAP-SHA256 to the `Access control | Passwords section`.
- Removed the phrase that `"chap-sha1" is the only supported mechanism right now` from all places.
- Added information about new fields related to authentication to the [box.space._user](https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_space/_user/) page.
- Added info about a new key of IPROTO_ID: IPROTO_AUTH_TYPE
- Some minor fixes (wording, Cyrillic characters).


doc/reference/reference_lua/box_schema/user_create.rst
```diff
@@ -37,6 +36,6 @@ box.schema.user.create()
 
     .. code-block:: lua
 
-        box.schema.user.create('Lena')
-        box.schema.user.create('Lena', {password = 'X'})
-        box.schema.user.create('Lena', {if_not_exists = false})
+        box.schema.user.create('testuser')
+        box.schema.user.create('testuser', {password = 'foobar'})
+        box.schema.user.create('testuser', {if_not_exists = false})
```
locker, line 41: What's wrong with "Lena"?

locker (approved):

doc/reference/reference_lua/box_schema/user_create.rst
andreyaksenov: Nothing wrong - just trying to use more neutral or well-known placeholder names (like John Doe) in docs.

doc/book/admin/access_control.rst
```diff
@@ -71,35 +71,28 @@ system space with a
...
-which the client must mix with the hashed-password before sending
-to the instance. Thus the original value ‘x’ is never stored anywhere except
-in the user’s head, and the hashed value is never passed down a network wire
-except when mixed with a random salt.
+Tarantool supports two protocols for authenticating users:
 
-.. NOTE::
+*   `CHAP <https://en.wikipedia.org/wiki/Challenge-Handshake_Authentication_Protocol>`_ with ``SHA-1`` hashing
+
+    In this case, password hashes are stored in the ``_user`` space unsalted.
```
p7nov: The salting concept is never introduced on this page; shouldn't we add 1-2 sentences about it somewhere before?

p7nov (approved):
LGTM, just one question

doc/book/admin/access_control.rst
andreyaksenov: good point... I think adding a link to this article would be enough: https://en.wikipedia.org/wiki/Salt_(cryptography)

PR #3588, opened 2023-07-19
Fix outdated default value of vinyl_write_threads

### LINKED ISSUE #1268: Update box.cfg default values
At least I see that [doc](https://www.tarantool.io/en/doc/1.10/reference/reference_lua/box_cfg/#box-cfg-default) says `vinyl_write_threads` by default is 2, meanwhile in fact default value of `vinyl_write_threads` is 4. Please check other default values as well.

p7nov:
Fixes #1268


ImeevMA (approved):
Thanks for the patch!

PR #3605, opened 2023-08-02
New msgpack api (2.11 and 2.10)

### LINKED ISSUE #2417: Document msgpack.cfg.encode_error_as_ext
**Product:** Tarantool
**Since:**
**Root document:** https://www.tarantool.io/en/doc/latest/reference/reference_lua/msgpack/#msgpack-cfg
**SME:** @ locker
**Peer reviewer:** @ 

# Details
The new option determines how error objects (see box.error.new) are 
encoded in the msgpack format:
 - If it's set, errors are encoded as the MP_ERROR msgpack extension.
   This is the default behavior.
 - If it's unset, the encoded format depends on other msgpack
   configuration options (encode_load_metatables, encode_use_tostring,
   encode_invalid_as_nil). With the otherwise default configuration,
   they are encoded as strings (see error.message).

Functions affected by the default configuration (msgpack.cfg):
 - msgpack and msgpackffi modules
 - Storing errors in tuples and spaces (box.tuple.new)
 - Returning errors from IPROTO CALL/EVAL

Requested by @locker in https://github.com/tarantool/tarantool/commit/fa1652dfec4d77ae86531747d87d8558a4deccdc

### LINKED ISSUE #2749: Document msgpack object iterator take_array method
**Root document**: https://www.tarantool.io/en/doc/latest/reference/reference_lua/msgpack/#msgpack-object-methods

The new method copies the given number of msgpack values starting from
the iterator cursor position to a new msgpack array object. On success
it returns the new msgpack object and advances the iterator cursor. If
there isn't enough values to decode, it raises a Lua error and leaves
the iterator cursor unchanged.

This function could be implemented in Lua like this:

```lua
function take_array(iter, count)
    local array = {}
    for _ = 1, count do
        table.insert(array, iter:take())
    end
    return msgpack.object(array)
end
```

Usage example:

```lua
local mp = msgpack.object({10, 20, 30, 40})
local it = mp:iterator()
it:decode_array_header()       -- returns 4
it:decode()                    -- returns 10
local mp2 = it:take_array(2)
mp2:decode()                   -- returns {20, 30}
it:decode()                    -- returns 40
```
Requested by @locker in https://github.com/tarantool/tarantool/commit/68bf38850b5573960a4f4d720d149d923eb0be32.

### LINKED ISSUE #3257: Document `msgpack.object:index`
**Related dev. issue(s):** tarantool/tarantool#7898

**Product:** Tarantool
**Since:** 2.11
**Root document:** https://www.tarantool.io/en/doc/latest/reference/reference_lua/msgpack/
**SME:** @ CuriousGeorgiy

# Details

For the API description and usage exmaples, see [design document](https://www.notion.so/tarantool/box-iproto-override-44935a6ac7e04fb5a2c81ca713ed1bce#73867e0f33c844f3822e62acdc4e7814).

Requested by @ CuriousGeorgiy in https://github.com/tarantool/tarantool/commit/dcc3fcd672ee67a39baf82f2b6b70eeea17270ed.

andreyaksenov:
Documented the following API members:
- `msgpack_object[key]` and the corresponding `msgpack_object:get(key)` function: 
    - https://docs.d.tarantool.io/en/doc/2.11-msgpack-api/reference/reference_lua/msgpack/#msgpack-object-item
    - https://docs.d.tarantool.io/en/doc/2.11-msgpack-api/reference/reference_lua/msgpack/#lua-function.msgpack_object.get
- The `iterator_object:take_array(count)` function: https://docs.d.tarantool.io/en/doc/2.11-msgpack-api/reference/reference_lua/msgpack/#lua-function.iterator_object.take_array
- The `cfg.encode_error_as_ext` configuration property: https://docs.d.tarantool.io/en/doc/2.11-msgpack-api/reference/reference_lua/msgpack/#msgpack-cfg

Note that updated docs reference testable [code snippets](https://github.com/tarantool/doc/tree/latest/doc/code_snippets) for making sure the code shown in docs is correct.


doc/reference/reference_lua/msgpack.rst
```diff
@@ -657,65 +658,100 @@ and :ref:`YAML <yaml-cfg>`.
...
+        :return: an :ref:`iterator object <msgpack-object-iterator-methods>` over the MsgPack data
 
         :rtype: userdata
 
+    ..  method:: get(key)
+
+        **Since:** :doc:`2.11.0 </release/2.11.0>`
+
+        Get an element of the MsgPack array by the specified index key.
+        You can also use the indexed notation (``[]``) to get an array element.
```
CuriousGeorgiy: Not sure what the standard approach for metamethods is, but I would also document the indexed notation separately and refer to `get`, pointing out that the indexed notation resolves method names collisions in favor of methods (e.g., `obj['iterator']` or `obj.iterator` will return the `:iterator` method function instead of returning the value of an `'iterator'` key which `obj:get('iterator')` would return) and providing an example for this corner case.

By the way, indexed notation has two forms, `[]` and `.`, so I guess you should either mention them both or not mention any of them specifically and refer to a general Lua term.

CuriousGeorgiy (changes requested):

doc/reference/reference_lua/msgpack.rst
```diff
@@ -241,7 +236,7 @@ Below is a list of all ``msgpack`` functions and members.
 
 .. function:: decode_array_header(byte-array, size)
 
-    Call the mp_decode_array function in the `MsgPuck <http://rtsisyk.github.io/msgpuck/>`_ library
+    Call the `MsgPuck <https://rtsisyk.github.io/msgpuck/>`_'s ``mp_decode_array`` function
```
locker, line 239: May be worth mentioning here that this function expects the data to be a MsgPack array. If it isn't, an exception is raised. Same comment for `decode_map_header`.

doc/reference/reference_lua/msgpack.rst
```diff
@@ -528,7 +538,7 @@ userdata/cdata, you can use this code:
 .. NOTE::
 
     To achieve the same effect for only one call to ``msgpack.encode()``
-    (i.e. without changing the configuration permanently), you can use
+    (that is without changing the configuration permanently), you can use
     ``msgpack.encode({1, x, y, 2}, {encode_invalid_numbers = true})``.
```
locker: This isn't true:

```
tarantool> msgpack.encode({1, 2}, {encode_invalid_numbers = true})
---
- error: 'msgpack.encode: argument 2 must be of type ''struct ibuf'''
...
```

It's possible to do this though:

```Lua
msgpack.new({encode_invalid_numbers = true}).encode({1, 2})
```

doc/reference/reference_lua/msgpack.rst
```diff
@@ -657,65 +658,100 @@ and :ref:`YAML <yaml-cfg>`.
...
+        Get an element of the MsgPack array by the specified index key.
+        You can also use the indexed notation (``[]``) to get an array element.
+
+        :param number/string key: the index key used to get the array element, which might be one of the following:
+
+                    *   if a MsgPack object is an array, ``key`` is the integer value (start with 1) that specifies the element index.
+                    *   if a MsgPack object is an associative array, ``key`` is the string value that specifies the element key.
+
+        :return: an element of the MsgPack array.
+                 If the specified key is missing in the array, `get` returns ``nil``.
```
locker: May be worth mentioning that this function raises an error in case the value stored in the object isn't a MsgPack array or map.

doc/code_snippets/test/msgpack/msgpack_object_iterator_map_test.lua
```diff
@@ -0,0 +1,20 @@
...
+
+local mp_map = msgpack.object({ foo = 'a',
+                                bar = { 'b1', 'b2', 'b3' } })
+local mp_map_iterator = mp_map:iterator()
+
+local size = mp_map_iterator:decode_map_header() -- returns 2
+local first = mp_map_iterator:decode()           -- returns 'foo'
+local second = mp_map_iterator:decode()          -- returns 'a'
+mp_map_iterator:skip()                           -- returns none, skips 'bar'
+local fourth = mp_map_iterator:decode()          -- returns ['b1', 'b2', 'b3']
```
locker: Strictly speaking, map elements will not necessarily be returned in this order. A Lua map is unordered so the order in which it's encoded in MsgPack depends on Lua internals. It may be better to rework the example to put decoded elements back into a map/set.

doc/reference/reference_lua/msgpack.rst
```diff
@@ -657,65 +658,100 @@ and :ref:`YAML <yaml-cfg>`.
 
     **Example:**
 
-    ..  code-block:: lua
+    ..  literalinclude:: /code_snippets/test/msgpack/msgpack_object_test.lua
+        :language: lua
+        :lines: 1,29-32
```
locker: Looks like you forgot line 5: line 31 refers to `mp_from_string` defined at line 5.

It's difficult to review these examples. IMO better inline them in the text.

doc/code_snippets/test/msgpack/msgpack_object_iterator_take_test.lua
```diff
@@ -0,0 +1,18 @@
+local msgpack = require('msgpack')
+
+local mp_array = msgpack.object({ 10, 20, 30 })
+local mp_array_iterator = mp_array:iterator()
+
+local size = mp_array_iterator:decode_array_header()  -- returns 3
+local first = mp_array_iterator:decode()              -- returns 10
+mp_array_iterator:skip()                              -- returns none, skips 20
+local mp_array_new = mp_array_iterator:take()
```
locker: Why is it called `mp_array_new`? It's a MsgPack object that contains 30, i.e. it doesn't have anything to with MsgPack arrays.

andreyaksenov: mixed up with `take_array`, thanks for pointing out

doc/reference/reference_lua/msgpack.rst
andreyaksenov: Thanks. Also found out the the example above with `httpc.curl` doesn't work:
```
tarantool> msgpack.encode(httpc.curl)
---
- error: unsupported Lua type 'userdata'
...

tarantool> msgpack.encode(httpc.curl, {encode_use_tostring=true})
---
- error: 'msgpack.encode: argument 2 must be of type ''struct ibuf'''
...
```

Changed to `msgpack.cfg{encode_use_tostring = true}`.

doc/reference/reference_lua/msgpack.rst
andreyaksenov: Thanks, fixed. We decided to use testable examples to make sure we don't miss any breaking changes that might break the code snippets used in docs. Looks like more 'atomic' examples that show only the minimal API usage would be more convenient for review - `msgpack_object_test.lua` actually mixes examples that can be split up.

doc/reference/reference_lua/msgpack.rst
andreyaksenov: Thanks for review!

> I would also document the indexed notation separately and refer to get

Added a separate description for [msgpack_object[key]](https://docs.d.tarantool.io/en/doc/2.11-msgpack-api/reference/reference_lua/msgpack/#msgpack-object-item) with a link to the `get` description. Converted the example to indexed notation.

> pointing out that the indexed notation resolves method names collisions in favor of methods

Added a note to `msgpack_object[key]` about possible name conflicts.

> providing an example for this corner case

Looks like a runnable example in docs is excessive in this case. I think, a note should be enough.

> By the way, indexed notation has two forms, [] and ., so I guess you should either mention them both or not mention any of them specifically and refer to a general Lua term.

Mentioned dot notation in the `msgpack_object[key]` description for associative arrays.

CuriousGeorgiy (approved):

doc/reference/reference_lua/msgpack.rst
```diff
@@ -63,16 +65,16 @@ Below is a list of all ``msgpack`` functions and members.
             -   Convert a raw MsgPack string to a Lua object
 
         *   -   :ref:`msgpack.decode_array_header(byte-array, size) <msgpack-decode_array_header>`
-            -   Skip an array header in a raw MsgPack string
+            -   Call the `MsgPuck <https://rtsisyk.github.io/msgpuck/>`_'s ``mp_decode_array`` function and return the array size and a pointer to the first array component
 
         *   -   :ref:`msgpack.decode_map_header(byte-array, size) <msgpack-decode_map_header>`
-            -   Skip a map header in a raw MsgPack string
+            -   Call the `MsgPuck <https://rtsisyk.github.io/msgpuck/>`_'s ``mp_decode_map`` function and return the map size and a pointer to the first map component.
```
xuniq:
```suggestion
            -   Call the `MsgPuck <https://rtsisyk.github.io/msgpuck/>`_'s ``mp_decode_map`` function and return the map size and a pointer to the first map component
```

doc/reference/reference_lua/msgpack.rst
```diff
@@ -301,8 +305,8 @@ The MsgPack output structure can be specified with the ``__serialize`` parameter
 
 * 'seq', 'sequence', 'array' - table encoded as an array
 * 'map', 'mappping' - table encoded as a map
-* function - the meta-method called to unpack serializable representation
-  of table, cdata or userdata objects
+* function - the meta-method called to unpack the serializable representation
```
xuniq:
```suggestion
* function -- the meta-method called to unpack the serializable representation
```

Same for lines 306-307

doc/reference/reference_lua/msgpack.rst
```diff
@@ -472,18 +486,18 @@ An array is excessively sparse when **all** the following conditions are met:
...
+By default, attempting to encode an excessively sparse array
+generates an error. If ``encode_sparse_convert`` is set to ``true``,
 excessively sparse arrays will be handled as maps.
 
 **msgpack.cfg() example 1:**
 
 If ``msgpack.cfg.encode_invalid_numbers = true`` (the default),
 then NaN and Inf are legal values. If that is not desirable, then
-ensure that ``msgpack.encode()`` will not accept them, by saying
+ensure that ``msgpack.encode()`` do not accept them, by saying
```
xuniq:
```suggestion
ensure that ``msgpack.encode()`` does not accept them, by saying
```

doc/reference/reference_lua/msgpack.rst
```diff
@@ -657,65 +667,126 @@ and :ref:`YAML <yaml-cfg>`.
...
+    ..  method:: msgpack_object[key]
+
+        **Since:** :doc:`2.11.0 </release/2.11.0>`
+
+        Get an element of the MsgPack array by the specified index key.
+        You can also use the :ref:`get(key) <msgpack-object-get>` method to get an array element.
+
+        The index key used to get the array element might be one of the following:
+
+        *   if a MsgPack object is an array, ``key`` is the integer value (start with 1) that specifies the element index.
```
xuniq:
```suggestion
        *   if a MsgPack object is an array, the ``key`` is an integer value (starting with 1) that specifies the element index.
```

doc/reference/reference_lua/msgpack.rst
```diff
@@ -657,65 +667,126 @@ and :ref:`YAML <yaml-cfg>`.
...
+    ..  method:: get(key)
+
+        **Since:** :doc:`2.11.0 </release/2.11.0>`
+
+        Get an element of the MsgPack array by the specified index key.
+        You can also use the indexed notation (:ref:`msgpack_object[key] <msgpack-object-item>`) to get an array element.
+
+        :param number/string key: the index key used to get the array element, which might be one of the following:
+
+                    *   if a MsgPack object is an array, ``key`` is the integer value (start with 1) that specifies the element index.
```
xuniq:
```suggestion
                    *   if a MsgPack object is an array, the ``key`` is an integer value (starting with 1) that specifies the element index.
```

xuniq (approved):

PR #3611, opened 2023-08-08
Document the 'checks' module

### LINKED ISSUE #3331: embedded checks
**Related dev. issue(s):** https://github.com/tarantool/tarantool/issues/7726

**Product:** Tarantool
**Since:** 2.11
**Root document:** https://www.tarantool.io/en/doc/latest/reference/reference_lua/
**SME:** @ DifferentialOrange

# Details
Now tarantool has checks module on its board. checks is a lua module
previously distributed as a separate rock which is widely used by
many other tarantool lua modules (like cartridge, metrics and crud) and
tarantool applications. checks has its own repo with README covering its
API usage: https://github.com/tarantool/checks/blob/master/README.md .
Requested by @ DifferentialOrange in https://github.com/tarantool/tarantool/commit/5b68b6d624060cbee40c684ad608a03e1b290682.

andreyaksenov:
Added a new page for the ``checks`` module:
https://docs.d.tarantool.io/en/doc/embedded-checks/reference/reference_lua/checks/

> Note that new documentation topics now references testable code snippets from the [code_snippets](https://github.com/tarantool/doc/tree/latest/doc/code_snippets) folder.

### DIFF
--- a/doc/reference/reference_lua/checks.rst
+++ b/doc/reference/reference_lua/checks.rst
@@ -1,0 +1,523 @@
+..  checks-module:
+
+Module checks
+=============
+
+**Since:** :doc:`2.11.0 </release/2.11.0>`
+
+The ``checks`` module provides the ability to check the types of arguments passed to a Lua function.
+You need to call the :ref:`checks(type_1, ...) <checks-checks>` function inside the target Lua function and pass :ref:`one or more <checks_number_of_arguments>` type qualifiers to check the corresponding argument types.
+There are two types of type qualifiers:
+
+-   A :ref:`string type qualifier <checks_string_type_qualifier>` checks whether a function's argument conforms to the specified type. Example: ``'string'``.
+-   A :ref:`table type qualifier <checks_table_type_qualifiers>` checks whether the values of a table passed as an argument conform to the specified types. Example: ``{ 'string', 'number' }``.
+
+.. NOTE::
+
+    For earlier versions, you can install the ``checks`` module from the `Tarantool rocks repository <https://www.tarantool.io/en/download/rocks>`_.
+
+.. _checks_loading:
+
+Loading checks
+--------------
+
+In Tarantool :doc:`2.11.0 </release/2.11.0>` and later versions, the :ref:`checks API <checks_api_reference>` is available in a script without loading the module.
+
+For earlier versions, you need to install the ``checks`` module from the Tarantool rocks repository and load the module using the ``require()`` directive:
+
+.. code-block:: lua
+
+    local checks = require('checks')
+
+
+.. _checks_number_of_arguments:
+
+Number of arguments to check
+----------------------------
+
+For each argument to check, you need to specify its own type qualifier in the :ref:`checks(type_1, ...) <checks-checks>` function.
+
+.. _checks_one_argument:
+
+One argument
+~~~~~~~~~~~~
+
+In the example below, the ``checks`` function accepts a ``string`` type qualifier to verify that only a string value can be passed to the ``greet`` function.
+Otherwise, an error is raised.
+
+..  literalinclude:: /code_snippets/test/checks/checks_string_test.lua
+    :language: lua
+    :end-before: -- Tests
+    :dedent:
+
+
+.. _checks_multiple_arguments:
+
+Multiple arguments
+~~~~~~~~~~~~~~~~~~
+
+To check the types of several arguments, you need to pass the corresponding type qualifiers to the ``checks`` function.
+In the example below, both arguments should be string values.
+
+..  literalinclude:: /code_snippets/test/checks/checks_string_multiple_test.lua
+    :language: lua
+    :end-before: -- Tests
+    :dedent:
+
+To skip checking specific arguments, use the :ref:`? placeholder <checks_any_type>`.
+
+
+.. _checks_variable_number_of_arguments:
+
+Variable number of arguments
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can check the types of explicitly specified arguments for functions that accept a variable number of arguments.
+
+..  literalinclude:: /code_snippets/test/checks/checks_vararg_test.lua
+    :language: lua
+    :end-before: -- Tests
+    :dedent:
+
+
+
+.. _checks_string_type_qualifier:
+
+String type qualifier
+---------------------
+
+This section describes how to check a specific argument type using a string type qualifier:
+
+- The :ref:`Supported types <checks_string_supported_types>` section describes all the types supported by the ``checks`` module.
+- If required, you can make a :ref:`union type <checks_union_type>` to allow an argument to accept several types.
+- You can make any of the supported types :ref:`optional <checks_optional_type>`.
+- To skip checking specific arguments, use the :ref:`? placeholder <checks_any_type>`.
+
+.. _checks_string_supported_types:
+
+Supported types
+~~~~~~~~~~~~~~~
+
+.. _checks_lua_types:
+
+Lua types
+*********
+
+A string type qualifier can accept any of the `Lua types <https://www.lua.org/pil/2.html>`_, for example, ``string``, ``number``, ``table``, or ``nil``.
+In the example below, the ``checks`` function accepts ``string`` to validate that only a string value can be passed to the ``greet`` function.
+
+..  literalinclude:: /code_snippets/test/checks/checks_string_test.lua
+    :language: lua
+    :end-before: -- Tests
+    :dedent:
+
+
+.. _checks_tarantool_types:
+
+Tarantool types
+***************
+
+You can use :ref:`Tarantool-specific types <index_box_field_type_details>` in a string qualifier.
+The example below shows how to check that a function argument is a decimal value.
+
+..  literalinclude:: /code_snippets/test/checks/checks_decimal_test.lua
+    :language: lua
+    :end-before: -- Tests
+    :dedent:
+
+This table lists all the checks available for Tarantool types:
+
+..  container:: table
+
+    ..  list-table::
+        :widths: 30 40 30
+        :header-rows: 1
+
+        *   -   Check
+            -   Description
+            -   See also
+
+        *   -   ``checks('datetime')``
+            -   Check whether the specified value is :ref:`datetime_object <datetime_obj>`
+            -   :ref:`checkers.datetime(value) <checks-checkers-datetime>`
+
+        *   -   ``checks('decimal')``
+            -   Check whether the specified value has the :ref:`decimal <decimal>` type
+            -   :ref:`checkers.decimal(value) <checks-checkers-decimal>`
+
+        *   -   ``checks('error')``
+            -   Check whether the specified value is :ref:`error_object <box_error-error_object>`
+            -   :ref:`checkers.error(value) <checks-checkers-error>`
+
+        *   -   ``checks('int64')``
+            -   Check whether the specified value is an ``int64`` value
+            -   :ref:`checkers.int64(value) <checks-checkers-int64>`
+
+        *   -   ``checks('interval')``
+            -   Check whether the specified value is :ref:`interval_object <interval_obj>`
+            -   :ref:`checkers.interval(value) <checks-checkers-interval>`
+
+        *   -   ``checks('tuple')``
+            -   Check whether the specified value is a :ref:`tuple <box_tuple-new>`
+            -   :ref:`checkers.tuple(value) <checks-checkers-tuple>`
+
+        *   -   ``checks('uint64')``
+            -   Check whether the specified value is a ``uint64`` value
+            -   :ref:`checkers.uint64(value) <checks-checkers-uint64>`
+
+        *   -   ``checks('uuid')``
+            -   Check whether the specified value is :ref:`uuid_object <uuid-module>`
+            -   :ref:`checkers.uuid(value) <checks-checkers-uuid>`
+
+        *   -   ``checks('uuid_bin')``
+            -   Check whether the specified value is :ref:`uuid <uuid-module>` represented by a 16-byte binary string
+            -   :ref:`checkers.uuid_bin(value) <checks-checkers-uuid_bin>`
+
+        *   -   ``checks('uuid_str')``
+            -   Check whether the specified value is :ref:`uuid <uuid-module>` represented by a 36-byte hexadecimal string
+            -   :ref:`checkers.uuid_str(value) <checks-checkers-uuid_str>`
+
+
+
+.. _checks_custom_function:
+
+Custom function
+***************
+
+A string type qualifier can accept the name of a custom function that performs arbitrary validations.
+To achieve this, create a function returning ``true`` if the value is valid and add this function to the :ref:`checkers <checks-checkers>` table.
+
+The example below shows how to use the ``positive`` function to check that an argument value is a positive number.
+
+..  literalinclude:: /code_snippets/test/checks/checks_function_test.lua
+    :language: lua
+    :end-before: -- Tests
+    :dedent:
+
+
+.. _checks_metatable_type:
+
+Metatable type
+**************
+
+A string qualifier can accept a value stored in the ``__type`` field of the argument `metatable <https://www.lua.org/pil/13.html>`_.
+
+..  literalinclude:: /code_snippets/test/checks/checks_metatable_type_test.lua
+    :language: lua
+    :end-before: -- Tests
+    :dedent:
+
+
+.. _checks_union_type:
+
+Union types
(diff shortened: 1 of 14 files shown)


doc/reference/reference_lua/checks.rst
```diff
@@ -0,0 +1,506 @@
+..  checks-module:
+
+Module checks
+=============
+
+**Since:** :doc:`2.11.0 </release/2.11.0>`
```
DifferentialOrange, line 6: 1. It is available as a separate rock for all relevant Tarantool versions (1.10+)
2. Embedded checks package is provided with core Tarantool since 2.11.**1**

doc/reference/reference_lua/checks.rst
```diff
@@ -0,0 +1,506 @@
...
+        *   -   **Functions**
+            -
+
+        *   -   :ref:`checks <checks-checks>`
+            -   When called inside a function, checks that the function's arguments conform to the specified types
+
+        *   -   **Objects**
+            -
+
+        *   -   :ref:`checkers <checks-checkers>`
```
DifferentialOrange, line 310: `checkers` is a global Lua variable (available disregarding whether one had called `require('checks')` in script or not). It's not clear from this documentation page.

doc/reference/reference_lua/checks.rst
```diff
@@ -0,0 +1,506 @@
...
+
+..  class:: checkers
+
+    Checkers for different types.
+    This object provides access to checkers for Tarantool-specific types.
+    You can also add a :ref:`custom checker <checks_custom_function>` that performs arbitrary validations.
+
+    .. _checks-checkers-datetime:
+
+    .. function:: datetime(value)
```
DifferentialOrange: It is compiled like `checkers:datetime` when it's actually `checkers.datetime`

doc/reference/reference_lua/checks.rst
```diff
@@ -0,0 +1,506 @@
...
+..  function:: checks(type_1, ...)
+
+    When called inside a function, checks that the function's arguments conform to the specified types.
+
+    :param string/table type_1: a :ref:`string <checks_string_type_qualifier>` or :ref:`table <checks_table_type_qualifiers>` type qualifier used to check the argument type
+    :param ...: optional type qualifiers used to check the types of :ref:`other arguments <checks_number_of_arguments>`
+
+..  _checks-checkers:
+
+checkers
```
DifferentialOrange, line 328: The only reason that `checkers` is public is that so user may add its own custom checkers. Existing checkers may be used as components for custom checkers, but there is no any other reason for one to call them explicitly. Maybe we should say that one shouldn't use `assert(checkers.datetime(arg))` instead of `checks('datetime')`

doc/reference/reference_lua/checks.rst
andreyaksenov: > It is available as a separate rock for all relevant Tarantool versions (1.10+)

Added a note after the introduction: https://docs.d.tarantool.io/en/doc/embedded-checks/reference/reference_lua/checks/.

> Embedded checks package is provided with core Tarantool since 2.11.1

Could you please clarify what do you mean by `Embedded checks package`? I have 2.11.0 on my machine and I can use `checks` without loading it explicitly:

```
tarantool
Tarantool 2.11.0-0-g247a9a4
type 'help' for interactive help
tarantool> checkers.
checkers.datetime(  checkers.error(     checkers.interval(  checkers.uint64(    checkers.uuid_bin(  
checkers.decimal(   checkers.int64(     checkers.tuple(     checkers.uuid(      checkers.uuid_str(  
```

doc/reference/reference_lua/checks.rst
andreyaksenov: Added a separate section about loading the module: https://docs.d.tarantool.io/en/doc/embedded-checks/reference/reference_lua/checks/#loading-checks

doc/reference/reference_lua/checks.rst
andreyaksenov: Fixed: https://docs.d.tarantool.io/en/doc/embedded-checks/reference/reference_lua/checks/#checkers

doc/reference/reference_lua/checks.rst
andreyaksenov: Updated the description to clarify that its main purpose is adding a custom checker: https://docs.d.tarantool.io/en/doc/embedded-checks/reference/reference_lua/checks/#checkers

doc/reference/reference_lua/checks.rst
DifferentialOrange: Oh, it seems that I was wrong and this one was cherry-picked to 2.11.0. I had thought that the product approach would be the same as with `metrics`, which will appear only in 2.11.1.

doc/reference/reference_lua/checks.rst
DifferentialOrange: Well, let's mention `checkers` in this sentence too.

DifferentialOrange: (To be honest, I had forgot that `checks` are global too.)

DifferentialOrange (approved):
LGTM from my side.

The documentation seems really nice and useful. It would be good to add a documentation page link to the [`checks`](https://github.com/tarantool/checks) repo README after merge.

andreyaksenov:
> LGTM from my side.
> 
> The documentation seems really nice and useful. It would be good to add a documentation page link to the [`checks`](https://github.com/tarantool/checks) repo README after merge.

Thanks, will make a PR after merging.

doc/reference/reference_lua/checks.rst
```diff
@@ -0,0 +1,516 @@
+..  checks-module:
+
+Module checks
+=============
+
+**Since:** :doc:`2.11.0 </release/2.11.0>`
+
+The ``checks`` module provides the ability to check the types of arguments passed to a Lua function.
+You need to call the :ref:`checks(type_1, ...) <checks-checks>` function inside the target Lua function and pass :ref:`one or more <checks_number_of_arguments>` type qualifiers to check the corresponding argument types.
```
p7nov, line 9: Nit: I'd start with "to check a function's argument types, call ..." to be more goal-oriented. In the current wording, the path to reader's goal is too long.

doc/reference/reference_lua/checks.rst
```diff
@@ -0,0 +1,516 @@
...
+There are two types of type qualifiers:
+
+-   A :ref:`string type qualifier <checks_string_type_qualifier>` checks whether a function's argument conforms to the specified type.
+-   A :ref:`table type qualifier <checks_table_type_qualifiers>` checks whether the values of a table passed as an argument conform to the specified types.
```
p7nov: Maybe add short examples directly in the text?

doc/reference/reference_lua/checks.rst
```diff
@@ -0,0 +1,516 @@
...
+.. NOTE::
+
+    For earlier versions, you can install the ``checks`` module from the `Tarantool rocks repository <https://www.tarantool.io/en/download/rocks>`_.
+
+.. _checks_loading:
+
+Loading checks
+--------------
+
+The :ref:`checks API <checks_api_reference>` is available in a script without loading the module using ``require('checks')``.
```
p7nov: Maybe say that `require` is needed only if the module is installed from repo?

I actually don't get how our devs decide which modules need `require` and which don't, but the explicit sentence "this module doesn't need require" looks a bit awkward.

doc/reference/reference_lua/checks.rst
```diff
@@ -0,0 +1,516 @@
...
+
+
+.. _checks_number_of_arguments:
+
+Number of arguments to check
+----------------------------
+
+Depending on the number of arguments to be checked, you can pass one or more type qualifiers to the :ref:`checks(type_1, ...) <checks-checks>` function.
```
p7nov: The sentence reads a bit too formal.
Maybe smth simpler, like: "For each argument to check, you need to specify its own type qualifier"?

doc/reference/reference_lua/checks.rst
```diff
@@ -0,0 +1,516 @@
...
+
+To check the types of several arguments, you need to pass the corresponding type qualifiers to the ``checks`` function.
+In the example below, both arguments should be string values.
+
+..  literalinclude:: /code_snippets/test/checks/checks_string_multiple_test.lua
+    :language: lua
+    :end-before: -- Tests
+    :dedent:
+
+To turn off checking of specific arguments, use the :ref:`? placeholder <checks_any_type>`.
```
p7nov: Do I get it right that `?` is for checking arbitrary subsets of arguments? For example, in `f(a, b, c, d)`, check only `a` and `d`? If yes, then maybe show an example?

p7nov: Also, `skip checking` seems better to me here than `turn off checking`.

doc/reference/reference_lua/checks.rst
```diff
@@ -0,0 +1,516 @@
...
+
+
+.. _checks_string_type_qualifier:
+
+String type qualifier
+---------------------
+
+This section describes how to check a specific argument type using a string type qualifier:
+
+- The :ref:`Supported types <checks_string_supported_types>` section describes all the types supported by the ``checks`` module.
```
p7nov, line 91: This list lacks consistency, which makes it hard to read. Maybe rephrase?

doc/reference/reference_lua/checks.rst
```diff
@@ -0,0 +1,516 @@
...
+
+Supported types
+~~~~~~~~~~~~~~~
+
+.. _checks_lua_types:
+
+Lua types
+*********
+
+A string type qualifier can accept any of the `Lua types <https://www.lua.org/pil/2.html>`_, for example, ``string``, ``number``, ``table``, or ``nil``.
```
p7nov, line 106: Why not just "can be any Lua type"? It is easier to read.

doc/reference/reference_lua/checks.rst
```diff
@@ -0,0 +1,516 @@
...
+    ..  list-table::
+        :widths: 30 40 30
+        :header-rows: 1
+
+        *   -   Check
+            -   Description
+            -   See also
+
+        *   -   ``checks('datetime')``
+            -   Check whether the specified value is :ref:`datetime_object <datetime_obj>`
```
p7nov, line 141: I don't see much point in repeating the same sentence in all rows changing only the type (which is obvious from the first column). Can we restructure the table somehow?

doc/reference/reference_lua/checks.rst
```diff
@@ -0,0 +1,516 @@
...
+
+
+
+.. _checks_custom_function:
+
+Custom function
+***************
+
+A string type qualifier can accept the name of a custom function that performs arbitrary validations.
+To achieve this, the function should be added to the :ref:`checkers <checks-checkers>` table and should return ``true`` if the value is valid.
```
p7nov: Maybe rephrase to direct instruction: "write a function and add to the table"?

doc/reference/reference_lua/checks.rst
```diff
@@ -0,0 +1,516 @@
...
+    :language: lua
+    :end-before: -- Tests
+    :dedent:
+
+As for a specific type, you can make a :ref:`union type <checks_union_type>` value optional: ``?number|string``.
+
+
+.. _checks_any_type:
+
+Turning off argument checking
```
p7nov: Turn off > Skip?

doc/reference/reference_lua/checks.rst
```diff
@@ -0,0 +1,516 @@
...
+
+
+
+
+.. _checks_table_type_qualifiers:
+
+Table type qualifier
+--------------------
+
+The type qualifier may be a table. In this case, the following checks are made:
```
p7nov: This description wasn't enough for me to understand what this is used for. Maybe write that it's for cases when arguments are tables?

doc/reference/reference_lua/checks.rst
```diff
@@ -0,0 +1,516 @@
...
+In the next example, the same checks are made for the specified keys.
+
+..  literalinclude:: /code_snippets/test/checks/checks_table_keys_test.lua
+    :language: lua
+    :end-before: -- Tests
+    :dedent:
+
+.. NOTE::
+
+    A table type qualifier may be nested and use tables, too.
```
p7nov: "Table qualifiers can be nested"?

p7nov (approved):
Generally OK, but some places seem hard to read; I suggested simpler/shorter wordings.

doc/reference/reference_lua/checks.rst
DifferentialOrange: > I actually don't get how our devs decide which modules need require and which don't

Someone just decides that's `rawset(_G, 'checks', checks)` in the module is good idea five years ago and then no one breaks it since they don't have a reason to do so. Actually, in case of repo-installed `checks` you also don't need to require it more than once. Once you have a require in any of your scripts, it is available in any code called later. It's just that built-in modules are required inside the core Tarantool code, so the first require already had happened for any application developer.

Actually, I prefer calling `local checks = require('checks')` in any script, but, well, it is a fact that it is not actually needed. My original comment was related to `checkers` -- it is available in global after first `require('checks')` (which had happen already internally), but you cannot do `require('checkers')` -- it is only in global.

doc/reference/reference_lua/checks.rst
DifferentialOrange: > Do I get it right that `?` is for checking arbitrary subsets of arguments? For example, in `f(a, b, c, d)`, check only `a` and `d`? If yes, then maybe show an example?

Input count must match with checks expected count (excluding varargs).

```lua
tarantool> g = function(a, b) checks('number', 'number') end
---
...

tarantool> g(1, 2)
---
...

tarantool> g(1)
---
- error: '[string "g = function(a, b) checks(''number'', ''number'')..."]:1: bad argument
    #2 to nil (number expected, got nil)'
... 

tarantool> g = function(a, b) checks('number') end
---
...

tarantool> g(1, 2)
---
- error: '[string "g = function(a, b) checks(''number'') end"]:1: checks: argument
    "b" is not checked'
...

tarantool> g = function(a, b) checks('number', '?') end
---
...

tarantool> g(1)
---
...

```

So, in case you want to check some args and you do not care about other args, you must check them with `'?'` since you cannot simply skip them. It makes even more sence in case you don't want to check some argument in the middle. It is similar for table fields -- if you want to allow optional field X with arbitrary value, you use `'?'`.

doc/reference/reference_lua/checks.rst
andreyaksenov: Consistency is good for lists with more 'homogeneous' items. This list shows different features of string qualifiers and I think it's okay to describe them more like a regular text but using a list for convenience.

doc/reference/reference_lua/checks.rst
andreyaksenov: Strictly speaking, `type qualifier != type`. A type qualifier can accept one type, several types, the `?` value. `Accept` sounds more technically correct here.

doc/reference/reference_lua/checks.rst
andreyaksenov: Auto-generated lists of  API members often look like this. I think, it's not a problem.

doc/reference/reference_lua/checks.rst
andreyaksenov: Not sure am I right or not but to me `skip` has a slightly different meaning and suits better for cases when we have some kind of a loop. Here, we turn off checking of the specified argument.

doc/reference/reference_lua/checks.rst
andreyaksenov: Replaced with a sentence from intro.

doc/reference/reference_lua/checks.rst
p7nov: I see it this way: 
- by default, no args are checked - checking is _turned off_ for all of them.
- _turn off_ checking means return the default (no check) after _turning on_ checking.
- but this text tells about the case when nobody turned on checking for these args :) 

"skip" came to me because I see args as a sequence. So I read a `check` call with a `?` like this: check `arg1`, check `arg2`, skip `arg3`, check `arg4`.

doc/reference/reference_lua/checks.rst
andreyaksenov: > Do I get it right that ? is for checking arbitrary subsets of arguments? For example, in f(a, b, c, d), check only a and d? If yes, then maybe show an example?

You are right. Will update the example to make it more clear.

doc/reference/reference_lua/checks.rst
andreyaksenov: Now I got this :) Will update

PR #3622, opened 2023-08-17
Add page for mixed transactions recovery

### LINKED ISSUE #3526: correct recovery of mixed transactions
**Related dev. issue(s):** https://github.com/tarantool/tarantool/issues/7932

**Product:** Tarantool
**Since:** 2.11
**Root document:** <s>https://www.tarantool.io/en/doc/latest/reference/configuration/#confval-force_recovery</s>
 new page in https://www.tarantool.io/en/doc/latest/book/admin/upgrades/
**SME:** @ yanshtunder

# Details
In this patch implemented correct recovery of mixed transactions. To do
this, set  `box.cfg.force_recovery` to `true`. If you need to revert to
the old behavior, don't set the `force_recovery` option.

What to do when one node feeds the other a xlog with mixed transactions?
Let there be two nodes (`node#1` and `node#2`). And let the data be
replicated from `node#1` to `node#2`. Suppose that at some point in time,
`node#1` is restoring data from an xlog containing mixed transactions. To
replicate data from `node#1` to `node#2`, do the following:
1. Stop `node#2` and delete all xlog files from it
2. Restart `node#1` by setting `force_recovery` to `true`
3. Make `node#2` rejoin to `node#1` again.
Requested by @ yanshtunder in https://github.com/tarantool/tarantool/commit/2b1c8713b54f94b0fd3c6e106426b583c06f64ad.

p7nov:
Resolves #3526 

Deployment: https://docs.d.tarantool.io/en/doc/gh-3526-mixed-transactions/book/admin/upgrades/2.11.0/

### DIFF
--- a/doc/book/admin/upgrades/2.11.0.rst
+++ b/doc/book/admin/upgrades/2.11.0.rst
@@ -27,7 +27,7 @@ version due to a recovery error:
 1.  Start these instances with the :ref:`force_recovery <cfg_binary_logging_snapshots-force_recovery>`
     option to ``true``.
 2.  Make new snapshots on the instances so that the old WALs with mixed transactions
-    aren't used for recovery anymore.
+    aren't used for recovery anymore. To do this, call :ref:`box.snapshot() <box-snapshot>`.
 3.  Set ``force_recovery`` back to ``false``.
 
 Replication doesn't work
--- a/doc/book/admin/upgrades/2.11.0.rst
+++ b/doc/book/admin/upgrades/2.11.0.rst
@@ -18,10 +18,22 @@ from such a WAL.
 Starting from version :doc:`2.11.0 </release/2.11.0>`, Tarantool can recover from
 WALs with mixed transactions in the ``force_recovery`` mode.
 
-To restore data from a WAL with mixed transactions in a replica set:
+Instances fail to start
+-----------------------
 
-1.  Stop read-only instances and delete their WALs.
-2.  Restart the master instance setting the :ref:`force_recovery <cfg_binary_logging_snapshots-force_recovery>`
-    option to ``true`` on it.
-3.  Start read-only instances and wait for them to join the master.
+If all instances or some of them fail to start after upgrading to 2.11 due to
+a recovery error:
 
+1.  Start these instances with the :ref:`force_recovery <cfg_binary_logging_snapshots-force_recovery>`
+    option to ``true``.
+2.  Make new snapshots on the instances, so that the old snapshots aren't used
+    for recovery anymore.
+3.  Set ``force_recovery`` back to ``false.
+
+Replication doesn't work
+------------------------
+
+After all the instances are started successfully, WALs with mixed transactions
+may still lead to replication issues. If some instances can't replicate from other
+instances because they are sending incorrect WALs. To fix the replication issues,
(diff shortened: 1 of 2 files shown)


doc/book/admin/upgrades/2.11.0.rst
```diff
@@ -0,0 +1,27 @@
...
+between a transaction's ``begin`` and ``commit`` operations to belong to one transaction.
+If there is an operation belonging to another transaction, Tarantool fails to recover
+from such a WAL.
+
+Starting from version :doc:`2.11.0 </release/2.11.0>`, Tarantool can recover from
+WALs with mixed transactions in the ``force_recovery`` mode.
+
+To restore data from a WAL with mixed transactions in a replica set:
+
+1.  Stop read-only instances and delete their WALs.
```
sergepetrenko: Let's rewrite this guide a bit. There are two separate problems:

1st problem: all or part of instances updated to 2.11 won't start due to recovery error with mixed transactions
In this case all the user has to do is start the faulty nodes with `force_recovery` = true.
After recovering, the user has to make a snapshot on all the faulty nodes, so that the broken xlogs are not used for recovery anymore. After that `force_recovery` may be reset to false.

2nd problem:
May or may not emerge after the 1st one is solved. Here some of the replicas can't replicate from other replicas, because they are sending broken xlogs.
In this case indeed the user has to rebootstrap all the replicas, who can't connect to others (because others have new data, but in a broken xlog).
After the rebootstraps are done, everything should work as expected.

sergepetrenko (changes requested):
Thanks for the patch!

doc/book/admin/upgrades/2.11.0.rst
```diff
@@ -0,0 +1,39 @@
...
+
+Instances fail to start
+-----------------------
+
+If all instances or some of them fail to start after upgrading to 2.11 or a newer
+version due to a recovery error:
+
+1.  Start these instances with the :ref:`force_recovery <cfg_binary_logging_snapshots-force_recovery>`
+    option to ``true``.
+2.  Make new snapshots on the instances so that the old snapshots with mixed transactions
```
sergepetrenko: > so that the old snapshots with mixed transactions

Should be "so that the old xlogs with mixed transactions".

sergepetrenko (changes requested):
Thanks for the fixes! Just one remark from me.

sergepetrenko (approved):
Thanks!

LGTM.

doc/book/admin/upgrades/2.11.0.rst
```diff
@@ -0,0 +1,39 @@
...
+
+Instances fail to start
+-----------------------
+
+If all instances or some of them fail to start after upgrading to 2.11 or a newer
+version due to a recovery error:
+
+1.  Start these instances with the :ref:`force_recovery <cfg_binary_logging_snapshots-force_recovery>`
+    option to ``true``.
+2.  Make new snapshots on the instances so that the old WALs with mixed transactions
```
andreyaksenov, line 29: Maybe worth mentioning how to do this (AFAIU, using https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_snapshot/)

andreyaksenov (approved):

PR #3643, opened 2023-08-25
Document SQL_EXPR

### LINKED ISSUE #3245: SQL: document SQL_EXPR functions
**Product:** Tarantool
**Since:** 2.11
**Root document:** https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_schema/func_create/
https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_sql/
https://www.tarantool.io/en/doc/latest/reference/reference_sql/sql_statements_and_clauses/
https://www.tarantool.io/en/doc/latest/reference/reference_sql/sql_plus_lua/#calling-lua-routines-from-sql
**SME:** @ ImeevMA

# Details

Now it is possible to create functions with `SQL_EXPR` language. The
`body` field of the function definition is the SQL expression.
Currently, these functions can only be used as constraints on tuples and
fields, as there is no interface to define argument names and types
otherwise.

If the SQL_EXPR function is set as a constraint on a field, the function
will throw a validation error if more than one variable is used in the
function. However, the variable name of the only variable will only be
checked if the SQL_EXPR function was created as a field constraint from
SQL.
Requested by @ImeevMA in https://github.com/tarantool/tarantool/commit/1e099019d6699ce03833c01329b913d005198613.

### LINKED ISSUE #3624: feedback: box.schema.func.create() | Tarantool
> <…>function data, see the
> reference for the box.space._func space.
> **|The possible options are|:**
> 
> if_not_exists = true|false (default = false) –
> true means the<…>

https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_schema/func_create/

Some SQL options that can be passed to `func_create` are not listed here: `exports`, `param_list`, `returns`. 

Source: https://github.com/tarantool/tarantool/blob/28ec245d0a7ed4f7f07547ba0d22724db83df2a5/src/box/lua/schema.lua#L3047

andreyaksenov:
- [box.schema.func.create()](https://docs.d.tarantool.io/en/doc/sql-expr-func/reference/reference_lua/box_schema/func_create/): refactored a function description to remove duplicated info. Created a separate `function_options` table description. The function description now contains the SQL_EXPR example for a tuple constraint. The [language](https://docs.d.tarantool.io/en/doc/sql-expr-func/reference/reference_lua/box_schema/func_create/#lua-data.function_options.language) attribute description is extended with the new `SQL_EXPR` value.
- Mentioned the possibility to use SQL expressions in the [Constraint functions](https://docs.d.tarantool.io/en/doc/sql-expr-func/concepts/data_model/value_store/#constraint-functions) section.
- Removed unnecessary info from [box.space._func](https://docs.d.tarantool.io/en/doc/sql-expr-func/reference/reference_lua/box_space/_func/) as this info related to other API members and already documented in other places.


doc/code_snippets/test/constraints/constraint_sql_expr_test.lua
```diff
@@ -0,0 +1,62 @@
...
+            -- insert_name_error_end
+        end)
+
+        -- Tests --
+        t.assert_equals(customers:count(), 1)
+        t.assert_equals(customers:get(1), { 1, "Alice", 30 })
+        t.assert_equals(age_err:unpack().message, 'Check constraint \'check_person\' failed for tuple')
+        t.assert_equals(name_err:unpack().message, 'Check constraint \'check_person\' failed for tuple')
+    end)
+end
```
ImeevMA: Please add a newline.

doc/code_snippets/test/constraints/constraint_sql_expr_test.lua
```diff
@@ -0,0 +1,62 @@
...
+    cg.server:drop()
+end)
+
+g.test_constraints = function(cg)
+    cg.server:exec(function()
+        -- create_constraint_start
+        box.schema.func.create('check_person', {
+            language = 'SQL_EXPR',
+            is_deterministic = true,
+            body = [["age" > 21 AND "name" !='Admin']]
```
ImeevMA: It might be better to add space between `!=` and `Admin`.

doc/concepts/data_model/value_store.rst
```diff
@@ -598,11 +598,11 @@ a wider range of limitations.
 Constraint functions
 ~~~~~~~~~~~~~~~~~~~~
 
-Constraints use stored Lua functions, which must return ``true`` when the constraint
+Constraints use stored Lua functions or :ref:`SQL expressions <sql_expressions>`, which must return ``true`` when the constraint
```
ImeevMA, line 601: Not sure it's worth mentioning that the Lua and SQL_EXPR functions can be used as constraints. For example, C functions can also be used as constraints. Maybe not mention it here?

ImeevMA (approved):
Hi! Thank you for the patch. I have three comments, but they are all minor and can be ignored.

Also, I didn't mention it, but it would be nice to add that SQL_EXPR functions can be created when creating a SQL CHECK constraint using SQL. Moreover, if the above constraint is dropped, the SQL_EXPR function is not removed, which may not be obvious.

doc/reference/reference_lua/box_schema/func_create.rst
```diff
@@ -6,38 +6,235 @@ box.schema.func.create()
...
+
+.. function:: box.schema.func.create(func_name [, function_options])
+
+    Create a function.
+    The created function can be used in different usage scenarios,
+    for example, in :ref:`field or tuple constraints <index-constraints>` or
+    :ref:`functional indexes <box_space-index_func>`.
+
+    Using the :ref:`body <function_options_body>` option, you can make a function "persistent".
+    In this case, the function is "persistent" because its definition is stored in a snapshot (the :ref:`box.space._func <box_space-func>` system space) and can be recovered if the server restarts.
```
xuniq:
```suggestion
    Using the :ref:`body <function_options_body>` option, you can make a function *persistent*.
    In this case, the function is persistent because its definition is stored in a snapshot (the :ref:`box.space._func <box_space-func>` system space) and can be recovered if the server restarts.
```

doc/reference/reference_lua/box_schema/func_create.rst
```diff
@@ -6,38 +6,235 @@ box.schema.func.create()
...
+        | Type: string
+        | Default: ``LUA``
+
+    ..  _function_options_is_sandboxed:
+
+    .. data:: is_sandboxed
+
+        Whether the function should be executed in an isolated environment.
+        This means that any operation that accesses the world outside the sandbox is forbidden or has no effect.
+        Therefore a sandboxed function can only use modules and functions
```
xuniq:
```suggestion
        Therefore, a sandboxed function can only use modules and functions
```

xuniq (approved):
The new structure seems good to me.

PR #3645, opened 2023-08-29
Document SQL SEQSCAN

### LINKED ISSUE #3281: SQL: document a new SEQSCAN keyword for the SELECT scan queries
**Product:** Tarantool
**Since:** 2.11
**SME:** @ ImeevMA
**Root document:** https://www.tarantool.io/en/doc/latest/reference/reference_sql/sql_statements_and_clauses/#sql-select
https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_space/_session_settings/

Also, check these links:
* https://www.tarantool.io/en/doc/latest/how-to/sql/sql_tutorial/
* https://www.tarantool.io/en/doc/latest/reference/reference_sql/sql-features/
* https://www.tarantool.io/en/doc/latest/how-to/sql/sql_beginners_guide/

Keep in mind that the default value for sql_seq_scan session setting is also described in the compat module docs (since v.3.0): https://www.tarantool.io/en/doc/latest/reference/reference_lua/compat/sql_seq_scan_default/

# Details
Now scanning SELECT will not run and will throw an error if the new
SEQSCAN keyword is not used for scanned spaces. This change only affects
SELECT and does not affect UPDATE and DELETE. A SELECT is recognized as
a scanning SELECT if `EXPLAIN QUERY PLAN SELECT ...` indicates that the
SELECT `scans` rather than `searches`.

For example, if we have spaces created with these queries:
```
CREATE TABLE t(i INT PRIMARY KEY, a INT);
CREATE TABLE s(i INT PRIMARY KEY, a INT);
```

Then these queries will throw an error:
```
SELECT * FROM t;
SELECT * FROM t WHERE a > 1;
SELECT * FROM t WHERE i + 1 = 5;
SELECT * FROM t, s;
SELECT * FROM t JOIN s;
```

And these will not:
```
SELECT * FROM t WHERE i > 1;
SELECT * FROM SEQSCAN t;
SELECT * FROM SEQSCAN t WHERE i + 1 = 5;
SELECT * FROM SEQSCAN t, SEQSCAN s;
SELECT * FROM SEQSCAN t JOIN SEQSCAN s;
```

Scanning can be allowed or disallowed by default. To do this, a new
session setting is introduced: `sql_seq_scan`. The default value for
setting is `true`, i.e. scanning is allowed. When set to `false`, the
scanning SELECTs will throw a `scanning is not allowed` error.
Requested by @ImeevMA in https://github.com/tarantool/tarantool/commit/77648827326ad268ec0ffbcd620c2371b65ef2b4.

# Definition of done

- [x] SQL reference contains and describes the new keyword `SEQSCAN` in the [FROM clause](https://www.tarantool.io/en/doc/latest/reference/reference_sql/sql_statements_and_clauses/#from-clause) section.
- [x] [Session settings list](https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_space/_session_settings/) contains the new setting `sql_seq_scan`.
- [x] All SQL `SELECT` examples across the docs (mostly in [SQL how-tos](https://www.tarantool.io/en/doc/latest/how-to/sql/)) contain the SEQSCAN keyword if the do perform the sequential scanning.
- [x] S[QL user guide](https://www.tarantool.io/en/doc/latest/reference/reference_sql/sql_user_guide/) mentions the `SEQSCAN` keyword.

p7nov:
Resolves #3281 

- Added the SEQSCAN description to the FROM clause reference: https://docs.d.tarantool.io/en/doc/gh-3281-sql-seqscan/reference/reference_sql/sql_statements_and_clauses/#from-clause
- Added SEQSCAN to the list of reserved words in the SQL user guide: https://docs.d.tarantool.io/en/doc/gh-3281-sql-seqscan/reference/reference_sql/sql_user_guide/#reserved-words
- Added the SEQSCAN section to the SQL tutorial and updated all SELECT examples across this tutorial with SEQSCAN (where it is required): https://docs.d.tarantool.io/en/doc/gh-3281-sql-seqscan/how-to/sql/sql_tutorial/#the-seqscan-keyword
- Added a note about allowing scan queries to the SQL beginners' guide: https://docs.d.tarantool.io/en/doc/gh-3281-sql-seqscan/how-to/sql/sql_beginners_guide/#selecting-with-where
- Added `sql_seq_scan` session setting to the session settings reference: https://docs.d.tarantool.io/en/doc/gh-3281-sql-seqscan/reference/reference_lua/box_space/_session_settings/

### DIFF
--- a/doc/how-to/sql/sql_tutorial.rst
+++ b/doc/how-to/sql/sql_tutorial.rst
@@ -117,6 +117,50 @@ Then try to put another row:
 This ``INSERT`` fails because of a primary-key violation: the row with the primary
 key ``1, 'AB'`` already exists.
 
+The SEQSCAN keyword
+~~~~~~~~~~~~~~~~~~~
+
+In Tarantool, SQL queries that perform sequential scans (that is, go through all
+the table instead of using indexes) are prohibited by default. For example, this
+query leads to the error ``Scanning is not allowed for 'table2':
+
+.. code-block:: sql
+
+    SELECT * FROM table2
+
+To execute a scan query, put the ``SEQSCAN`` keyword before the table name:
+
+.. code-block:: sql
+
+    SELECT * FROM SEQSCAN table2
+
+If a query uses indexed fields in filters, it can still be a scan query.
+Try to execute these queries that seem to return the same result:
+
+.. code-block:: sql
+
+    SELECT * FROM table2 WHERE column1 = 1;
+    SELECT * FROM table2 WHERE column1 + 1 = 2
+
+The result is:
+
+*   the first query returns
+
+    .. code-block:: tarantoolsession
+
+        - [1, 'AB', 'AB', 10.5]
+        - [1, 'CD', '  ', 10005]
+
+*   the second query fails with the error ``Scanning is not allowed for 'TABLE2'``.
+
+.. note::
+
+    You can allow SQL scan queries without ``SEQSCAN`` for the current session
+    by running ``SET SESSION "sql_seq_scan" = true;``.
+
+
+Learn more about using ``SEQSCAN`` in the :ref:`SQL FROM clause description <sql_from>`.
+
 SELECT with ORDER BY clause
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
@@ -127,8 +171,11 @@ Retrieve the 4 rows in the table, in descending order by ``column2``, then
 
 .. code-block:: sql
 
-    SELECT * FROM table2 ORDER BY column2 DESC, column4 ASC;
+    SELECT * FROM SEQSCAN table2 ORDER BY column2 DESC, column4 ASC;
 
+.. important::
+
+    Tarantool has its own
 The result is:
 
 .. code-block:: tarantoolsession
@@ -152,7 +199,7 @@ Retrieve some of what you inserted:
 
 .. code-block:: sql
 
-    SELECT column1, column2, column1 * column4 FROM table2 WHERE column2
+    SELECT column1, column2, column1 * column4 FROM SEQSCAN table2 WHERE column2
     LIKE 'A%';
     SELECT column1, column2, column3, column4 FROM table2
         WHERE (column1 < 2 AND column4 < 10)
@@ -184,8 +231,8 @@ The rows which have the same values for ``column2`` are grouped and are aggregat
 
 .. code-block:: sql
 
-    SELECT column2, SUM(column4), COUNT(column4), AVG(column4)
-    FROM table2
+    SELECT SEQSCAN column2, SUM(column4), COUNT(column4), AVG(column4)
+    FROM SEQSCAN table2
     GROUP BY column2;
 
 The result is:
@@ -251,8 +298,8 @@ from the result table.
 
     CREATE TABLE table3 (column1 INTEGER, column2 VARCHAR(100), PRIMARY KEY
     (column2));
-    INSERT INTO table3 SELECT column1, column2 FROM table2 WHERE column1 <> 2;
-    SELECT * FROM table3;
+    INSERT INTO table3 SELECT column1, column2 FROM SEQSCAN table2 WHERE column1 <> 2;
+    SELECT * FROM SEQSCAN table3;
 
 The result is:
 
@@ -274,8 +321,8 @@ present in ``table3``.
 
 .. code-block:: sql
 
-    SELECT * FROM table2 WHERE (column1, column2) NOT IN (SELECT column1,
-    column2 FROM table3);
+    SELECT * FROM SEQSCAN table2 WHERE (column1, column2) NOT IN (SELECT column1,
+    column2 FROM SEQSCAN table3);
 
 The result is the single row that was excluded when inserting the rows with
 the ``INSERT ... SELECT`` statement:
@@ -295,7 +342,7 @@ column values from another table.
 
 .. code-block:: sql
 
-    SELECT * FROM table2, table3
+    SELECT * FROM SEQSCAN table2, table3
         WHERE table2.column1 = table3.column1 AND table2.column2 = table3.column2
         ORDER BY table2.column4;
 
@@ -327,7 +374,7 @@ containing ``13`` in ``column2``. Then try to insert such a row.
     INSERT INTO table4 VALUES (12, 13);
 
 Result: the insert fails, as it should, with the message
-``Check constraint failed ''ck_unnamed_TABLE4_1'': column2 <> 13``.
+``Check constraint 'ck_unnamed_TABLE4_1' failed for tuple``.
 
 CREATE TABLE with a FOREIGN KEY clause
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -348,7 +395,7 @@ Result:
 *   The first ``INSERT`` statement succeeds because
     ``table3`` contains a row with ``[2, 'AB', ' ', 12.34567]``.
 *   The second ``INSERT`` statement, correctly, fails with the message
-    ``Failed to execute SQL statement: FOREIGN KEY constraint failed``.
+    ``Foreign key constraint ''fk_unnamed_TABLE5_1'' failed: foreign tuple was not found``.
 
 UPDATE
 ~~~~~~
@@ -361,7 +408,7 @@ Use ``SELECT`` to see what happened to ``column4``.
 .. code-block:: sql
 
     UPDATE table2 SET column4 = column4 + 5 WHERE column4 <> 0;
-    SELECT column4 FROM table2 ORDER BY column4;
+    SELECT column4 FROM SEQSCAN table2 ORDER BY column4;
 
 The result is: ``{NULL, NULL, 0, 10.5, 17.34567, 10005}``.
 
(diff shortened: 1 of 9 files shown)


doc/how-to/sql/sql_tutorial.rst
```diff
@@ -117,6 +117,54 @@ Then try to put another row:
 This ``INSERT`` fails because of a primary-key violation: the row with the primary
 key ``1, 'AB'`` already exists.
 
+The SEQSCAN keyword
+~~~~~~~~~~~~~~~~~~~
+
+In Tarantool, SQL queries that perform sequential scans (that is, go through all
```
ImeevMA: I think it's better to mention that this only applies to `SELECT` queries and subqueries.

doc/how-to/sql/sql_tutorial.rst
```diff
@@ -117,6 +117,54 @@ Then try to put another row:
...
+
+    SELECT * FROM table2
+
+To execute a scan query, put the ``SEQSCAN`` keyword before the table name:
+
+.. code-block:: sql
+
+    SELECT * FROM SEQSCAN table2
+
+If a query uses indexed fields in filters, it can still be a scan query.
```
ImeevMA: In our terminology, `scan queries` are queries that do not use indexes. This means that if a `SELECT` query uses indexes, then that query is not a `scan query`, so it can be executed without `SEQSCAN` keyword.

doc/how-to/sql/sql_tutorial.rst
```diff
@@ -184,8 +235,8 @@ The rows which have the same values for ``column2`` are grouped and are aggregat
 
 .. code-block:: sql
 
-    SELECT column2, SUM(column4), COUNT(column4), AVG(column4)
-    FROM table2
+    SELECT SEQSCAN column2, SUM(column4), COUNT(column4), AVG(column4)
```
ImeevMA: I think this `SEQSCAN` is unnecessary. We use `SEQSCAN` for tables, not for columns.

doc/how-to/sql/sql_tutorial.rst
```diff
@@ -117,6 +117,54 @@ Then try to put another row:
...
+The SEQSCAN keyword
+~~~~~~~~~~~~~~~~~~~
+
+In Tarantool, SQL queries that perform sequential scans (that is, go through all
+the table instead of using indexes) are prohibited by default. For example, this
+query leads to the error ``Scanning is not allowed for 'table2'``:
+
+.. code-block:: sql
+
+    SELECT * FROM table2
```
ImeevMA: I think `;` should be added since we set `;` as delimiter.

doc/how-to/sql/sql_tutorial.rst
```diff
@@ -117,6 +117,54 @@ Then try to put another row:
...
+
+To execute a scan query, put the ``SEQSCAN`` keyword before the table name:
+
+.. code-block:: sql
+
+    SELECT * FROM SEQSCAN table2
```
ImeevMA: I think `;` is missing.

doc/how-to/sql/sql_tutorial.rst
```diff
@@ -117,6 +117,54 @@ Then try to put another row:
...
+The result is:
+
+*   The first query returns rows:
+
+    .. code-block:: tarantoolsession
+
+        - [1, 'AB', 'AB', 10.5]
+        - [1, 'CD', '  ', 10005]
+
+*   The second query fails with the error ``Scanning is not allowed for 'TABLE2'``.
```
ImeevMA, line 153: It might be better to mention that the second query is a scan query, since such an expression cannot be executed using index.

It might be worth mentioning somewhere which expressions can be executed using indexes. However, this can be quite a large section since we also need to describe how multi-field indexes work..

doc/how-to/sql/sql_tutorial.rst
```diff
@@ -152,7 +203,7 @@ Retrieve some of what you inserted:
 
 .. code-block:: sql
 
-    SELECT column1, column2, column1 * column4 FROM table2 WHERE column2
+    SELECT column1, column2, column1 * column4 FROM SEQSCAN table2 WHERE column2
     LIKE 'A%';
     SELECT column1, column2, column3, column4 FROM table2
```
ImeevMA: The `SEQSCAN` keyword is missing.

doc/how-to/sql/sql_tutorial.rst
```diff
@@ -516,7 +567,7 @@ with arithmetic on a number column and ordering by a string column.
     INSERT INTO t6 VALUES (+1234567890, 'GD', 1e30);
     INSERT INTO t6 VALUES (10, 'FADEW?', 0.000001);
     INSERT INTO t6 VALUES (5, 'ABCDEFG', NULL);
-    SELECT column1 + 1, column2, column4 * 2 FROM t6 ORDER BY column2;
+    SELECT column1 + 1, column2, column4 * 2 FROM SEQSCAN t6 ORDER BY column2;
```
ImeevMA, line 571: In `CREATE VIEW v3 AS SELECT SUBSTR(column2,1,2), column4 FROM t6 WHERE column4 >= 0;` we should also add `SEQSCAN` keyword. Otherwise `SELECT * FROM v3;` fails.

Same for `WITH cte AS (SELECT SUBSTR(column2,1,2), column4 FROM t6 WHERE column4 >= 0) SELECT * FROM cte;`.

Same for `SELECT * FROM "_space";`, `SELECT * FROM "_index` and `SELECT * FROM "_trigger";`.

doc/reference/reference_sql/sql_statements_and_clauses.rst
```diff
@@ -1889,18 +1889,35 @@ Parentheses are allowed, and ``[[AS] correlation-name]`` is allowed.
...
 
+The ``SEQSCAN`` keyword (since :doc:`2.11 </release/2.11.0>`) marks the queries that
+perform sequential scans during the execution. It happens if the query can't use indexes,
+and goes through all the table rows one by one, sometimes causing a heavy load.
+Such queries are called *scan queries*. If a scan query doesn't have the
+``SEQSCAN`` keyword, Tarantool raises an error. ``SEQSCAN`` must precede all
+names of the tables that the query scans.
+
+To find out if a query performs a sequential scan, use ``EXPLAIN QUERY PLAN``.
+For scan queries, the result contains ``SCAN_TABLE``.
```
ImeevMA: I think it is better to write `SCAN TABLE ...` or `SCAN TABLE table_name` instead of `SCAN_TABLE`. For example:
```
tarantool> box.execute([[EXPLAIN QUERY PLAN SELECT * FROM "_space", "_index" WHERE "_space"."id" == 512;]])
---
- metadata:
  - name: selectid
    type: integer
  - name: order
    type: integer
  - name: from
    type: integer
  - name: detail
    type: text
  rows:
  - [0, 0, 0, 'SEARCH TABLE _space USING PRIMARY KEY (id=?) (~1 row)']
  - [0, 1, 1, 'SCAN TABLE _index (~1048576 rows)']
...
```

ImeevMA (commented):
Hi! Thank you for the patch. I have some comments, please take a look at them.

doc/how-to/sql/sql_tutorial.rst
p7nov: Thanks, fixed these examples.

doc/how-to/sql/sql_tutorial.rst
p7nov: Rewrote the example description.

doc/reference/reference_sql/sql_statements_and_clauses.rst
```diff
@@ -1889,18 +1889,35 @@ Parentheses are allowed, and ``[[AS] correlation-name]`` is allowed.
...
 
+The ``SEQSCAN`` keyword (since :doc:`2.11 </release/2.11.0>`) marks the queries that
+perform sequential scans during the execution. It happens if the query can't use indexes,
+and goes through all the table rows one by one, sometimes causing a heavy load.
+Such queries are called *scan queries*. If a scan query doesn't have the
+``SEQSCAN`` keyword, Tarantool raises an error. ``SEQSCAN`` must precede all
+names of the tables that the query scans.
+
+To find out if a query performs a sequential scan, use ``EXPLAIN QUERY PLAN``.
+For scan queries, the result contains ``SCAN_TABLE tabe_name``.
```
ImeevMA: It is actually `SCAN TABLE table_name` instead of `SCAN_TABLE tabe_name`. See example in the comment above.

ImeevMA (approved):
Thank you for the fixes! I have only one comment.

doc/reference/reference_sql/sql_statements_and_clauses.rst
```diff
@@ -1889,18 +1889,35 @@ Parentheses are allowed, and ``[[AS] correlation-name]`` is allowed.
...
+names of the tables that the query scans.
+
+To find out if a query performs a sequential scan, use ``EXPLAIN QUERY PLAN``.
+For scan queries, the result contains ``SCAN TABLE table_name``.
+
+..  note::
+
+    For backward compatibility, the scan queries without the ``SEQSCAN`` keyword
+    are allowed in Tarantool 2.11. The errors on scan queries will be the default
+    behavior starting from 3.0. You can change the default behavior of scan queries
```
andreyaksenov, line 1906: > will be the default behavior starting from 3.0

I'd rephrase this to the present tense - 3.0 (albeit an alpha version) already published and the behavior is already changed. Looks like we can eliminate a future promise here. Additionally, we won't update this note to the present tense when the 3.0 release version is actually published.

andreyaksenov (approved):

doc/reference/reference_sql/sql_statements_and_clauses.rst
p7nov: Thanks, fixed.

PR #3648, opened 2023-08-29
Foreign keys constraints

### LINKED ISSUE #3243: SQL: document changes in parsing rules
**Product:** Tarantool
**Since:** 2.11
**Root document:** https://www.tarantool.io/en/doc/latest/reference/reference_sql/sql_statements_and_clauses/#table-constraint-definition-for-foreign-keys
**SME:** @ ImeevMA

# Details

Three rules related to foreign keys have been removed.

The first is the "reference trigger action" rule, which determines the
behavior when a referenced tuple is deleted. This rule has been dropped
because the new foreign keys only support RESTRICT.

The second is "constraint check time" rule, which determines when a
foreign key constraint should be validated. This rule has been dropped
because the new foreign keys only support INITIALLY IMMEDIATELY.

The third is "match type" rule, which determines how inserted values are
validated. This rule has been dropped because the new foreign keys only
support FULL.
Requested by @ImeevMA in https://github.com/tarantool/tarantool/commit/5ab45ba33ea54b19e2fe668743c6417e03eba4ba.

1) RESTRICT
2) INITIALLY IMMEDIATE
3) MATCH FULL

andreyaksenov:
Updated the section about foreign key constraints: https://docs.d.tarantool.io/en/doc/sql-table-constraint/reference/reference_sql/sql_statements_and_clauses/#sql-foreign-key

Changes:
- Updated the syntax used for foreign key constraints - removed deprecated referencing options.
- Added a syntax when a constraint is created in a column definition.
- Added a note with deprecated referencing options.
- Added a testable code sample with constraint error examples.
- Tried to make the entire section a bit clearer.

### DIFF
--- a/doc/code_snippets/test/sql/foreign_key_table_constraint_test.lua
+++ b/doc/code_snippets/test/sql/foreign_key_table_constraint_test.lua
@@ -1,0 +1,82 @@
+local fio = require('fio')
+local server = require('luatest.server')
+local t = require('luatest')
+local g = t.group()
+g.before_each(function(cg)
+    cg.server = server:new {
+        box_cfg = {},
+        workdir = fio.cwd() .. '/tmp'
+    }
+    cg.server:start()
+end)
+
+g.after_each(function(cg)
+    cg.server:stop()
+    cg.server:drop()
+end)
+
+g.test_space_is_updated = function(cg)
+    cg.server:exec(function()
+        box.execute([[SET SESSION "sql_seq_scan" = true;]])
+        box.execute([[
+            -- create_parent_start
+            CREATE TABLE author (
+                id INTEGER PRIMARY KEY,
+                name STRING NOT NULL
+            );
+            -- create_parent_end
+        ]])
+        box.execute([[
+            -- insert_parent_start
+            INSERT INTO author VALUES (1, 'Leo Tolstoy'),
+                                      (2, 'Fyodor Dostoevsky');
+            -- insert_parent_end
+        ]])
+        box.execute([[
+            -- create_child_start
+            CREATE TABLE book (
+                id INTEGER PRIMARY KEY,
+                title STRING NOT NULL,
+                author_id INTEGER NOT NULL UNIQUE,
+                FOREIGN KEY (author_id)
+                    REFERENCES author (id)
+            );
+            -- create_child_end
+        ]])
+        box.execute([[
+            -- insert_child_start
+            INSERT INTO book VALUES (1, 'War and Peace', 1),
+                                    (2, 'Crime and Punishment', 2);
+            -- insert_child_end
+        ]])
+        local _, insert_err = box.execute([[
+            -- insert_error_start
+            INSERT INTO book VALUES (3, 'Eugene Onegin', 3);
+            /*
+            - 'Foreign key constraint ''fk_unnamed_BOOK_1'' failed: foreign tuple was not found'
+            */
+            -- insert_error_end
+        ]])
+        local _, update_err = box.execute([[
+            -- update_error_start
+            UPDATE book SET author_id = 10 WHERE id = 1;
+            /*
+            - 'Foreign key constraint ''fk_unnamed_BOOK_1'' failed: foreign tuple was not found'
+            */
+            -- update_error_end
+        ]])
+        local _, delete_err = box.execute([[
+            -- delete_error_start
+            DELETE FROM author WHERE id = 2;
+            /*
+            - 'Foreign key ''fk_unnamed_BOOK_1'' integrity check failed: tuple is referenced'
+            */
+            -- delete_error_end
+        ]])
+
+        -- Tests
+        t.assert_equals(insert_err:unpack().message, "Foreign key constraint 'fk_unnamed_BOOK_1' failed: foreign tuple was not found")
+        t.assert_equals(update_err:unpack().message, "Foreign key constraint 'fk_unnamed_BOOK_1' failed: foreign tuple was not found")
+        t.assert_equals(delete_err:unpack().message, "Foreign key 'fk_unnamed_BOOK_1' integrity check failed: tuple is referenced")
+    end)
+end
--- a/doc/reference/reference_sql/sql_statements_and_clauses.rst
+++ b/doc/reference/reference_sql/sql_statements_and_clauses.rst
@@ -404,18 +404,18 @@ The column-constraint or default clause may be as follows:
    * - NOT NULL
      - means "it is illegal to assign a NULL to this column"
    * -  PRIMARY KEY
-     - explained in the later section
-       :ref:`"Table Constraint Definition" <sql_table_constraint_def>`
+     - explained in the
+       :ref:`Table constraint definition <sql_table_constraint_def>` section
    * - UNIQUE
-     - explained in the later section
-       "Table Constraint Definition"
+     - explained in the
+       :ref:`Table constraint definition <sql_table_constraint_def>` section
    * - CHECK (expression)
-     - explained in the later section
-       "Table Constraint Definition"
+     - explained in the
+       :ref:`Table constraint definition <sql_table_constraint_def>` section
    * - foreign-key-clause
-     - explained in the later section
-       :ref:`"Table Constraint Definition for foreign keys"
-       <sql_foreign_key>`
+     - explained in the
+       :ref:`Table constraint definition for foreign keys
+       <sql_foreign_key>` section
    * - DEFAULT expression
      - means
        "if INSERT does not assign to this column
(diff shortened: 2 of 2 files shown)


doc/code_snippets/test/sql/foreign_key_table_constraint_test.lua
```diff
@@ -0,0 +1,82 @@
...
+end)
+
+g.test_space_is_updated = function(cg)
+    cg.server:exec(function()
+        box.execute([[SET SESSION "sql_seq_scan" = true;]])
+        box.execute([[
+            -- create_parent_start
+            CREATE TABLE author (
+                id INTEGER PRIMARY KEY,
+                name VARCHAR(100) NOT NULL
```
ImeevMA: I suggest using `STRING` instead of `VARCHAR(...)` here and below, since our SQL does not support fixed length string values and will automatically replace them with `STRING`.

doc/reference/reference_sql/sql_statements_and_clauses.rst
```diff
@@ -620,140 +620,95 @@ causes no error message, although (s1, s1) is probably a user error.
...
+    *   ``DEFERRABLE`` constraints. The ``INITIALLY IMMEDIATE`` constraint check time rule is used implicitly.
 
-.. COMMENT
-   Constraint Conflict Clauses are temporarily disabled.
-   However, the description is here, as a big comment.
 
-   Constraint Conflict Clauses
+The referenced column names must be defined in the referenced table and have similar data types.
+The values in the referencing columns must equal values in the referenced columns of the referenced table,
+or at least one of the referencing columns must contain ``NULL``.
```
ImeevMA: This is currently not the case. This describes `MATCH SIMPLE`, which is usually the default. However, we only support `MATCH FULL`, which requests a full match.

doc/reference/reference_sql/sql_statements_and_clauses.rst
```diff
@@ -620,140 +620,95 @@ causes no error message, although (s1, s1) is probably a user error.
...
-   If behavior is REPLACE: the second statement is not rolled back, the first (2) is replaced by the second (2), there is no error message. The table now contains (1), (2), (3), (5).
+    ..  literalinclude:: /code_snippets/test/sql/foreign_key_table_constraint_test.lua
+        :language: sql
+        :start-after: insert_error_start
+        :end-before: insert_error_end
+        :dedent:
 
-   If behavior is ROLLBACK: the statement is rolled back, and the first statement is rolled back,
-   and there is an error message. The table now contains nothing.
+    On an attempt to delete an author that already has books in the ``book`` table, the following error is raised:
```
ImeevMA, line 699: The error shown is actually not what I expected from the description. This is a new limitation for FKs: referencing fields must now have an index in order to remove anything from the referenced space. Otherwise, we will not be able to remove from the referenced space, even if the row is not referenced. For example, if we add `(3, 'Alexander Pushkin')` to the `author` space and do not add anything to the `book` space, we won't be able to remove the added row. To remove it, we need to create an index on `book.author_id`.

To receive expected error I suggest to replace
```
        box.execute([[
            -- create_child_start
            CREATE TABLE book (
                id INTEGER PRIMARY KEY,
                title VARCHAR(200) NOT NULL,
                author_id INTEGER NOT NULL,
                FOREIGN KEY (author_id)
                    REFERENCES author (id)
            );
            -- create_child_end
        ]])
```
by
```
        box.execute([[
            -- create_child_start
            CREATE TABLE book (
                id INTEGER PRIMARY KEY,
                title VARCHAR(200) NOT NULL,
                author_id INTEGER NOT NULL UNIQUE,
                FOREIGN KEY (author_id)
                    REFERENCES author (id)
            );
            -- create_child_end
        ]])
```
and mention why we add UNIQUE constraint here.

Then the error will be:
```'Foreign key ''fk_unnamed_BOOK_1'' integrity check failed: tuple is referenced'```

File: doc/reference/reference_sql/sql_statements_and_clauses.rst [hunk 4]
```diff
@@ -620,140 +620,95 @@ causes no error message, although (s1, s1) is probably a user error.
 Table Constraint Definition for foreign keys
 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
-FOREIGN KEY constraints look like this: |br|
-:samp:`FOREIGN KEY ({referencing-column-name} [, {referencing-column-name}...]) REFERENCES {referenced-table-name} [({referenced-column-name} [, {referenced-column-name}...]]) [MATCH FULL] [update-or-delete-rules]`
+A foreign key is a constraint that can be used to enforce data integrity across related tables.
+A foreign key constraint is defined on the child table that references the parent table's column values.
 
-There is a shorthand: specifying REFERENCES in a :ref:`column definition <sql_column_def_constraint>`.
+Foreign key constraints look like this:
```
ImeevMA, line 626: I see that the described FK constraint is known as the FK tuple constraint. What about the FK field constraint? I'm not sure if the FK description should be duplicated due to different types of constraints, but they have different error descriptions and are described differently in '_space', so at least add a quick overview of that and links to 'tuple constraint' and ' field constraint ' looks like a good idea.

ImeevMA (commented):
Thank you for the patch! I have a few comments, please look at them.

doc/reference/reference_sql/sql_statements_and_clauses.rst
andreyaksenov: Looks like this requires some investigation, created a separate ticket: https://github.com/tarantool/doc/issues/3703

doc/reference/reference_sql/sql_statements_and_clauses.rst
```diff
@@ -620,140 +620,90 @@ causes no error message, although (s1, s1) is probably a user error.
 Table Constraint Definition for foreign keys
 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
-FOREIGN KEY constraints look like this: |br|
-:samp:`FOREIGN KEY ({referencing-column-name} [, {referencing-column-name}...]) REFERENCES {referenced-table-name} [({referenced-column-name} [, {referenced-column-name}...]]) [MATCH FULL] [update-or-delete-rules]`
+A foreign key is a constraint that can be used to enforce data integrity across related tables.
+A foreign key constraint is defined on the child table that references the parent table's column values.
```
p7nov, line 624: Maybe link to box foreign keys documentation? https://www.tarantool.io/en/doc/latest/concepts/data_model/value_store/#foreign-keys

doc/reference/reference_sql/sql_statements_and_clauses.rst
```diff
@@ -620,140 +620,90 @@ causes no error message, although (s1, s1) is probably a user error.
...
+        REFERENCES {referenced_table_name}({referenced_column_name})
 
-Limitations: |br|
-* Foreign keys can have a MATCH clause (`Issue#3455 <https://github.com/tarantool/tarantool/issues/3455>`_).
+.. NOTE::
 
-.. COMMENT
-   Constraint Conflict Clauses are temporarily disabled.
-   However, the description is here, as a big comment.
+    With :doc:`2.11.0 </release/2.11.0>`, the following referencing options aren't supported anymore:
```
p7nov: Since 2.11.0?

doc/reference/reference_sql/sql_statements_and_clauses.rst
```diff
@@ -620,140 +620,90 @@ causes no error message, although (s1, s1) is probably a user error.
...
+-   A referenced column has a UNIQUE index.
 
-   The standard way to handle a constraint violation is "statement rollback" -- all rows affected by the statement are restored to their original values -- and an error is returned. However, Tarantool allows the user to specify non-standard ways to handle PRIMARY KEY, UNIQUE, CHECK, and NOT NULL constraint violations.
+**Example**
 
-   ABORT -- do statement rollback and return an error. This is the default and is recommended, so a user's best strategy is to never use constraint conflict clauses.
+This example shows how to create a relation between the parent and child tables through a single-column foreign key:
 
-   FAIL -- return an error but do not do statement rollback.
+1.  First, create a parent ``author`` table:
```
p7nov, line 658: Nit: "first", "then", and similar words are usually redundant in procedures with numbered steps.
Let's keep it shorter

doc/reference/reference_sql/sql_statements_and_clauses.rst
```diff
@@ -620,140 +620,90 @@ causes no error message, although (s1, s1) is probably a user error.
 Table Constraint Definition for foreign keys
```
p7nov: Maybe fix capitalization in this and previous heading while we are here?
image

p7nov (approved):
LGTM 👍

PR #3677, opened 2023-09-04
Update Python connectors page

DifferentialOrange:
Update Python connectors comparison table, remove outdated info. See commits for more info.


p7nov:
Thanks for this update!

Issue #947, opened 2019-10-14
 The auto-increment feature for multipart PK


TarantoolBot:
The auto-increment feature can be set to any INTEGER or UNSIGNED
field of PRIMARY KEY using one of two ways:
1) AUTOINCREMENT in column definition:
CREATE TABLE t (i INT, a INT AUTOINCREMENT, PRIMARY KEY (i, a));
CREATE TABLE t (i INT AUTOINCREMENT, a INT, PRIMARY KEY (i, a));
2) AUTOINCREMENT in PRIMARY KEY definition:
CREATE TABLE t (i INT, a INT, PRIMARY KEY (i, a AUTOINCREMENT));
CREATE TABLE t (i INT, a INT, PRIMARY KEY (i AUTOINCREMENT, a));
Requested by @ImeevMA in https://github.com/tarantool/tarantool/commit/08c7d7c13051c04d3227bc76e4e9244a7922711c.

pgulutzan:
SQL constraint definitions will be published when approved.

veod32:
Do we have the description of this feature in our 2.3 documentation? 08c7d7c commit that closes the feature ticket gh-4217 is in the 2.3.1 release scope already.

pgulutzan:
That part of the documentation has not been approved. Perhaps K. Yukhin knows whether or not it will be approved.

Totktonada:
@pgulutzan, can you share the commit? I guess I can review it from correctness point of view and it is enough for now.

I'll ask help from @artembo to push it in the right way (same as for #879).

PR #3707, opened 2023-09-12
SQL constraint changes

### LINKED ISSUE #3246: SQL: changes in using the constraint mechanism
**Product:** Tarantool
**Since:** 2.11
**Root document:** https://www.tarantool.io/en/doc/latest/reference/reference_sql/sql_statements_and_clauses/#table-constraint-definition
https://www.tarantool.io/en/doc/latest/concepts/data_model/value_store/#constraints
**SME:** @ ImeevMA

# Details

SQL constraints will now use the tuple constraint mechanism instead of
the original SQL mechanism. This allows to set constraints that are
independent of the language used for creation. However, this disables
some SQL constraint features that are not supported by tuple
constraints.
Requested by @ImeevMA in https://github.com/tarantool/tarantool/commit/930674810863513ca4651a016bef02cbb448add6.

### LINKED ISSUE #3703: feedback: SQL statements and clauses | Tarantool
> **|FOREIGN KEY constraints look like this:|** 
> FOREIGN KEY (referencing-column-name [, referencing-column-nam<…>

From Mergen:

https://www.tarantool.io/en/doc/latest/reference/reference_sql/sql_statements_and_clauses/#table-constraint-definition-for-foreign-keys

I see that the described FK constraint is known as the FK tuple constraint. What about the FK field constraint? I'm not sure if the FK description should be duplicated due to different types of constraints, but they have different error descriptions and are described differently in '_space', so at least add a quick overview of that and links to 'tuple constraint' and ' field constraint ' looks like a good idea.

andreyaksenov:
- Updated a table constraint syntax to make it more readable and added new testable examples for all constraint types:
   https://docs.d.tarantool.io/en/doc/sql-constraint-mechanism/reference/reference_sql/sql_statements_and_clauses/#table-constraint-definition
- Added a note about a foreign key's uniqueness checking:
   https://docs.d.tarantool.io/en/doc/sql-constraint-mechanism/reference/reference_sql/sql_statements_and_clauses/#table-constraint-definition-for-foreign-keys

### DIFF
--- a/doc/reference/reference_sql/sql_statements_and_clauses.rst
+++ b/doc/reference/reference_sql/sql_statements_and_clauses.rst
@@ -241,7 +241,7 @@ Syntax:
 
 :samp:`column-name data-type [, column-constraint]`
 
-Define a column, which is a table-element used in a :ref:`CREATE TABLE <sql_create_table>` statement.
+Define a column, which is a table element used in a :ref:`CREATE TABLE <sql_create_table>` statement.
 
 The ``column-name`` must be an identifier which is valid according to the rules
 for identifiers.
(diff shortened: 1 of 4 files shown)


doc/reference/reference_sql/sql_statements_and_clauses.rst
```diff
@@ -496,54 +496,101 @@ Data types may also appear in :ref:`CAST <sql_function_cast>` functions.
...
 
-|br|
 
-Define a constraint, which is a table-element used in a CREATE TABLE statement.
+Define a constraint, which is a table element used in a CREATE TABLE statement.
 
-The constraint-name must be an identifier which is valid according to the rules for identifiers.
-The constraint-name must be unique within the table.
+A constraint name must be an identifier that is valid according to the rules for identifiers.
+A constraint name must be unique within the table.
```
ImeevMA: This is not exactly so. There cannot be another table foreign key (= tuple foreign key) constraint with the same name, but it is possible to have a CHECK or UNIQUE table constraint, as well as any column constraint (= field constraint) with the same name.

doc/reference/reference_sql/sql_statements_and_clauses.rst
```diff
@@ -496,54 +496,101 @@ Data types may also appear in :ref:`CAST <sql_function_cast>` functions.
...
+
+-   In a column definition after the words PRIMARY KEY, as in ``CREATE TABLE t (c INTEGER PRIMARY KEY AUTOINCREMENT);``
+
+-   In a PRIMARY KEY (column-list) after a column name, as in ``CREATE TABLE t (c INTEGER, PRIMARY KEY (c AUTOINCREMENT));``
+
+When AUTOINCREMENT is specified, the column must be a primary-key column and it must be INTEGER or UNSIGNED.
+
 Only one column in the table may be autoincrement.
 However, it is legal to say ``PRIMARY KEY (a, b, c AUTOINCREMENT)`` -- in that case, there
 are three columns in the primary key but only the first column (``a``) is AUTOINCREMENT.
```
ImeevMA: I saw this by chance. This is not entirely true. I see that this was added to close https://github.com/tarantool/doc/issues/947, but in fact `AUTOINCREMENT` will be added here to column `c`.

doc/reference/reference_sql/sql_statements_and_clauses.rst
```diff
@@ -651,6 +749,10 @@ Note that a referenced column should meet one of the following requirements:
 -   A referenced column has a UNIQUE constraint.
 -   A referenced column has a UNIQUE index.
 
+Note that before the :doc:`2.11.0 </release/2.11.0>` version, uniqueness of a foreign key is checked when creating a constraint (for example, using ``CREATE TABLE`` or ``ALTER TABLE``).
```
ImeevMA: I think it is more `existence of an index on the referenced columns` then `uniqueness of a foreign key`.

ImeevMA (commented):
Hi! Thank you for the patch. I have three comments, see them below.

doc/reference/reference_sql/sql_statements_and_clauses.rst
andreyaksenov: Thanks, changed to `... but only the third column (``c``) is AUTOINCREMENT.`

doc/reference/reference_sql/sql_statements_and_clauses.rst
andreyaksenov: Changed like this:
```
A constraint name must be unique within the table for a specific constraint type.
For example, the CHECK and FOREIGN KEY constraints can have the same name.
```

doc/reference/reference_sql/sql_statements_and_clauses.rst
andreyaksenov: Updated to:
```
Note that before the :doc:`2.11.0 </release/2.11.0>` version, an index existence for the referenced columns is checked when creating a constraint (for example, using ``CREATE TABLE`` or ``ALTER TABLE``).
Starting with 2.11.0, this check is weakened and existence of an index is checked during data insertion.
```

doc/reference/reference_sql/sql_statements_and_clauses.rst
```diff
@@ -496,95 +496,185 @@ Data types may also appear in :ref:`CAST <sql_function_cast>` functions.
...
 
-Every table must have one and only one primary key. |br|
-Primary-key columns are automatically NOT NULL. |br|
-Primary-key columns are automatically indexed. |br|
-Primary-key columns are unique, that is, it is illegal to have two rows which
-have the same values for the columns specified in the constraint.
+-   Every table must have one and only one primary key.
+-   Primary-key columns are automatically ``NOT NULL``.
+-   Primary-key columns are automatically indexed.
+-   Primary-key columns are unique, that is, it is illegal to have two rows that have the same values for the columns specified in the constraint.
```
xuniq:
```suggestion
-   Primary-key columns are unique. That means it is illegal to have two rows with the same values for the columns specified in the constraint.
```

doc/reference/reference_sql/sql_statements_and_clauses.rst
```diff
@@ -496,95 +496,185 @@ Data types may also appear in :ref:`CAST <sql_function_cast>` functions.
...
-:samp:`UNIQUE ({column-name} [, {column-name}...])`
+UNIQUE constraints look like this:
+
+.. code-block:: sql
+
+    UNIQUE (column_name, ...)
 
 There is a shorthand: specifying UNIQUE in a :ref:`column definition <sql_column_def_constraint>`.
 
 Unique constraints are similar to primary-key constraints, except that:
```
xuniq, line 616:
```suggestion
*   Unique constraints are similar to primary-key constraints, except that:
    a table may have any number of unique keys, and unique keys are not automatically ``NOT NULL``.
*    Unique columns are automatically indexed.
*   Unique columns are unique. That means it is illegal to have two rows with the same values in the unique-key columns.
```
Let's make the formatting consistent

doc/reference/reference_sql/sql_statements_and_clauses.rst
```diff
@@ -651,6 +750,10 @@ Note that a referenced column should meet one of the following requirements:
 -   A referenced column has a UNIQUE constraint.
 -   A referenced column has a UNIQUE index.
 
+Note that before the :doc:`2.11.0 </release/2.11.0>` version, an index existence for the referenced columns is checked when creating a constraint (for example, using ``CREATE TABLE`` or ``ALTER TABLE``).
+Starting with 2.11.0, this check is weakened and existence of an index is checked during data insertion.
```
xuniq:
```suggestion
Starting with 2.11.0, this check is weakened and the existence of an index is checked during data insertion.
```

doc/reference/reference_sql/sql_statements_and_clauses.rst
```diff
@@ -496,95 +496,185 @@ Data types may also appear in :ref:`CAST <sql_function_cast>` functions.
...
+are three columns in the primary key but only the third column (``c``) is AUTOINCREMENT.
 
 As the name suggests, values in an autoincrement column are automatically incremented.
 That is: if a user inserts NULL in the column, then the stored value will be the smallest
 non-negative integer that has not already been used.
 This occurs because autoincrement columns are associated with :doc:`sequences </reference/reference_lua/box_schema_sequence/create_index>`.
 
-UNIQUE constraints look like this: |br|
-:samp:`UNIQUE ({column-name} [, {column-name}...])`
+UNIQUE constraints look like this:
```
xuniq, line 608:
```suggestion
**UNIQUE constraints**

UNIQUE constraints look like this:
```
I think it's necessary to separate the paragraphs about PRIMARY | UNIQUE | CHECK constraints from each other somehow (with **bold**?). It's a bit difficult to figure out where a new paragraph starts.

andreyaksenov: Agree, will do

xuniq (approved):

PR #3753, opened 2023-10-04
Update Vale rules

andreyaksenov:
- Added a suggestion for swapping `flag` with `option` in a case of CLI options.
- Made some rules a bit weaker.
- Added missing terms.

### DIFF
--- a/styles/Tarantool/Uppercase.yml
+++ b/styles/Tarantool/Uppercase.yml
@@ -251,3 +251,6 @@ exceptions:
   - YAML
   - ZAP
   - ZIP
+  - PAP
+  - CHAP
+  - APT
--- a/styles/Vocab/Tarantool/accept.txt
+++ b/styles/Vocab/Tarantool/accept.txt
@@ -34,3 +34,15 @@ dereference
 upsert
 tt
 SEQSCAN
+iproto
+autoincrement
+etcd
+canonicalization
+libcurl
+CMake
+fuzzer
+subcommand
+vshard
+boolean
+failover
+rebalancer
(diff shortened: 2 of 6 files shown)


styles/Vocab/Tarantool/accept.txt
```diff
@@ -33,4 +33,15 @@ storages
...
+autoincrement
+etcd
+xlog
+canonicalization
+libcurl
+CMake
+fuzzer
+subcommand
+vshard
+boolean
```
p7nov:
```suggestion
boolean
failover
```

styles/Vocab/Tarantool/accept.txt
```diff
@@ -33,4 +33,15 @@ storages
 dereference
 upsert
 tt
-SEQSCAN
\ No newline at end of file
+SEQSCAN
+iproto
+autoincrement
+etcd
+xlog
```
p7nov: Not sure we should encourage using "xlog" as a word in plain text. The correct entity name is "write-ahead log".

styles/Tarantool/Uppercase.yml
```diff
@@ -251,3 +251,5 @@ exceptions:
   - YAML
   - ZAP
   - ZIP
+  - PAP
+  - CHAP
```
p7nov, line 255:
```suggestion
  - CHAP
  - APT
```

p7nov (approved):
Thanks for this update!
Added a couple more suggestions.

xuniq (approved):

Issue #3664, opened 2023-09-01
[Config] Configuration reference (YAML)


andreyaksenov:
**Product:** Tarantool
**Since:** 3.0
**Root document:** https://www.tarantool.io/en/doc/latest/reference/configuration/configuration_reference/
**SME:** @ Totktonada

**Note:** Looks like adding descriptions for specific options makes sense when describing corresponding functionality. So, this issue should be used to verify that all necessary options are added to `Configuration reference` at some moment.

# Details

Create a configuration reference for a new approach. Instance config schema: https://github.com/tarantool/tarantool/blob/master/src/box/lua/config/instance_config.lua
Add information about applicable configuration scopes for each option.

See the comparison table: https://docs.google.com/spreadsheets/d/1_yEK9N_hBVWIga9mOk7zZzA6Zd3mAq7e5KZsnrbgj30/edit#gid=0

Old approach reference are kept here: https://www.tarantool.io/en/doc/latest/reference/configuration/.

## instance/replicaset/group names validation

Note from @ Totktonada: https://github.com/tarantool/doc/pull/3825#discussion_r1382292032

> We should also mention validation rules for instance/replicaset/group names somewhere. See https://github.com/tarantool/tarantool/blob/10372d0d9ded18ff64c256e125ce740bb0605587/src/box/lua/config/cluster_config.lua#L120-L141.

# PRs

Here are PRs that already added some configuration properties related to corresponding features:

- https://github.com/tarantool/doc/pull/3825
- https://github.com/tarantool/doc/pull/3834

Issue #3661, opened 2023-09-01
[Config] Migrating configuration from 2.11 to 3.х


andreyaksenov:
**Product:** Tarantool
**Since:** 3.0
**Root document:** [Releases](https://www.tarantool.io/en/doc/latest/release/) | `3.0`
**SME:** @ Totktonada

# Details

Create a guide that shows how to migrate the Tarantool configuration from the old approach to a new config. This guide may include two parts:
- Basic migration scenarios: https://docs.google.com/document/d/1bH22TuMDuJEp50vwsgGUCBxt3jt22ve6cKgAKDuAumA/edit?usp=sharing
- Full reference for all options: https://docs.google.com/spreadsheets/d/1_yEK9N_hBVWIga9mOk7zZzA6Zd3mAq7e5KZsnrbgj30/edit?usp=sharing

Totktonada:
(Irrelevant now.)

~Tarantool's new startup flow unconditionally passes instance name and replicaset name to the underneath `box.cfg()` call. It means that upgrading from snapshots without saved instance and replicaset names requires a one-time startup in the force-recovery mode.~

~We've two tasks planned in these regards:~

* https://github.com/tarantool/tarantool/issues/8978 (teamS)
* https://github.com/tarantool/tarantool/issues/8876 (teamL)

andreyaksenov:
See also: https://github.com/tarantool/doc/issues/3219

andreyaksenov:
See a note about UUIDs: https://github.com/tarantool/doc/pull/3834#discussion_r1397980470

Issue #3658, opened 2023-09-01
[Config] Storing configuration in 'etcd'


andreyaksenov:
**Product:** Tarantool
**Since:** 3.0
**Root document:** https://docs.d.tarantool.io/en/doc/3.0-config/concepts/configuration/configuration_etcd/

**SME:** @ Totktonada

# Details

Create a guide that shows how to store Tarantool config in etcd.
See the `Хранение конфигурации в etcd` section here: https://www.notion.so/tarantool/Getting-Started-with-3-0-03d461efc01c43dd9e09f2bc4903717b

Notes from @ Totktonada:
- The idea of a centralized configuration available in Tarantool EE (https://github.com/tarantool/tarantool-ee/issues/476).
- Etcd source: content of the etcd keys (yaml), how the keys are searched, priority of the keys, how the keys are merged.
- Etcd source: config.reload option and when the configuration is reloaded.
- Recipe: how to work with etcd.
- Recipe: how to start an instance without a local file config in EE (configure etcd source from env variables).

## Schema

https://github.com/tarantool/tarantool/blob/d79e7a3c05fb6d8d638e989093d82293f2f894b2/src/box/lua/config/instance_config.lua#L394

## Tests

https://github.com/tarantool/tarantool/blob/d79e7a3c05fb6d8d638e989093d82293f2f894b2/test/config-luatest/instance_config_schema_test.lua#L88

## Environment variables

```
tarantool --help-env-list | grep TT_CONFIG
| TT_CONFIG                                   | string         | nil                                           | Community Edition  |
| TT_CONFIG_ETCD_ENDPOINTS                    | array          | nil                                           | Enterprise Edition |
| TT_CONFIG_ETCD_HTTP_REQUEST_TIMEOUT         | number         | nil                                           | Enterprise Edition |
| TT_CONFIG_ETCD_HTTP_REQUEST_UNIX_SOCKET     | string         | nil                                           | Enterprise Edition |
| TT_CONFIG_ETCD_PASSWORD                     | string         | nil                                           | Enterprise Edition |
| TT_CONFIG_ETCD_PREFIX                       | string         | nil                                           | Enterprise Edition |
| TT_CONFIG_ETCD_SSL_CA_FILE                  | string         | nil                                           | Enterprise Edition |
| TT_CONFIG_ETCD_SSL_CA_PATH                  | string         | nil                                           | Enterprise Edition |
[...]

PR #3773, opened 2023-10-17
(Obsolete) 3.0 config

andreyaksenov:
**UPD: Obsolete, see https://github.com/tarantool/doc/pull/3825.**

- [Configuration](https://docs.d.tarantool.io/en/doc/3.0-config/concepts/configuration/)
- [Configuration in code](https://docs.d.tarantool.io/en/doc/3.0-config/concepts/configuration/configuration_code/)
- [Configuration reference (box.cfg)](https://docs.d.tarantool.io/en/doc/3.0-config/reference/configuration/)
- [Migrating to declarative configuration](https://docs.d.tarantool.io/en/doc/3.0-config/concepts/configuration/configuration_migrating/) (TODO: https://github.com/tarantool/doc/issues/3661)
- [Configuration reference](https://docs.d.tarantool.io/en/doc/3.0-config/reference/configuration/configuration_reference/) (TODO: https://github.com/tarantool/doc/issues/3664)
- [Storing configuration in etcd](https://docs.d.tarantool.io/en/doc/3.0-config/concepts/configuration/configuration_etcd/) (TODO: https://github.com/tarantool/doc/issues/3658)


doc/concepts/configuration.rst
```diff
@@ -0,0 +1,618 @@
...
+-   ``credentials`` (*global*)
+
+    Options in this section grant the specified privileges to the *replicator* user used for replication and
+    the *client* user that can perform any action.
+    These options are applied globally to all instances.
+
+-   ``iproto`` (*global*, *instance*)
+
+    The ``iproto`` section is specified on both global and instance levels.
+    The ``iproto.advertise.peer`` option specifies a user name and a host used by replicas to connect to each other.
```
Totktonada, line 133: Is it describe an idea of `iproto.advertise.peer` at whole or in regards of this particular example. It seems, it is inaccurate in both senses.

There are four variants of `iproto.advertise.peer` syntax, described in `instance_config.lua`.

In this particular example `username@` syntax is used. The option has no host information.

doc/concepts/configuration/configuration_code.rst
```diff
@@ -0,0 +1,341 @@
...
+
+
+..  _index-init_label:
+
+Initialization file
+-------------------
+
+If the command to :ref:`start Tarantool <configuration_code_run_instance_tarantool>` includes ``LUA_INITIALIZATION_FILE``, then
+Tarantool begins by invoking the Lua program in the file, which by convention
+may have the name ``script.lua``. The Lua program may get further arguments
```
Totktonada, line 27: `LUA_INITIALIZATION_FILE` is used as a term, while we have no specific term for the instance file. One may assume that there is a same named environment variable or other configuration facility that is related to the non-existing term.

`script.lua` is used like it is a usual name for the instance file, while it isn't. Several years ago there were no usual instance file name -- it is called according to a microservice name. Now, many applications that are based on cartridge framework use `init.lua`. It is not necessary, though.

I have two options in the mind. First, call it 'instance file'.

Second, if it is easier to write the documentation using certain name, add a remark.

> While the instance file name is arbitrary, this article uses `init.lua` file name for simplicity.

Or something of this kind.

doc/concepts/configuration/configuration_code.rst
```diff
@@ -0,0 +1,341 @@
...
+    box.cfg{
+        listen              = os.getenv("LISTEN_URI"),
+        memtx_memory        = 33554432,
+        pid_file            = "tarantool.pid",
+        wal_max_size        = 2500
+    }
+    print('Starting ', arg[1])
+
+and suppose the environment variable LISTEN_URI contains 3301,
+and suppose the command line is ``~/tarantool/src/tarantool script.lua ARG``.
```
Totktonada, line 47: Nit: I would use `tarantool` assuming that it is in the `PATH` directories.

doc/concepts/configuration/configuration_code.rst
```diff
@@ -0,0 +1,341 @@
...
+    ... main/101/script.lua I> mapping 33554432 bytes for memtx tuple arena...
+    ... main/101/script.lua I> recovery start
+    ... main/101/script.lua I> recovering from './00000000000000000000.snap'
+    ... main/101/script.lua I> set 'listen' configuration option to "3301"
+    ... main/102/leave_local_hot_standby I> ready to accept requests
+    Starting  ARG
+    ... main C> entering the event loop
+
+If you wish to start an interactive session on the same terminal after
+initialization is complete, you can use :ref:`console.start() <console-start>`.
```
Totktonada, line 65: I guess it is better to suggest the `-i` option. See https://github.com/tarantool/tarantool/issues/7017 for details.

doc/concepts/configuration/configuration_code.rst
```diff
@@ -0,0 +1,341 @@
...
+------------------------
+
+Configuration parameters have the form:
+
+:extsamp:`{**{box.cfg}**}{[{*{key = value}*} [, {*{key = value ...}*}]]}`
+
+Since ``box.cfg`` may contain many configuration parameters and since some of the
+parameters (such as directory addresses) are semi-permanent, it's best to keep
+``box.cfg`` in a Lua file. Typically this Lua file is the initialization file
+which is specified on the Tarantool command line.
```
Totktonada, line 120: > it's best to keep ``box.cfg`` in a Lua file

It's simplest, but why the best? Cartridge, for example, allow to store box.cfg parameters in the cluster wide config.

And it is unclear how it is related to non-dynamic options.

doc/concepts/configuration/configuration_code.rst
```diff
@@ -0,0 +1,341 @@
...
+
+Most configuration parameters are for allocating resources, opening ports, and
+specifying database behavior. All parameters are optional.
+A few parameters are dynamic, that is, they can be changed at runtime by calling ``box.cfg{}`` a second time.
```
Totktonada, line 124: > A few parameters are dynamic

A rough calculation gives ~90 dynamic parameters, 3/4 of the overall count.

doc/concepts/configuration/configuration_code.rst
```diff
@@ -0,0 +1,341 @@
...
+
+Some configuration parameters and some functions depend on a URI (Universal Resource Identifier).
+The URI string format is similar to the
+`generic syntax for a URI schema <https://en.wikipedia.org/wiki/List_of_URI_schemes>`_.
+It may contain (in order):
+
+* user name for login
+* password
+* host name or host IP address
+* port number.
```
Totktonada, line 176: (Looks unrelated to the changes in this PR, but I would highlight it.)

It seems, we should add query parameters into the list.

doc/concepts/configuration/configuration_code.rst
```diff
@@ -0,0 +1,341 @@
...
+
+    $ tarantool LUA_INITIALIZATION_FILE [OPTION ...]
+
+The ``tarantool`` command also provides a set of :ref:`options <configuration_command_options>` that might be helpful for development purposes.
+
+The command below starts a Tarantool instance configured in the ``script.lua`` file:
+
+.. code-block:: console
+
+    $ tarantool script.lua
```
Totktonada, line 341: Shouldn't we place it at beginning?

Totktonada (approved):
The changes that are present is this PR don't contract with my understanding of the topic. I've left some nits around particular details here. Feels free to proceed with them in any way you'll find convenient.

doc/concepts/configuration/configuration_code.rst
andreyaksenov: Thanks, updated to `Most of the parameters are dynamic`.

doc/concepts/configuration/configuration_code.rst
andreyaksenov: Thanks, will apply your suggestions

doc/concepts/configuration.rst
Totktonada: It seems that PR #3825 fixes the inaccuracy.

PR #3831, opened 2023-11-02
ci: added workflow to generate and upload docx

### LINKED ISSUE #3548: Build docx and PDF versions of doc
Use pandoc module from tdg-doc and build actual word doc and PDF absolutely the same with content on the website, including the table of content sections:

- [x] Build singlehtml with `make singlehtml`
- [x] Build doc with `make docx-only`
- [x] Build PDF with `make pdf`
- [x] Save PDF to S3  and make link from website (<lang>/doc/<version>/Tarantool.pdf)
- [x] Save docx to S3 and make link from website

lastoCHka42:
Added workflow to generate the DOCX documentation and upload it to the S3 storage.

Because `pandoc` works only on Debian Bookworm and documentation build                                                                             
works only on Debian Bullseye, we use separate jobs in separate                                                                                    
containers to build `singlepage` and `pandoc`.

Resolves #3548

### DIFF
--- a/.github/workflows/upload-docx.yml
+++ b/.github/workflows/upload-docx.yml
@@ -1,0 +1,94 @@
+name: Generate DOCX
+
+on:
+  push:
+    branches:
+      - 'latest'
+      - '3.0'
+      - 'lastochka42/generate-docx'
+
+jobs:
+  generate-singlehtml:
+    runs-on: ubuntu-latest
+    container: tarantool/doc-builder:fat-4.3
+    steps:
+      - uses: actions/checkout@v3
+        id: checkout
+        with:
+          submodules: recursive
+
+      - name: generate singlehtml
+        run: |
+          cmake .
+          make pull-modules
+          make build-modules
+          make singlehtml
+          make singlehtml-ru
+
+      - name: save singlehtml-en
+        uses: actions/upload-artifact@v3
+        with:
+          name: singlehtml-en
+          path: output/html/en/singlehtml.html
+
+      - name: save singlehtml-ru
+        uses: actions/upload-artifact@v3
+        with:
+          name: singlehtml-ru
+          path: output/html/ru/singlehtml.html
+
+      - name: save pandoc filters from the repo
+        uses: actions/upload-artifact@v3
+        with:
+          name: pandoc
+          path: pandoc/filter-espd.py
+  
+  generate-docx:
+    needs: generate-singlehtml
+    runs-on: ubuntu-latest
+    container: ghcr.io/tarantool/pandoc-builder:latest
+    env:
+      AWS_ACCESS_KEY_ID: ${{secrets.AWS_ACCESS_KEY_ID}}
+      AWS_SECRET_ACCESS_KEY: ${{secrets.AWS_SECRET_ACCESS_KEY}}
+      AWS_DEFAULT_REGION: ${{secrets.AWS_DEFAULT_REGION}}
+      S3_ENDPOINT_URL: ${{secrets.S3_ENDPOINT_URL}}
+      S3_UPLOAD_PATH: ${{secrets.S3_DOCX_PATH}}
+    steps:
+      - name: download artifacts
+        uses: actions/download-artifact@v3
+
+      - name: generate docx-en
+        run: |
+          pandoc singlehtml-en/singlehtml.html \
+            --standalone \
+            --verbose \
+            --from html \
+            --to docx \
+            --output tarantool-en.docx \
+            --metadata lang=en \
+            --metadata toc-title='Table of content' \
+            --filter pandoc/filter-espd.py \
+            --table-of-contents \
+            --toc-depth=2
+
+      - name: generate docx-ru
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/upload-docx.yml
```diff
@@ -0,0 +1,103 @@
...
+    runs-on: ubuntu-latest
+    container: python:3.8-slim-bookworm
+    env:
+      AWS_ACCESS_KEY_ID: ${{secrets.AWS_ACCESS_KEY_ID}}
+      AWS_SECRET_ACCESS_KEY: ${{secrets.AWS_SECRET_ACCESS_KEY}}
+      AWS_DEFAULT_REGION: ${{secrets.AWS_DEFAULT_REGION}}
+      S3_ENDPOINT_URL: ${{secrets.S3_ENDPOINT_URL}}
+      S3_UPLOAD_PATH: ${{secrets.S3_DOCX_PATH}}
+    steps:
+      - name: install dependencies
```
ochaplashkin: It seems to me that this should be done once (create and save an image with dependencies), and then use the already image in our workflow.

pandoc/filter-espd.py
```diff
@@ -0,0 +1,98 @@
...
+            attributes={'custom-style': 'image'}
+            # if hasattr(elem.parent, 'attributes'):
+            # debug(elem.parent)
+
+        return elem
+    if isinstance(elem, MetaInlines):
+        """
+        Title page header
+        """
+        global version
```
ochaplashkin: Global variables are a bad practice in programming (because magical effects occur).

pandoc/filter.py
```diff
@@ -0,0 +1,97 @@
...
+        return elem
+
+    if isinstance(elem, Link):
+        """
+        Transform html urls in docx format
+        """
+        elem.url = elem.url.replace('index-docx.html', '')
+        return elem
+
+    # if hasattr(elem, 'attributes'):
```
ochaplashkin: Why do we need this block here?

pandoc/filter.py
```diff
@@ -0,0 +1,97 @@
...
+    BulletList, RawBlock, Span, Emph, Strong
+
+logging.basicConfig(level=10)
+
+version = None
+
+pb = RawBlock(u"<w:p><w:r><w:br w:type=\"page\" /></w:r></w:p>", format=u"openxml")
+
+
+def action(elem, doc):
```
ochaplashkin: There is a more delicate solution for filtering.

My test data is: `[1, 'a', {}, [1, 'a', 2], (1, 2)]` and i want to:
1. filter by type;
2. if something is unknown, filter it "differently".

Default code:

```
def filter(e):
    if isinstance(e, int):
        return 'this is int'
    if isinstance(e, str):
        return 'this is str'
    if isinstance(e, dict):
        return 'this is dict'
    if isinstance(e, list) and isinstance(element[0], int):
        return 'this is list[int]'
    return 'this is some object'
```
This is not a bad design when we have a **"one-line"** logic.
However, if we have complex logic, this block becomes difficult to understand.

I suggest using the following approach:

```
def handle_int(e):
    return 'this is int'

def handle_str(e):
    return 'this is str'

def handle_dict(e):
    return 'this is dict'

def handle_list(e):
    if isinstance(e[0], int):
        return 'this is list[int]'

def handle_object(e):
    return 'this is some object'

switcher = {
    int: handle_int,
    str: handle_str,
    dict: handle_dict,
    list: handle_list
}

def handle_filter(e):
    return switcher.get(type(e), handle_object)(e)
```

Tests:
```
for e in test_data:
    print(filter(e) == handle_filter(e), handle_filter(e))
True this is int
True this is str
True this is dict
True this is list[int]
True this is some object
```

pandoc/filter.py
```diff
@@ -0,0 +1,97 @@
+#! /usr/bin/env python
+# encoding: utf-8
+
+import logging
+from panflute import Link, run_filter, Para, Str, Header, MetaInlines, Space, debug, Div, HorizontalRule, \
```
ochaplashkin:
```suggestion
from panflute import run_filter, debug
from panflute import Link, Para, Str, Header, MetaInlines, ...
```

ochaplashkin (changes requested):
General comment (about the approach): it seems to keep two files (filter.py and filter_espd.py ) with almost the same content(handlers) - not the best idea.

.github/workflows/upload-docx.yml
lastoCHka42: Fixed

pandoc/filter.py
lastoCHka42: Fixed

pandoc/filter.py
lastoCHka42: fixed

pandoc/filter-espd.py
lastoCHka42: fixed

ochaplashkin (approved):
LGTM :rocket: 
Good refactoring :+1:

Issue #3666, opened 2023-09-01
[Config] Document the 'credentials' option


andreyaksenov:
**Product:** Tarantool
**Since:** 3.0
**Root document:** https://www.tarantool.io/en/doc/latest/reference/configuration/configuration_reference/#credentials
**SME:** @ Totktonada
**Related doc issues:** https://github.com/tarantool/doc/issues/3953, https://github.com/tarantool/doc/issues/3628

# Details

A new `credentials` section introduced in a new YAML configuration requires documenting what are users, roles, and privileges. Currently this information lives in the [Access control](https://www.tarantool.io/en/doc/latest/book/admin/access_control/) topic related to user administration.

Given that we need earlier introduction of access control concepts, we can have the following content structure:

1) Add a new `Credentials` topic in the [Configuration](https://www.tarantool.io/en/doc/latest/concepts/configuration/) section. This topic describes specific of configuring service users (replication, sharding). All the concepts (users, roles, and privileges) are described in the existing `Access control` topic.
   - In a new topic, mention how to extract passwords from env variables and external files: https://github.com/tarantool/doc/issues/3954
2) Update [Access control](https://www.tarantool.io/en/doc/latest/book/admin/access_control/) to make it clearer. The [Privileges and Access Control explained](https://github.com/tarantool/tarantool/wiki/Privileges-and-Access-Control-explained) page might be very helpful here.
   - Add information about [new privileges](https://github.com/tarantool/doc/issues/3628).
   - Mention that specific roles ([sharding](https://github.com/tarantool/doc/issues/3953), replication) should be used for a cluster config (a new `Credentials` topic).
3) Reference page: 
    -   document all the options in the [credentials](https://www.tarantool.io/en/doc/latest/reference/configuration/configuration_reference/#credentials) section. See [instance_config.lua](https://github.com/tarantool/tarantool/blob/f58bfc97c627f1838696697339fb3e443c05767a/src/box/lua/config/instance_config.lua#L1499).
    - document new `config.context` options used to load secrets from safe storage. See [instance_config.lua](https://github.com/tarantool/tarantool/blob/f58bfc97c627f1838696697339fb3e443c05767a/src/box/lua/config/instance_config.lua#L602).

# Related admin functions

- https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_schema/user_create/
- https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_schema/user_grant/
[...]

PR #3834, opened 2023-11-03
3.0 config: replication reference

### LINKED ISSUE #3432: new configuration option - `bootstrap_leader`.
**Related dev. issue(s):** https://github.com/tarantool/tarantool/issues/7999

**Product:** Tarantool
**Since:** 3.0
**Root document:**
- https://github.com/tarantool/doc/issues/3664
- https://www.tarantool.io/en/doc/latest/reference/configuration/

**SME:** @ sergepetrenko

# Details

Configuration receives a new parameter - `bootstrap_leader`. The
parameter specifies the desired node to bootstrap from.

The parameter is valid only when another option - `bootstrap_strategy` -
is set to `'config'`.
The user may pass either the bootstrap leader's URI or its UUID string.

If bootstrap leader is passed as a UUID, the node will bootstrap from
the peer whose UUID matches the given one.

If bootstrap leader is passed as a URI, the URI must contain the same
host:port pair (or unix socket path) used for this node in
`box.cfg.replication`. In this case the node will bootstrap from the
remote peer listed in the corresponding replication entry.

For example, this call will return an error:
```lua
box.cfg{
    bootstrap_strategy = 'config',
    bootstrap_leader = 'localhost:3301',
    replication = {'127.0.0.1:3301'},
}
```
But this call will succeed:
```lua
box.cfg{
    bootstrap_strategy = 'config',
    bootstrap_leader = 127.0.0.1:3301',
    replication = {'127.0.0.1:3301'},
}
```
In the example above, setting either or both of `bootstrap_leader` and
`replication` to `"user:password@127.0.0.1:3301` will work just fine,
i.e. only the bootstrap leader's host and port are matched against
replication entries.

If `bootstrap_leader` points at the instance being configured, the same
URI entry for the instance must still be present in `replication`
configuration parameter, like this:
```lua
box.cfg{
    listen = 'localhost:1234',
    bootstrap_leader = 'localhost:1234',
    replication = {'localhost:1234', ...},
}
```
The only exception to this is setting `bootstrap_leader` to the same
UUID as instance_uuid, like this:
```lua
box.cfg{
    listen = 'something',
    bootstrap_leader = '11111111-1111-1111-1111-111111111111',
    instance_uuid = '11111111-1111-1111-1111-111111111111',
```
In this case the node will bootstrap the cluster on its own.
Requested by @ sergepetrenko in https://github.com/tarantool/tarantool/commit/c68fe0339430f449c9d0784c893eca1b57aafb51.

### LINKED ISSUE #3504: `box.cfg.bootstrap_leader` accepts instance names now
The option `box.cfg.bootstrap_leader`, which specifies the desired
bootstrap leader when bootstrap_strategy is "config" now accepts
instance names.

For example, this is a valid config without replication:
```lua
box.cfg{
    instance_name = 'main-server',
    bootstrap_strategy = 'config',
    bootstrap_leader = 'main-server'
}
```

When `box.cfg` contains some entries in `replication`, the node will
bootstrap from the node which has the instance name specified in
`box.cfg.bootstrap_leader`.

This is an addition to tarantool/doc#3432
Requested by @sergepetrenko in https://github.com/tarantool/tarantool/commit/8ceef02bef40fbe29db426e424a348b759de0bb0.

### LINKED ISSUE #3851: [Config] Document configuration options related to replication
**Next step**: https://github.com/tarantool/doc/issues/3654

**Product:** Tarantool
**Since:** 3.0
**Root document:** 
- https://docs.d.tarantool.io/en/doc/3.0-config-replication/reference/configuration/configuration_reference/
- https://docs.d.tarantool.io/en/doc/3.0-config-replication/concepts/configuration/

**SME:** @ Totktonada

# Details

Schema: https://github.com/tarantool/tarantool/blob/941a85c39735b789914da04acad3faabd63b5dd6/src/box/lua/config/instance_config.lua#L1068

Related YAML sections:

- replicasets (<replicaset_name>, leader)
- groups (<group_name>.replicasets)
- replication
- iproto (listen, advertise.peer)
- instances
- credentials (replicator user)
- database (mode: ro, rw; replica set UUID)

See notes here: https://github.com/tarantool/doc/issues/3654.

andreyaksenov:
This PR adds [reference for configuration options related to replication](https://docs.d.tarantool.io/en/doc/3.0-config-replication/reference/configuration/configuration_reference/). Here are a summary of all changes:

- The `code_snippets/snippets` folder contains runnable tt applications referenced in docs. In addition to exiting examples, this PR adds the `bootstrap_leader`, `peers`, and `sharding` samples. These samples are included in a reference page in various places to demonstrate various features. Moreover, all these samples will be reused in other places in docs (tutorials, how-tos, etc.) for keeping consistency and making sure all the examples work fine.
- In addition to `config` and `config.etcd`, the following sections are added to `Configuration reference`:
    - [groups](https://docs.d.tarantool.io/en/doc/3.0-config-replication/reference/configuration/configuration_reference/#groups) (with the nested `replicasets` and `instances` sections)
    - [credentials](https://docs.d.tarantool.io/en/doc/3.0-config-replication/reference/configuration/configuration_reference/#credentials) (will be documented in https://github.com/tarantool/doc/issues/3666)
    - [database](https://docs.d.tarantool.io/en/doc/3.0-config-replication/reference/configuration/configuration_reference/#database)
    - [iproto](https://docs.d.tarantool.io/en/doc/3.0-config-replication/reference/configuration/configuration_reference/#iproto)
    - [replication](https://docs.d.tarantool.io/en/doc/3.0-config-replication/reference/configuration/configuration_reference/#replication)
- Added/updated READMEs.
- Updated tt configs to the 2.0 version.

Note that `replication.anon` and `database.hot_standby` are not documented yet. Will create separate tickets for documenting these options (https://github.com/tarantool/doc/issues/3857 and https://github.com/tarantool/doc/issues/3858).

Links to new configuration options in main docs are not updated yet. Most likely, this will be done in separate PRs when updating corresponding topics.

>  Additional changes in the base topic (not directly related to this PR):
    - https://docs.d.tarantool.io/en/doc/3.0-config-replication/concepts/configuration/#configuration-scopes-replica-set-example
    - https://docs.d.tarantool.io/en/doc/3.0-config-replication/concepts/configuration/#access-control

### DIFF
--- a/doc/code_snippets/snippets/config/tt.yaml
+++ b/doc/code_snippets/snippets/config/tt.yaml
@@ -1,63 +1,54 @@
-tt:
-  modules:
-    # Directory where the external modules are stored.
-    directory: "modules"
+modules:
+  # Directory where the external modules are stored.
+  directory: modules
 
-  app:
-    # Directory that stores various instance runtime
-    # artifacts like console socket, PID file, etc.
-    run_dir: "var/run"
+env:
+  # Restart instance on failure.
+  restart_on_failure: false
 
-    # Directory that stores log files.
-    log_dir: var/log
+  # Directory that stores binary files.
+  bin_dir: bin
 
-    # The maximum size in MB of the log file before it gets rotated.
-    log_maxsize: 100
+  # Directory that stores Tarantool header files.
+  inc_dir: include
 
-    # The maximum number of days to retain old log files.
-    log_maxage: 8
+  # Path to directory that stores all applications.
+  # The directory can also contain symbolic links to applications.
+  instances_enabled: instances.enabled
 
-    # The maximum number of old log files to retain.
-    log_maxbackups: 10
+  # Tarantoolctl artifacts layout compatibility: if set to true tt will not create application
+  # sub-directories for control socket, pid files, log files, etc.. Data files (wal, vinyl,
+  # snap) and multi-instance applications are not affected by this option.
+  tarantoolctl_layout: false
 
-    # Restart instance on failure.
-    restart_on_failure: false
+app:
+  # Directory that stores various instance runtime
+  # artifacts like console socket, PID file, etc.
+  run_dir: var/run
 
-    # Directory where write-ahead log (.xlog) files are stored.
-    wal_dir: "var/lib"
+  # Directory that stores log files.
+  log_dir: var/log
 
-    # Directory where memtx stores snapshot (.snap) files.
-    memtx_dir: "var/lib"
+  # Directory where write-ahead log (.xlog) files are stored.
+  wal_dir: var/lib
 
-    # Directory where vinyl files or subdirectories will be stored.
-    vinyl_dir: "var/lib"
+  # Directory where memtx stores snapshot (.snap) files.
+  memtx_dir: var/lib
 
-    # Directory that stores binary files.
-    bin_dir: "bin"
+  # Directory where vinyl files or subdirectories will be stored.
+  vinyl_dir: var/lib
 
-    # Directory that stores Tarantool header files.
-    inc_dir: "include"
+# Path to file with credentials for downloading Tarantool Enterprise Edition.
+# credential_path: /path/to/file
+ee:
+  credential_path:
 
-    # Path to directory that stores all applications.
-    # The directory can also contain symbolic links to applications.
-    instances_enabled: "instances.enabled"
+templates:
+  # The path to templates search directory.
+  - path: templates
 
-    # Tarantoolctl artifacts layout compatibility: if set to true tt will not create application
-    # sub-directories for control socket, pid files, log files, etc.. Data files (wal, vinyl,
-    # snap) and multi-instance applications are not affected by this option.
-    tarantoolctl_layout: false
-
-  # Path to file with credentials for downloading Tarantool Enterprise Edition.
-  # credential_path: /path/to/file
-  ee:
-    credential_path: ""
-
-  templates:
-    # The path to templates search directory.
-    - path: "templates"
-
-  repo:
-    # Directory where local rocks files could be found.
-    rocks: ""
-    # Directory that stores installation files.
-    distfiles: "distfiles"
+repo:
+  # Directory where local rocks files could be found.
+  rocks:
+  # Directory that stores installation files.
+  distfiles: distfiles
--- a/doc/code_snippets/snippets/replication/instances.enabled/manual_leader/config.yaml
+++ b/doc/code_snippets/snippets/replication/instances.enabled/manual_leader/config.yaml
@@ -3,9 +3,6 @@ credentials:
     replicator:
       password: 'topsecret'
       roles: [replication]
-    client:
-      password: 'secret'
-      roles: [super]
 
 iproto:
   advertise:
--- a/doc/reference/configuration/configuration_reference.rst
+++ b/doc/reference/configuration/configuration_reference.rst
@@ -3,12 +3,279 @@
 Configuration reference
 =======================
 
+..  TODO
+    https://github.com/tarantool/doc/issues/3664
+
 This topic describes all :ref:`configuration parameters <configuration>` provided by Tarantool.
+
+
+..  _configuration_reference_config_parameter:
+
+<config_parameter>
+------------------
+
+Any configuration parameter that can be defined in the global :ref:`scope <configuration_scopes>`.
+For example, :ref:`iproto <configuration_reference_iproto>` and :ref:`database <configuration_reference_database>` configuration parameters defined at the global level are applied to all instances in the cluster.
+
+
+..  _configuration_reference_groups:
+
+groups
+------
+
+The ``groups`` section provides the ability to define the :ref:`full topology of a Tarantool cluster <configuration_overview>`.
+
+.. NOTE::
+
+    ``groups`` can be defined in the global :ref:`scope <configuration_scopes>` only.
+
+-   :ref:`groups.\<group_name\> <configuration_reference_groups_name>`
+-   :ref:`groups.\<group_name\>.replicasets <configuration_reference_groups_name_replicasets>`
+-   :ref:`groups.\<group_name\>.\<config_parameter\> <configuration_reference_groups_name_config_parameter>`
+
+.. _configuration_reference_groups_name:
+
+.. confval:: groups.<group_name>
+
+    A group name.
+
+
+.. _configuration_reference_groups_name_replicasets:
+
+.. confval:: groups.<group_name>.replicasets
+
+    Replica sets that belong to this group. See :ref:`replicasets <configuration_reference_replicasets>`.
+
+
+.. _configuration_reference_groups_name_config_parameter:
+
+.. confval:: groups.<group_name>.<config_parameter>
+
+    Any configuration parameter that can be defined in the group :ref:`scope <configuration_scopes>`.
+    For example, :ref:`iproto <configuration_reference_iproto>` and :ref:`database <configuration_reference_database>` configuration parameters defined at the group level are applied to all instances in this group.
+
+
+
+..  _configuration_reference_replicasets:
+
+replicasets
+~~~~~~~~~~~
+
+.. NOTE::
+
+    ``replicasets`` can be defined in the group :ref:`scope <configuration_scopes>` only.
+
+-   :ref:`replicasets.\<replicaset_name\> <configuration_reference_replicasets_name>`
+-   :ref:`replicasets.\<replicaset_name\>.leader <configuration_reference_replicasets_name_leader>`
+-   :ref:`replicasets.\<replicaset_name\>.bootstrap_leader <configuration_reference_replicasets_name_bootstrap_leader>`
+-   :ref:`replicasets.\<replicaset_name\>.instances <configuration_reference_replicasets_name_instances>`
+-   :ref:`replicasets.\<replicaset_name\>.\<config_parameter\> <configuration_reference_replicasets_name_config_parameter>`
+
+.. _configuration_reference_replicasets_name:
+
+.. confval:: replicasets.<replicaset_name>
+
+    A replica set name.
+
+
+.. _configuration_reference_replicasets_name_leader:
+
+.. confval:: replicasets.<replicaset_name>.leader
+
+    A replica set leader.
+    This option is in effect when ``manual`` :ref:`replication.failover <configuration_reference_replication_failover>` is used.
+
+    **Example**
+
+    ..  literalinclude:: /code_snippets/snippets/replication/instances.enabled/manual_leader/config.yaml
+        :language: yaml
+        :start-at: replicaset001:
+        :end-at: listen: 127.0.0.1:3303
+        :dedent:
+
+
+.. _configuration_reference_replicasets_name_bootstrap_leader:
+
+.. confval:: replicasets.<replicaset_name>.bootstrap_leader
+
+    A bootstrap leader for a replica set.
+    To specify a bootstrap leader manually, you need to set :ref:`replication.bootstrap_strategy <configuration_reference_replication_bootstrap_strategy>` to ``config``.
+
+    **Example**
+
+    ..  literalinclude:: /code_snippets/snippets/replication/instances.enabled/bootstrap_strategy/config.yaml
+        :language: yaml
+        :start-at: replicaset001
+        :end-at: listen: 127.0.0.1:3303
+        :dedent:
+
+
+.. _configuration_reference_replicasets_name_instances:
+
+.. confval:: replicasets.<replicaset_name>.instances
+
+    Instances that belong to this replica set. See :ref:`instances <configuration_reference_instances>`.
+
+
+.. _configuration_reference_replicasets_name_config_parameter:
+
+.. confval:: replicasets.<replicaset_name>.<config_parameter>
+
+    Any configuration parameter that can be defined in the replica set :ref:`scope <configuration_scopes>`.
+    For example, :ref:`iproto <configuration_reference_iproto>` and :ref:`database <configuration_reference_database>` configuration parameters defined at the replica set level are applied to all instances in this replica set.
+
+
+
+..  _configuration_reference_instances:
+
+instances
+*********
+
+.. NOTE::
+
+    ``instances`` can be defined in the replica set :ref:`scope <configuration_scopes>` only.
+
+-   :ref:`instances.\<instance_name\> <configuration_reference_instances_name>`
+-   :ref:`instances.\<instance_name\>.\<config_parameter\> <configuration_reference_instances_name_config_parameter>`
+
+.. _configuration_reference_instances_name:
+
+.. confval:: instances.<instance_name>
+
+    An instance name.
+
+
+.. _configuration_reference_instances_name_config_parameter:
+
+.. confval:: instances.<instance_name>.<config_parameter>
+
+    Any configuration parameter that can be defined in the instance :ref:`scope <configuration_scopes>`.
+    For example, :ref:`iproto <configuration_reference_iproto>` and :ref:`database <configuration_reference_database>` configuration parameters defined at the instance level are applied to this instance only.
+
+
+..  _configuration_reference_credentials:
+
+credentials
+-----------
+
+..  TODO: https://github.com/tarantool/doc/issues/3666
+
+.. NOTE::
+
+    ``credentials`` can be defined in any :ref:`scope <configuration_scopes>`.
+
+
+..  _configuration_reference_credentials_role:
+
+roles
+~~~~~
+
+.. _configuration_reference_credentials_roles:
+
+.. confval:: credentials.roles
+
+    | Type: map
+    | Default: nil
+    | Environment variable: TT_CREDENTIALS_ROLES
+
+.. _configuration_reference_credentials_roles_name_roles:
+
+.. confval:: credentials.roles.<role_name>.roles
+
+
+.. _configuration_reference_credentials_roles_name_privileges:
+
+.. confval:: credentials.roles.<role_name>.privileges
+
+    See :ref:`privileges <configuration_reference_credentials_privileges>`.
+
+
+..  _configuration_reference_credentials_user:
+
+users
+~~~~~
+
+
+.. _configuration_reference_credentials_users:
+
+.. confval:: credentials.users
+
+    | Type: map
+    | Default: nil
+    | Environment variable: TT_CREDENTIALS_USERS
+
+
+.. _configuration_reference_credentials_users_name_password:
+
+.. confval:: credentials.users.<username>.password
+
+
+.. _configuration_reference_credentials_users_name_roles:
+
+.. confval:: credentials.users.<username>.roles
+
+
+.. _configuration_reference_credentials_users_name_privileges:
+
+.. confval:: credentials.users.<username>.privileges
+
+    See :ref:`privileges <configuration_reference_credentials_privileges>`.
+
+
+..  _configuration_reference_credentials_privileges:
+
+privileges
+~~~~~~~~~~
+
+.. _configuration_reference_credentials_users_name_privileges_permissions:
+
+.. confval:: <user_or_role_name>.privileges.permissions
+
+
+.. _configuration_reference_credentials_users_name_privileges_spaces:
+
+.. confval:: <user_or_role_name>.privileges.spaces
+
+
+.. _configuration_reference_credentials_users_name_privileges_functions:
+
+.. confval:: <user_or_role_name>.privileges.functions
+
+
+.. _configuration_reference_credentials_users_name_privileges_sequences:
+
+.. confval:: <user_or_role_name>.privileges.sequences
+
+
+.. _configuration_reference_credentials_users_name_privileges_lua_eval:
+
+.. confval:: <user_or_role_name>.privileges.lua_eval
+
+
+.. _configuration_reference_credentials_users_name_privileges_lua_call:
+
+.. confval:: <user_or_role_name>.privileges.lua_call
+
+
+.. _configuration_reference_credentials_users_name_privileges_sql:
+
+.. confval:: <user_or_role_name>.privileges.sql
+
+
+
+
+
 
 ..  _configuration_reference_config:
 
 config
 ------
+
+The ``config`` section defines various parameters related to centralized configuration.
+
+.. NOTE::
+
+    ``config`` can be defined in the global :ref:`scope <configuration_scopes>` only.
 
 * :ref:`config.reload <configuration_reference_config_reload>`
 * :ref:`config.version <configuration_reference_config_version>`
@@ -50,8 +317,8 @@ config
 
 .. _configuration_reference_config_etcd:
 
-etcd
-~~~~
+config.etcd.*
+~~~~~~~~~~~~~
 
 ..  include:: /concepts/configuration/configuration_etcd.rst
     :start-after: ee_note_etcd_start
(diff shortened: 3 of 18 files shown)


doc/code_snippets/snippets/replication/instances.enabled/manual_leader/config.yaml
```diff
@@ -3,9 +3,6 @@ credentials:
     replicator:
       password: 'topsecret'
       roles: [replication]
-    client:
-      password: 'secret'
-      roles: [super]
```
Totktonada, line 8: I'm in doubts here. We have a replicaset and nobody can connect to it (except using an admin console).

doc/reference/configuration/configuration_reference.rst
```diff
@@ -3,13 +3,280 @@
...
+
+    A replica set name.
+
+
+.. _configuration_reference_replicasets_name_leader:
+
+.. confval:: replicasets.<replicaset_name>.leader
+
+    A replica set leader.
+    This option is in effect when ``manual`` :ref:`replication.failover <configuration_reference_replication_failover>` is used.
```
Totktonada: No `leader` or `leader: null` is an important case. A usual master-replica switch process assumes a short period, when all the instances are in RO.

A correct master switch in the `failover: manual` mode would involve the following steps:

1. Remove the `leader` key or set it to `null`.
2. Wait until all the data goes to the future master (compare vclock values).
3. Set the new leader using the `leader` key.

(BTW, things are pretty same for the `failover: off` mode if a single master at time is assumed.)

doc/reference/configuration/configuration_reference.rst
```diff
@@ -3,13 +3,280 @@
...
+
+    **Example**
+
+    ..  literalinclude:: /code_snippets/snippets/replication/instances.enabled/manual_leader/config.yaml
+        :language: yaml
+        :start-at: replicaset001:
+        :end-at: listen: 127.0.0.1:3303
+        :dedent:
```
Totktonada: Can we highlight the line with the `leader` option here?

doc/reference/configuration/configuration_reference.rst
```diff
@@ -3,13 +3,280 @@
...
+    A bootstrap leader for a replica set.
+    To specify a bootstrap leader manually, you need to set :ref:`replication.bootstrap_strategy <configuration_reference_replication_bootstrap_strategy>` to ``config``.
+
+    **Example**
+
+    ..  literalinclude:: /code_snippets/snippets/replication/instances.enabled/bootstrap_strategy/config.yaml
+        :language: yaml
+        :start-at: replicaset001
+        :end-at: listen: 127.0.0.1:3303
+        :dedent:
```
Totktonada: Can we highlight the lines with the `boostrap_strategy` and the `boostrap_leader` options?

BTW, why `failover: election` is here? On the website it looks like it is included intentionally, but I guess that it isn't.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -3,13 +3,280 @@
...
 This topic describes all :ref:`configuration parameters <configuration>` provided by Tarantool.
 
+
+..  _configuration_reference_config_parameter:
+
+<config_parameter>
+------------------
+
+Any configuration parameter that can be defined in the global :ref:`scope <configuration_scopes>`.
+For example, :ref:`iproto <configuration_reference_iproto>` and :ref:`database <configuration_reference_database>` configuration parameters defined at the global level are applied to all instances in the cluster.
```
Totktonada: AFAIU, it is just a placeholder, a kind of 'any parameter, but see scope constraints for certain parameters'.

It feels confusing that we highliht the global scope here. Can we consider some alternative wordings?

doc/reference/configuration/configuration_reference.rst
```diff
@@ -3,13 +3,280 @@
...
+.. confval:: instances.<instance_name>.<config_parameter>
+
+    Any configuration parameter that can be defined in the instance :ref:`scope <configuration_scopes>`.
+    For example, :ref:`iproto <configuration_reference_iproto>` and :ref:`database <configuration_reference_database>` configuration parameters defined at the instance level are applied to this instance only.
+
+
+..  _configuration_reference_credentials:
+
+credentials
+-----------
```
Totktonada: If the order is supposed to be alphabetical, `config` should go before `credentials`. Moreover, I would describe the `config` section first, because it is a bit specific: more about the configuration itself rather than the configuration of an instance.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -3,13 +3,280 @@
...
+
+.. NOTE::
+
+    ``credentials`` can be defined in any :ref:`scope <configuration_scopes>`.
+
+
+..  _configuration_reference_credentials_role:
+
+roles
+~~~~~
```
Totktonada: It looks a bit confusing on the website:

* credentials
  * roles
    * credentials.roles
    * credentials.roles.<...>

I mean that we use just `roles` on the second level, but a fully qualified name at the lower level.

And it is inconsistent with options of the `config` section. It looks so:

* config
  * config.reload (not just reload!)
  * config.version
  * config.etcd.*

Totktonada: I would also add that without some indentation non-qualified names are confusing for me. I'm scrolling here and there and sometimes it is hard to spot, whether, say, the given title is about top level option `roles` or about sub-section `roles` in the `credentials` section. Some difference in title font sizes helps, but only after some practice.

I guess that it is not an easy task to restyle the website in order to better render hierarchical documents. So I think that using fully qualified names would be a good choice.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -3,13 +3,280 @@
...
+
+.. confval:: credentials.users.<username>.privileges
+
+    See :ref:`privileges <configuration_reference_credentials_privileges>`.
+
+
+..  _configuration_reference_credentials_privileges:
+
+privileges
+~~~~~~~~~~
```
Totktonada: It is a kind of template with two references to it (from `credentials.users.<username>.privileges` and `credentials.roles.<rolename>.privileges`). However, visually it looks like an ordinary sub-section within the hierarchy.

I think we should mark such templates somehow: some symbols like `<privileges>` (that looks a bit ugly for me, but anyway better) or some block with borders around.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -221,6 +488,718 @@ This section describes options related to :ref:`storing configuration in etcd <c
...
+    ``database.instance_uuid`` can be used to specify an instance identifier manually.
+
+    UUIDs should follow these rules:
+
+    *   The values must be true unique identifiers, not shared by other instances
+        or replica sets within the common infrastructure.
+
+    *   The values must be used consistently, not changed after the initial setup.
+        The initial values are stored in :ref:`snapshot files <index-box_persistence>`
+        and are checked whenever the system is restarted.
```
Totktonada, line 412: However, they can be deleted from the configuration. Don't know, should we say it here or not. It is important part of the upgrade process from 2.x.

The process is conceptually the following:

* Start with the given UUIDs from the existing snapshots.
* Upgrade the database schema (`box.schema.upgrade()` + `box.snapshot()`).
* Drop the UUIDs (because all the needed checks are now performed using the instance names).

(I can miss some details here, but that's the idea.)

So, ability to drop the instance/replicaset UUID from the config seems important.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -221,6 +488,718 @@ This section describes options related to :ref:`storing configuration in etcd <c
...
+    | Type: string
+    | Default: box.NULL
+    | Environment variable: TT_DATABASE_INSTANCE_UUID
+
+
+.. _configuration_reference_database_mode:
+
+.. confval:: database.mode
+
+    An instance's operating mode.
```
Totktonada, line 431: I would say that it seems important to highlight that this option is available only in the `failover: off` mode. (The relevant [check](https://github.com/tarantool/tarantool/blob/2c422a0b98675cab480ebb44256dfeb72bd2f820/src/box/lua/config/configdata.lua#L460-L478).)

doc/reference/configuration/configuration_reference.rst
```diff
@@ -221,6 +488,718 @@ This section describes options related to :ref:`storing configuration in etcd <c
...
+    The following modes are available:
+
+    -   ``rw``: an instance is in read-write mode.
+    -   ``ro``: an instance is in read-only mode.
+
+    If this option is not specified explicitly, its default value depends on the number of instances in a replica set. For a single instance, the ``rw`` mode is used, while for multiple instances, the ``ro`` mode is used.
```
Totktonada: It is (almost) correct for `failover: off` mode. However, the logic is different for different failover modes. Look [here](https://github.com/tarantool/tarantool/blob/2c422a0b98675cab480ebb44256dfeb72bd2f820/src/box/lua/config/applier/box_cfg.lua#L347-L487). It is a bit complicated and, maybe, requires some separate effort to explain to users.

Why almost?

There is the special logic on startup, which (in some cases, when it is possible) starts an instance that is configured as RW in the RO mode. After that it re-reads the configuration and if the instance still configured as RW, goes to RW. The idea is to prevent two masters in the off/manual failover modes.

Say, an administrator wants to recover a master, which want off beforehand. The administrator finds that the instance reads snapshots for a long time and decides to appoint one of the running instances as the new master. The configuration is changed and the new master goes to RW. After some time, the old master finish reading its snapshot and ~goes to RW as well~ goes to RO, re-reads config, keep itself as RO.

The relevant startup code is [here](https://github.com/tarantool/tarantool/blob/2c422a0b98675cab480ebb44256dfeb72bd2f820/src/box/lua/config/applier/box_cfg.lua#L563-L680). It also requires some separate effort to describe properly.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -221,6 +488,718 @@ This section describes options related to :ref:`storing configuration in etcd <c
...
+.. _configuration_reference_database_txn_timeout:
+
+.. confval:: database.txn_timeout
+
+    A timeout (in seconds) after which the transaction is rolled back.
+
+    See also: :ref:`box.begin() <box-begin>`
+
+    | Type: number
+    | Default: 3153600000
```
Totktonada: I think that such large numbers in seconds requires some short explanations, like so:

> Default: 3153600000 (~100 years)

doc/reference/configuration/configuration_reference.rst
```diff
@@ -221,6 +488,718 @@ This section describes options related to :ref:`storing configuration in etcd <c
...
+These options accept an URI in the following formats:
+
+-   User :ref:`credentials <configuration_reference_credentials>` and an address: ``username@host:port`` or ``username:password@host:port``.
+
+-   User credentials: ``username@`` or ``username:password@``.
+    In this case, an advertise address is taken from :ref:`iproto.listen <configuration_reference_iproto_listen>`.
+
+-   An address: ``host:port``.
+
+If ``password`` is missing, it is taken from :ref:`credentials <configuration_reference_credentials>` for the specified ``username``.
```
Totktonada: It is applicable to `peer` and `sharding`, but not `client`. The latter only accepts `host:port` (or a unix socket without credentials). See the [description](https://github.com/tarantool/tarantool/blob/2c422a0b98675cab480ebb44256dfeb72bd2f820/src/box/lua/config/instance_config.lua#L676-L723) of the options.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -221,6 +488,718 @@ This section describes options related to :ref:`storing configuration in etcd <c
...
+
+You can also use a Unix domain socket (``unix/:``) instead of ``host:port``.
+
+.. NOTE::
+
+    The ``host`` value cannot be ``0.0.0.0``.
```
Totktonada: It also can't be `[::]`. And the port can't be zero. See [here](https://github.com/tarantool/tarantool/blob/2c422a0b98675cab480ebb44256dfeb72bd2f820/src/box/lua/config/instance_config.lua#L686-L687).

doc/reference/configuration/configuration_reference.rst
```diff
@@ -221,6 +488,718 @@ This section describes options related to :ref:`storing configuration in etcd <c
...
+    In the example below, the following configuration options are specified:
+
+    -   In the :ref:`credentials <configuration_reference_credentials>` section, the ``replicator`` user with the ``replication`` role is created.
+    -   ``iproto.advertise.peer`` specifies that other instances should connect to an address defined in :ref:`iproto.listen <configuration_reference_iproto_listen>` using the ``replicator`` user.
+
+    ..  literalinclude:: /code_snippets/snippets/replication/instances.enabled/auto_leader/config.yaml
+        :language: yaml
+        :start-at: credentials:
+        :end-at: peer: replicator@
+        :dedent:
```
Totktonada: Shouldn't we show `iproto.listen` in this snippet too?

I would also highlight a line with `iproto.advertise.peer`.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -221,6 +488,718 @@ This section describes options related to :ref:`storing configuration in etcd <c
...
+    | Type: string
+    | Default: box.NULL
+    | Environment variable: TT_IPROTO_ADVERTISE_CLIENT
+
+.. _configuration_reference_iproto_advertise_peer:
+
+.. confval:: iproto.advertise.peer
+
+    An URI used to advertise the current instance to other cluster members.
+    For example, it can be used by an instance to connect to another instance as a replica.
```
Totktonada: I think 'can be used' is not really the same as 'enables super cool automatic that builds replicaset peer URIs for each instance' :)

I mean, the idea of the option is to let the automatic do the job. 'Can be used' means just that this information is present.

BTW, 'as a replica' is a bit inaccurate. The same URI list build logic works on a master too. I would use neural 'instance' or 'replicaset peer' in such contexts.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -221,6 +488,718 @@ This section describes options related to :ref:`storing configuration in etcd <c
...
+
+.. _configuration_reference_iproto_listen:
+
+.. confval:: iproto.listen
+
+    An address used to listen for incoming requests.
+    This address is used for different purposes, for example:
+
+    -   Communicating between cluster members.
+    -   Connecting to an instance using an :ref:`admin console <admin-security>`.
```
Totktonada: The admin console is a separate concept. First, it works by another protocol (text, not binary). Next, it is usually local (a unix socket). It is configured by another option: `console.socket`.

Totktonada: Well, `tt connect` can work using the binary protocol (over iproto), it creates some confusion.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -221,6 +488,718 @@ This section describes options related to :ref:`storing configuration in etcd <c
...
+    -   Connecting to an instance using :ref:`external APIs <index-box_connectors>`.
```
Totktonada: Hmm. Why 'external APIs'? It looks dubious in this context. I would say just 'connectors', 'drivers' or 'languages'. Or just 'external clients'.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -221,6 +488,718 @@ This section describes options related to :ref:`storing configuration in etcd <c
...
+..  _configuration_reference_replication:
+
+replication
+-----------
+
+The ``replication`` section defines configuration parameters related to :ref:`replication <replication>`.
+
+.. NOTE::
+
+    ``replication`` can be defined in the global, group, and replica set :ref:`scope <configuration_scopes>`.
```
Totktonada: It is not exactly so.

The only option of this section that is really constrained by a scope is `replication.failover` ([code](https://github.com/tarantool/tarantool/blob/2c422a0b98675cab480ebb44256dfeb72bd2f820/src/box/lua/config/instance_config.lua#L83-L104)).

It is actually have a little sense to set most of the options of the `replication` section for individual instances. However, `replication.anon` is to be used per-instance.

I would move the note about the contraint to `replication.failover` and I wouldn't say anything about `replication` section at whole (because there is no common rule).

doc/reference/configuration/configuration_reference.rst
```diff
@@ -221,6 +488,718 @@ This section describes options related to :ref:`storing configuration in etcd <c
...
+
+    *   ``candidate``
+
+        A node should be able to become a leader.
+
+    *   ``manual``
+
+        Allow to control which instance is the leader explicitly instead of relying on automated leader election.
+
+        By default, the instance acts like a voter -- it is read-only and may vote for other candidate instances.
```
Totktonada, line 981: `replication.election_mode` is effectively ignored if `failover` is set to something other than `election`. And it is `candidate` by default if `failover: election` is enabled. See [here](https://github.com/tarantool/tarantool/blob/2c422a0b98675cab480ebb44256dfeb72bd2f820/src/box/lua/config/applier/box_cfg.lua#L376-L380).

Totktonada: Ouch, got it. It is about `replication.election_mode: manual`. I interpreted this paragraph as if it is about the `replication.election_mode` itself.

Anyway, the information above is missing here and it worth to include it. The default is marked as `box.NULL`. It is technically correct (`config:get('replication.election_mode')` will return this value), but the effective default depends on `replication.failover`. Such effective defaults are usually marked in the config schema code, see [here](https://github.com/tarantool/tarantool/blob/2c422a0b98675cab480ebb44256dfeb72bd2f820/src/box/lua/config/instance_config.lua#L1181-L1183).

doc/reference/configuration/configuration_reference.rst
```diff
@@ -221,6 +488,718 @@ This section describes options related to :ref:`storing configuration in etcd <c
...
+    during every new election, from 100% to 110% of the original timeout value.
+    For example, if the timeout is 300 ms and there are 3 nodes started
+    the election simultaneously in the same term,
+    they can set their election timeouts to 300, 310, and 320 respectively,
+    or to 305, 302, and 324, and so on. In that way, the votes will never be split
+    because the election on different nodes won't be restarted simultaneously.
+
+    | Type: number
+    | Default: 5
+    | Environment variable: TT_REPLICATION_ELECTION_TIMEOUT
```
Totktonada: Don't know why, but these blocks aren't splitted on the website.

image

The same for the next option.

Totktonada:
Thank you for the decent work! Please, consider my comments above.

doc/code_snippets/snippets/replication/instances.enabled/manual_leader/config.yaml
andreyaksenov: Yep, I understand your doubts but I'd like that replication samples demonstrate only options related to replication to make them more concise. For credentials options, I plan to create a separate sample that will demonstrate all use cases related to connecting, executing functions, and so on:

https://docs.d.tarantool.io/en/doc/3.0-config-replication/reference/configuration/configuration_reference/#credentials

doc/reference/configuration/configuration_reference.rst
andreyaksenov: Thanks for the clarification - very useful info for the upcoming replication tutorials. Will update the reference too.

doc/reference/configuration/configuration_reference.rst
andreyaksenov: Renamed the `<config_parameter>` section to `Configuration scopes` and rephrased a bit.

doc/reference/configuration/configuration_reference.rst
andreyaksenov: Sorted all sections alphabetically, `config` now follows the first `Configuration scopes` section.

doc/reference/configuration/configuration_reference.rst
andreyaksenov: Changed to `<user_or_role_name>.privileges`

doc/reference/configuration/configuration_reference.rst
andreyaksenov: Looks like this should be done when creating this guide: https://github.com/tarantool/doc/issues/3661
Added a note there.

doc/reference/configuration/configuration_reference.rst
andreyaksenov: > Shouldn't we show iproto.listen in this snippet too?

Yep, this might be helpful

doc/reference/configuration/configuration_reference.rst
andreyaksenov: Looks like we need to add links to the above section `iproto.advertise.*`, which describes all the details common for all the options.

doc/reference/configuration/configuration_reference.rst
andreyaksenov: Changed to:
```
Connecting to an instance remotely using an :ref:`admin console <admin-security>`.
```

doc/reference/configuration/configuration_reference.rst
andreyaksenov: Changed to:
```
Connecting to an instance using :ref:`connectors <index-box_connectors>` for different languages.
```

doc/reference/configuration/configuration_reference.rst
andreyaksenov: Looks like a vertical pipe (`|`) without any content add an empty string between blocks.

doc/reference/configuration/configuration_reference.rst
andreyaksenov: > BTW, why failover: election is here? On the website it looks like it is included intentionally, but I guess that it isn't.

Am I correct that it's is better to move `failover: election` at the global level to show only `replication. bootstrap_strategy` here? If yes, done:

https://docs.d.tarantool.io/en/doc/3.0-config-replication/reference/configuration/configuration_reference/#configuration-reference-replicasets-name-bootstrap-leader

doc/reference/configuration/configuration_reference.rst
andreyaksenov: My bad :) Please see the current staging here:

https://docs.d.tarantool.io/en/doc/3.0-config-replication/reference/configuration/configuration_reference/#credentials

doc/reference/configuration/configuration_reference.rst
andreyaksenov: Sound complicated for configuration reference :) Looks like such info should live in main topics, for example:
- https://www.tarantool.io/en/doc/latest/concepts/replication/repl_architecture/
- https://www.tarantool.io/en/doc/latest/dev_guide/internals/replication/replication_server_startup/

And then we can add a corresponding links. What do you think?

doc/reference/configuration/configuration_reference.rst
Totktonada: That's correct. Thanks!

doc/reference/configuration/configuration_reference.rst
Totktonada: Agreed to postpone the highlighting till we'll find a way to make it based on a content rather than a line number. The line numbers are unstable if the examples are changing.

doc/reference/configuration/configuration_reference.rst
Totktonada: > Most of the configuration options described in this reference can be applied to a specific instance, replica set, group, or to all instances globally. To do so, you need to define the required option at the [global level].

It still looks like we highlight the global options specifically. Can we, for example, say the following (global -> appropriate)?

> To do so, you need to define the required option at the [appropriate level].

We also highlight the global scope second time in the same paragraph.

> For example, [iproto] and [database] configuration parameters defined at the global level are applied to all instances in the cluster.

I think that it is redundant.

doc/reference/configuration/configuration_reference.rst
Totktonada: Sure. Now, when it is pointed that all the information is about the `failover: off` mode the description looks correct.

doc/reference/configuration/configuration_reference.rst
Totktonada: It is now described in the following way:

* `iproto.advertise.*` -- all the three syntax variants.
* `iproto.advertise.client` -- has constraints, just one syntax variant.

I would prefer something like the following:

* `iproto.advertise.client` -- host:port or unix socket, no user/pass at all.
* `iproto.advertise.{peer,sharding}` -- three syntax variants.

In my head the `client` URI and the `peer`/`sharding` ones are quite different entities.

Moreover, I can just look at the `iproto.advertise.*` description and think that it is applicable to `iproto.advertise.client` without additional syntax constraints.

doc/reference/configuration/configuration_reference.rst
Totktonada: I'm OK with solving it as part of another issue.

doc/reference/configuration/configuration_reference.rst
Totktonada: I still feel it as an improper term. The link (admin-security page) makes a clear difference:

1. Admin console. Text protocol, no password, all the commands are executed from `admin` user. Usually a Unix socket.
2. Iproto socket. Binary protocol, authorization is required (except for `guest` user). Inet or Unix socket.

`tt connect` can work using any of the protocols, but doesn't make the iproto socket being an admin console. It just means that the iproto socket can be used for administration purposes.

If we go to strictier terms, I would also rephrase the following paragraph.

> This address is used for different purposes, for example:
>
> * Communicating between cluster members.
> * Connecting to an instance remotely using an [admin console].
> * Connecting to an instance using [connectors] for different languages.

Look, the address itself is used for two purposes:

* Bind a listening socket for the binary protocol.
* Enrich `iproto.advertise.{peer,sharding}` if it is in form of `user@` or `user:password@` or replace if it is missed.

However, the listening socket itself can be used for all the mentioned purposes:

* Communicate between replicaset peers or cluster members.
* Connect from external services or tools.
* Remote administration using `tt connect`.

doc/reference/configuration/configuration_reference.rst
andreyaksenov: Ok, will add a corresponding TODO item not to forget about this:
image

doc/reference/configuration/configuration_reference.rst
andreyaksenov: Okay, now I got what the `admin console` means. I meant exactly `remote administration`, thanks for your clarification.

Updated the list as follows:
image

doc/reference/configuration/configuration_reference.rst
andreyaksenov: Thanks for your suggestions. To me,`iproto.advertise.{peer,sharding}` looks like a bit of overengineering for docs :) So, I decided to remove the top-level `iproto.advertise.*` and described an URI format for `iproto.advertise.client` and `iproto.advertise.peer` separately. `iproto.advertise.sharding` description just tells now that it has the same format as `iproto.advertise.peer`. 

One move variant is including the content describing the URI format from `iproto.advertise.peer` to `iproto.advertise.sharding` as is. It will look as copy/paste but a reader won't need to jump between descriptions because they are duplicated. If you think this might work better, let's try this variant.

P.S. If there is a new `iproto.advertise.<some_new_param>` option is added with the same URI format, we can consider extracting the URI format description to a separate section.

doc/reference/configuration/configuration_reference.rst
andreyaksenov: > It still looks like we highlight the global options specifically. Can we, for example, say the following (global -> appropriate)?

Sorry, looks like I fixed it right after you started reviewing it, now it looks as follows:
```
To do so, you need to define the required option at the 
:ref:`specified level <configuration_scopes>`.
For example, :ref:`iproto <configuration_reference_iproto>` and 
:ref:`database <configuration_reference_database>` configuration 
parameters defined at the global level are applied to all instances in the cluster.
```

> For example, [iproto] and [database] configuration parameters defined at the global level are applied to all instances in the cluster.

Removed. Also, removed the `Configuration scopes` heading as it now also looks redundant to me and distracts from the section names on the right Table of Contents.

doc/reference/configuration/configuration_reference.rst
Totktonada: Thanks! The current variant seems nice.

doc/reference/configuration/configuration_reference.rst
Totktonada: Looks OK, thanks!

Totktonada (approved):

doc/code_snippets/snippets/config/tt.yaml
```diff
@@ -1,63 +1,54 @@
...
 
-    # The maximum number of days to retain old log files.
-    log_maxage: 8
+  # Path to directory that stores all applications.
+  # The directory can also contain symbolic links to applications.
+  instances_enabled: instances.enabled
 
-    # The maximum number of old log files to retain.
-    log_maxbackups: 10
+  # Tarantoolctl artifacts layout compatibility: if set to true tt will not create application
```
xuniq, line 19:
```suggestion
  # Tarantoolctl artifacts layout compatibility: if set to true, tt will not create application
```
same for 
* doc/code_snippets/snippets/replication/tt.yaml
* doc/code_snippets/snippets/sharding/tt.yaml

doc/code_snippets/snippets/config/tt.yaml
```diff
@@ -1,63 +1,54 @@
...
+  # sub-directories for control socket, pid files, log files, etc.. Data files (wal, vinyl,
+  # snap) and multi-instance applications are not affected by this option.
+  tarantoolctl_layout: false
 
-    # Restart instance on failure.
-    restart_on_failure: false
+app:
+  # Directory that stores various instance runtime
+  # artifacts like console socket, PID file, etc.
```
xuniq, line 26:
```suggestion
  # artifacts such as console socket, PID file, etc.
```
same for 
* doc/code_snippets/snippets/replication/tt.yaml
* doc/code_snippets/snippets/sharding/tt.yaml

doc/reference/configuration/configuration_reference.rst
```diff
@@ -214,13 +239,1030 @@ This section describes options related to :ref:`storing configuration in etcd <c
...
+
+
+
+
+..  _configuration_reference_database:
+
+database
+--------
+
+The ``database`` section defines database-specific configuration parameters, such as an instance's read-write mode, transaction isolation level, and so on.
```
xuniq:
```suggestion
The ``database`` section defines database-specific configuration parameters, such as an instance's read-write mode or transaction isolation level.
```

doc/reference/configuration/configuration_reference.rst
```diff
@@ -214,13 +239,1030 @@ This section describes options related to :ref:`storing configuration in etcd <c
...
+
+    An instance's operating mode.
+    This option is in effect if :ref:`replication.failover <configuration_reference_replication_failover>` is set to ``off``.
+
+    The following modes are available:
+
+    -   ``rw``: an instance is in read-write mode.
+    -   ``ro``: an instance is in read-only mode.
+
+    If this option is not specified explicitly, its default value depends on the number of instances in a replica set. For a single instance, the ``rw`` mode is used, while for multiple instances, the ``ro`` mode is used.
```
xuniq: Suggestion: _If not specified explicitly, the default value depends on the number of instances in a replica set._

doc/reference/configuration/configuration_reference.rst
```diff
@@ -214,13 +239,1030 @@ This section describes options related to :ref:`storing configuration in etcd <c
...
+    | Default: ``best-effort``
+    | Possible values: ``best-effort``, ``read-committed``, ``read-confirmed``
+    | Environment variable: TT_DATABASE_TXN_ISOLATION
+
+
+.. _configuration_reference_database_txn_timeout:
+
+.. confval:: database.txn_timeout
+
+    A timeout (in seconds) after which the transaction is rolled back.
```
xuniq, line 489:
```suggestion
    A timeout in seconds after which the transaction is rolled back.
```

doc/reference/configuration/configuration_reference.rst
```diff
@@ -214,13 +239,1030 @@ This section describes options related to :ref:`storing configuration in etcd <c
...
+-   :ref:`iproto.threads <configuration_reference_iproto_threads>`
+
+
+.. _configuration_reference_iproto_advertise_client:
+
+.. confval:: iproto.advertise.client
+
+    An URI used to advertise the current instance to clients.
+
+    The ``iproto.advertise.client`` option accepts an URI in one of the following formats:
```
xuniq:
```suggestion
    The ``iproto.advertise.client`` option accepts an URI in the following formats:
```

doc/reference/configuration/configuration_reference.rst
```diff
@@ -214,13 +239,1030 @@ This section describes options related to :ref:`storing configuration in etcd <c
...
+    | Type: string
+    | Default: :ref:`box.NULL <box-null>`
+    | Environment variable: TT_IPROTO_LISTEN
+
+
+.. _configuration_reference_iproto_net_msg_max:
+
+.. confval:: iproto.net_msg_max
+
+    To handle messages, Tarantool allocates fibers.
```
xuniq: Link to `fiber` definition?

doc/reference/configuration/configuration_reference.rst
```diff
@@ -214,13 +239,1030 @@ This section describes options related to :ref:`storing configuration in etcd <c
...
+    To prevent fiber overhead from affecting the whole system,
+    Tarantool restricts how many messages the fibers handle,
+    so that some pending requests are blocked.
+
+    -   On powerful systems, increase ``net_msg_max`` and the scheduler
```
xuniq:
```suggestion
    -   On powerful systems, increase ``net_msg_max``, and the scheduler
```

doc/reference/configuration/configuration_reference.rst
```diff
@@ -214,13 +239,1030 @@ This section describes options related to :ref:`storing configuration in etcd <c
...
+        starts processing pending requests immediately.
+
+    -   On weaker systems, decrease ``net_msg_max`` and the overhead
```
xuniq:
```suggestion
    -   On weaker systems, decrease ``net_msg_max``, and the overhead
```

doc/reference/configuration/configuration_reference.rst
```diff
@@ -214,13 +239,1030 @@ This section describes options related to :ref:`storing configuration in etcd <c
...
+        may decrease although this may take some time because the
+        scheduler must wait until already-running requests finish.
```
xuniq: Consider splitting this long sentence in two separate sentences

doc/reference/configuration/configuration_reference.rst
```diff
@@ -214,13 +239,1030 @@ This section describes options related to :ref:`storing configuration in etcd <c
...
+
+.. confval:: replication.election_mode
+
+    A role of a replica set node in the :ref:`leader election process <repl_leader_elect>`.
+
+    The possible values are:
+
+    *   ``off``
+
+        A node doesn't participate in the election activities.
```
xuniq:
```suggestion
    *   ``off``:  a node doesn't participate in the election activities.
```

Same for other possible values below

doc/code_snippets/snippets/config/tt.yaml
andreyaksenov: Unfortunately, cannot fix this as `tt.yaml` is generated automatically using `tt init`. So, this is a reminder that we need to cteate a process for reviewing user-facing texts written by the tt development team.

doc/code_snippets/snippets/config/tt.yaml
andreyaksenov: The same as above

doc/reference/configuration/configuration_reference.rst
andreyaksenov: Much better, thanks

doc/reference/configuration/configuration_reference.rst
andreyaksenov: If you don't mind, I'd keep units in round brackets to emphasise this information a bit.

xuniq (approved):

PR #3890, opened 2023-12-01
3.0 configuration: replication administration

### LINKED ISSUE #3879: [Config] Update the 'Replication administration' section to using a new config
**Product:** Tarantool
**Since:** 3.0
**Root document:** 
- https://www.tarantool.io/en/doc/latest/book/admin/replication/
- https://www.tarantool.io/en/doc/latest/book/admin/disaster_recovery/
- https://www.tarantool.io/en/doc/latest/book/admin/troubleshoot/#problem-master-master-replication-is-stopped

**SME:** @ Totktonada

## Details

Take into account new replication tutorials:
- https://github.com/tarantool/doc/issues/3654
- https://github.com/tarantool/doc/issues/3655
- https://github.com/tarantool/doc/issues/3656

andreyaksenov:
Updated topics related replication administration. Looks like all these topics might be structured better to avoid duplication. Will create a separate issue for this task.

Replication administration:
- [Monitoring a replica set](https://docs.d.tarantool.io/en/doc/3.0-config-replication-administration/book/admin/replication/repl_monitoring/) (includes new diagrams for `box.info.replication`)
- [Recovering from a degraded state](https://docs.d.tarantool.io/en/doc/3.0-config-replication-administration/book/admin/replication/repl_recover/)
- [Reseeding a replica](https://docs.d.tarantool.io/en/doc/3.0-config-replication-administration/book/admin/replication/repl_reseed/)
- [Resolving replication conflicts](https://docs.d.tarantool.io/en/doc/3.0-config-replication-administration/book/admin/replication/repl_problem_solving/) (added a link to the `Resolving replication conflicts` section at the end of the `Preventing duplicate insert` section)

Misc:
- [Disaster recovery](https://docs.d.tarantool.io/en/doc/3.0-config-replication-administration/book/admin/disaster_recovery/)
- [Troubleshooting: Master-master replication is stopped](https://docs.d.tarantool.io/en/doc/3.0-config-replication-administration/book/admin/troubleshoot/#problem-master-master-replication-is-stopped)

Added diagrams to replication tutorials:
- https://docs.d.tarantool.io/en/doc/3.0-config-replication-administration/how-to/replication/repl_bootstrap_auto/#testing-automated-failover (step 3)
- https://docs.d.tarantool.io/en/doc/3.0-config-replication-administration/how-to/replication/repl_bootstrap_master_master/#inserting-conflicting-records (step 6)
- https://docs.d.tarantool.io/en/doc/3.0-config-replication-administration/how-to/replication/repl_bootstrap_master_master/#restarting-replication (step 1)


doc/book/admin/disaster_recovery.rst
```diff
@@ -1,126 +1,162 @@
...
+2.  Configure a new replica set leader using the :ref:`<replicaset_name>.leader <configuration_reference_replicasets_name_leader>` option.
 
-1. Find out the position of the crashed master, as reflected on the new master.
+3.  Reload configuration on all instances using :ref:`config:reload() <config-module>`.
 
-   a. Find out instance UUID from the crashed master :ref:`xlog <internals-wal>`:
+4.  Make sure that a new replica set leader is a master using :ref:`box.info.ro <box_introspection-box_info>`.
 
-      .. code-block:: console
+5.  Remove a crashed master from a replica set as described in :ref:`Removing instances <replication-remove_instances>`.
```
Totktonada: Shouldn't we link `how-to/replication/repl_bootstrap/#removing-an-instance-from-the-cluster-space` instead?

I mean, if the replacement instance from the next (sixth) step is supposed to have the same name as the old one, we need to just drop an entry from `_cluster` and start the instance somewhere again. (It also means that `Adding instances` section is not very relevant.)

So, the minimal working steps instead of these 5-6 ones are the following:

5. Remove the instance from the _cluster space.
6. Start the instance again on a spare host.

doc/book/admin/disaster_recovery.rst
```diff
@@ -1,126 +1,162 @@
...
-   (effective master), so it begins to catch up with the new master’s state.
-   The new instance should have :ref:`box.cfg.read_only <cfg_basic-read_only>`
-   parameter set to *true*.
+1.  Ensure the master is stopped.
+    For example, log in to the master machine and use ``tt stop``.
 
-You lose the few transactions in the master
-:ref:`write ahead log file <index-box_persistence>`, which it may have not
-transferred to the replica before crash. If you were able to salvage the master
-.xlog file, you may be able to recover these. In order to do it:
```
Totktonada, line 44: The note about losing the transactions is important.

I also like the recipe how to repair it.

(Well, I don't like that it is not very user friendly, but it is a way to solve the problem.)

Totktonada: I see now, it is described later. Maybe a 'see also' note would be helpful.

doc/book/admin/disaster_recovery.rst
```diff
@@ -1,126 +1,162 @@
...
+    Learn more from :ref:`Adding and removing instances <replication-master-master-add-remove-instances>`.
 
-2. Follow the same steps as in the
-:ref:`master-replica <admin-disaster_recovery-master_replica>` recovery scenario
-to create a new master and salvage lost data.
 
 .. _admin-disaster_recovery-data_loss:
 
---------------------------------------------------------------------------------
 Data loss
```
Totktonada: It is confusing to have two 'data loss' sections on different levels.

doc/book/admin/disaster_recovery.rst
```diff
@@ -1,126 +1,162 @@
...
+
+**Problem:** Data was deleted at one master and this data loss was propagated to the other node (master or replica).
+
+**Actions:**
+
+1.  Put all nodes in read-only mode.
+    Depending on the :ref:`replication.failover <configuration_reference_replication_failover>` mode, this can be done as follows:
+
+    -   ``manual``: change a replica set leader to ``null``.
+    -   ``election``: switch from the ``election`` failover mode to ``manual`` and change a replica set leader to ``null``.
```
Totktonada: I wouldn't change the `replication.failover` mode if there is a way solve the problem within the mode. Here we can keep `replication.failover: election` and set `replication.election_mode` to `voter` or `off`.

See details [here](https://github.com/tarantool/tarantool/blob/f6d6e4baf0dbef3333d37653b253f3811b3b36f2/src/box/lua/config/applier/box_cfg.lua#L417-L452).

doc/book/admin/disaster_recovery.rst
andreyaksenov: Thanks for the suggestion. I believe that having this info in a separate subsection (compared to the description in old docs) improves the visibility of this info. Not sure that we should add separate notes inside the `Master crash: manual failover` and `Master crash: automated failover` subsections because the `Data loss` subsection is already visible in the Table of contents at the right.
image

doc/book/admin/disaster_recovery.rst
andreyaksenov: Renamed to `Master-replica/master-master: data loss` for consistency. Still not the best variant but I feel that the entire `Administration` section will be restructured gradually when migrating other topics to a new config.

image

doc/book/admin/disaster_recovery.rst
andreyaksenov: Thanks, fixed.
image

doc/book/admin/disaster_recovery.rst
andreyaksenov: Thanks, fixed:

image

doc/book/admin/replication/repl_monitoring.rst
```diff
@@ -68,15 +47,10 @@ The primary indicators of replication health are:
...
     machines, do not be surprised if it’s negative: a time drift may lead to the
     remote master clock being consistently behind the local instance's clock.
 
-    For multi-master configurations, ``lag`` is the maximal lag.
-
-For better understanding, see the following diagram illustrating the ``upstream`` and ``downstream`` connections within the replica set of three instances:
-
-..  image:: /concepts/replication/images/replication.svg
-    :align: left
+    For a :ref:`master-master <replication-bootstrap-master-master>` configuration, ``lag`` is the maximal lag.
```
sergepetrenko: I don't understand why this sentence was here. The semantics of `box.info.replication[*].upstream.lag` don't change depending on how many masters there are. It shows a lag of this particular connection always.

In case of master-master, the user should pay attention to all the upstream lags

doc/book/admin/replication/repl_recover.rst
```diff
@@ -1,107 +1,53 @@
...
+        2023-12-04 13:16:59.690 [16615] main/112/applier/replicator@127.0.0.1:3303 I> RAFT: vote for 2, follow
+        2023-12-04 13:16:59.691 [16615] main/119/raft_worker I> RAFT: persisted state {term: 3}
+        2023-12-04 13:16:59.691 [16615] main/119/raft_worker I> RAFT: persisted state {term: 3, vote: 2}
+        2023-12-04 13:16:59.691 [16615] main/112/applier/replicator@127.0.0.1:3303 I> RAFT: message {term: 3, vote: 2, leader: 2, state: leader} from 2
+        2023-12-04 13:16:59.691 [16615] main/112/applier/replicator@127.0.0.1:3303 I> RAFT: vote request is skipped - this is a notification about a vote for a third node, not a request
+        2023-12-04 13:16:59.691 [16615] main/112/applier/replicator@127.0.0.1:3303 I> RAFT: leader is 2, follow
+
+
+
+The master's status is reported as ``disconnected`` when executing :ref:`box.info.replication <replication-monitoring>` on a replica:
```
sergepetrenko: Might be `disconnected` or `stopped`.

doc/book/admin/replication/repl_reseed.rst
```diff
@@ -4,25 +4,19 @@
...
-   The replica will then catch up with the master by retrieving all the master’s
-   tuples.
+3.  Start the replica using the :ref:`tt start <tt-start>` command.
+    The replica should catch up with the master by retrieving all the master's tuples.
 
-.. NOTE::
-
-   Remember that this procedure works only if the master’s WAL files are
-   present.
+4.  (Optional) If you're reseeding a replica after a replication conflict, you also need to :ref:`restart replication <replication-master-master-resolve-conflict>`.
```
sergepetrenko, line 22: You shouldn't restart replication on the replica which was just reseeded. The replication should be restarted on master, because it won't automatically reconnect to the reseeded replica.

doc/book/admin/replication/repl_problem_solving.rst
```diff
@@ -62,15 +59,16 @@ created. The resulting snippet would be the following:
...
 
-****************************
 Preventing duplicate insert
-****************************
+---------------------------
 
 .. _replication-replication_stops:
 
-**Case 2:** In a replica set of two masters, suppose master #1 tries to
+In a replica set of two masters, suppose master #1 tries to
```
sergepetrenko: Here we should say that both masters try to insert data by the same key.

doc/book/admin/replication/repl_problem_solving.rst
```diff
@@ -150,20 +148,14 @@ report), because the downstream has encountered the same error:
...
-    tarantool> box.cfg{replication={}}
-    tarantool> box.cfg{replication=original_value}
+To learn how to resolve a replication conflict by reseeding a replica, see :ref:`Resolving replication conflicts <replication-master-master-resolve-conflicts>`.
 
-... the faulty row in the write-ahead-log files is skipped.
 
 .. _replication-runs_out_of_sync:
 
-**Solution #1: replication runs out of sync**
+Solution 1: replication runs out of sync
```
sergepetrenko: I don't know why it was called "solution", it looks like another problem statement

doc/book/admin/replication/repl_problem_solving.rst
```diff
@@ -205,7 +198,10 @@ Consider for example the following command:
...
 the update is applied on the other masters.
 
-**Solution #3: trigger usage**
+.. _replication_trigger_usage:
+
+Solution 3: trigger usage
+~~~~~~~~~~~~~~~~~~~~~~~~~
 
 The logic and the snippet setting a trigger will be the same here as in case 1.
 But the trigger function will differ:
```
sergepetrenko: Let's mention here that the trigger below assumes that tuple has a timestamp in the second field

doc/book/admin/disaster_recovery.rst
```diff
@@ -1,126 +1,162 @@
...
+2.  Configure a new replica set leader using the :ref:`<replicaset_name>.leader <configuration_reference_replicasets_name_leader>` option.
 
-1. Find out the position of the crashed master, as reflected on the new master.
+3.  Reload configuration on all instances using :ref:`config:reload() <config-module>`.
 
-   a. Find out instance UUID from the crashed master :ref:`xlog <internals-wal>`:
+4.  Make sure that a new replica set leader is a master using :ref:`box.info.ro <box_introspection-box_info>`.
 
-      .. code-block:: console
+5.  On a new master, :ref:`remove a crashed instance from the '_cluster' space <replication-remove_instances-remove_cluster>`.
```
sergepetrenko, line 38: This means the user will have to reseed the old master from scratch next time it starts. It seems it's not needed when the master simply crashes. Yes, you have to promote the new master, but there is no need to remove the old one from _cluster.

sergepetrenko: This only makes sense when the user moves the master from one physical node to another. But even then, with the instance names, the master on a new physical node will occupy the same _cluster slot (it will find it by name).

It seems the user only has to delete nodes from _cluster when he wants to remove them permanently.

doc/book/admin/disaster_recovery.rst
```diff
@@ -1,126 +1,162 @@
...
-          ...
+**Actions:**
 
-   b. Play the records to the new master:
+1.  Use ``box.info.election`` to make sure a new master is elected automatically.
 
-      .. code-block:: console
+2.  Remove a crashed master from a replica set.
+
+3.  Set up a replacement for the crashed master on a spare host.
```
sergepetrenko, line 60: Is this tutorial aimed at cases where the physical machine with the old master goes down and we are setting up a replacement for the old master on a new host?
In this case ok, we may leave the part about _cluster deletion.

doc/book/admin/troubleshoot.rst
```diff
@@ -249,22 +249,15 @@ error message like
 
 **Solution**
 
-Restart replication with the following commands (at each master instance):
+Restart replication at each master instance.
```
sergepetrenko: The solution described here won't help in case of replication conflicts, like it's suggested in **Possible reasons**.
Replication will stop again right after restart. The user should set up conflict resolution triggers before trying to restart replication.

doc/book/admin/replication/repl_reseed.rst
sergepetrenko: Ok, I see this is mentioned in the link.

sergepetrenko (commented):
Thanks for working on this!
Please find my comments below.

doc/book/admin/replication/repl_monitoring.rst
andreyaksenov: thanks, removed this sentence

doc/book/admin/replication/repl_recover.rst
andreyaksenov: Changed from ...
`The master's status is reported as ...`
to ...
`The master's upstream status is reported as ...`

Note that the report above shows the `disconnected` status for `upstream` and `stopped` status for `downstream`.

doc/book/admin/replication/repl_problem_solving.rst
andreyaksenov: Changed the intro to:

`In a replica set of two masters, both of them try to insert data by the same unique key:`

doc/book/admin/replication/repl_problem_solving.rst
andreyaksenov: Removed `Solution` from all subheadings.

doc/book/admin/disaster_recovery.rst
andreyaksenov: Thanks for the detailed clarification, removed this step. Now looks as follows:
```
4.  Make sure that a new replica set leader is a master using :ref:`box.info.ro <box_introspection-box_info>`.

5.  Set up a replacement for the crashed master on a spare host.
```

doc/book/admin/disaster_recovery.rst
andreyaksenov: Sorry, I'm a bit confused :) Please help clarify this question. Here is the previous comment:

> This only makes sense when the user moves the master from one physical node to another. But even then, with the instance names, the master on a new physical node will occupy the same _cluster slot (it will find it by name).

Don't know what the author of the initial topic kept in mind but let's assume that the old master goes down and we are setting up a replacement for the old master on a new host (as you mentioned above). Am I correct that we should remove it from  `_cluster` after all?

sergepetrenko: Yes, we should remove the old host from _cluster in this case. Let's not change the logic of the previous author.

doc/book/admin/troubleshoot.rst
andreyaksenov: Thanks, updated as follows:

image

sergepetrenko: Nice, thanks!

sergepetrenko (approved):
Thanks for the patch!

doc/book/admin/replication/repl_reseed.rst
```diff
@@ -4,25 +4,19 @@
 Reseeding a replica
 ================================================================================
 
-If any of a replica's .xlog/.snap/.run files are corrupted or deleted, you can
-"re-seed" the replica:
+If any of a replica's write-ahead log or snapshot files are corrupted or deleted, you can "re-seed" the replica.
```
xuniq:
```suggestion
If any of a replica's write-ahead log or snapshot files are corrupted or deleted, you can reseed the replica.
```

doc/book/admin/replication/repl_recover.rst
```diff
@@ -1,107 +1,53 @@
...
+        2023-12-04 13:16:59.691 [16615] main/112/applier/replicator@127.0.0.1:3303 I> RAFT: vote request is skipped - this is a notification about a vote for a third node, not a request
+        2023-12-04 13:16:59.691 [16615] main/112/applier/replicator@127.0.0.1:3303 I> RAFT: leader is 2, follow
+
+
+
+The master's upstream status is reported as ``disconnected`` when executing :ref:`box.info.replication <replication-monitoring>` on a replica:
+
+..  include:: /how-to/replication/repl_bootstrap_auto.rst
+    :start-after: box_info_replication_auto_leader_disconnected_start
+    :end-before: box_info_replication_auto_leader_disconnected_end
```
xuniq, line 48: I like the way how a small part of the page is reused here - with `:start-after:` and `:end-before:`. Will definitely use this in future.

xuniq (approved):

PR #3909, opened 2023-12-07
3.0 new bootstrap strategy - "supervised"

### LINKED ISSUE #3460: new bootstrap strategy - "supervised"
**Related dev. issue(s):** https://github.com/tarantool/tarantool/issues/8509

**Product:** Tarantool
**Since:** 3.0.0-alpha3
**Root document:** https://www.tarantool.io/en/doc/latest/reference/configuration/#cfg-replication-bootstrap-strategy
**SME:** @ sergepetrenko

# Details

The `bootstrap_strategy` configuration option may now be set to
"supervised".

This strategy works as follows:
When bootstrapping a new replicaset, the nodes do not choose a bootstrap
leader automatically and instead wait for it to be appointed by the
user. The configuration will fail if no bootstrap leader is appointed
during a `replication_connect_timeout`.

In order to appoint a bootstrap leader, the user has to issue
`box.ctl.make_bootstrap_leader()` call on the desired node. This must be
done during the initial `box.cfg()` call.

Possible ways to achieve this are:

1. In interactive mode:
```lua
fiber.create(box.cfg, desired_config)
box.ctl.make_bootstrap_leader()
```
2. Via an init script:
```lua
-- Init script:
console.listen(admin_port)
box.cfg(desired_config)
-- User console:
tarantoolctl enter admin port
> box.ctl.make_bootstrap_leader()
```

When joining a new replica with `bootstrap_strategy` = "supervised" to
an existing replica set, the replica will not choose the bootstrap
leader automatically, but will instead join to the node on which
`box.ctl.make_bootstrap_leader()` was issued last. In case such a node
isn't found after a `replication_connect_timeout`, the configuration
fails.

Requested by @sergepetrenko in https://github.com/tarantool/tarantool/commit/c419cab7d4d66d7fab5ea424d0807cd3bbbf0b51

andreyaksenov:
Fixes https://github.com/tarantool/doc/issues/3460.

Staging:

- [replication.bootstrap_strategy](https://docs.d.tarantool.io/en/doc/3.0-bootstrap-strategy/reference/configuration/configuration_reference/#configuration-reference-replication-bootstrap-strategy)
- [box.ctl.make_bootstrap_leader()](https://docs.d.tarantool.io/en/doc/3.0-bootstrap-strategy/reference/reference_lua/box_ctl/make_bootstrap_leader/)
- [Submodule box.ctl](https://docs.d.tarantool.io/en/doc/3.0-bootstrap-strategy/reference/reference_lua/box_ctl/)

### DIFF
--- a/doc/reference/reference_lua/box_ctl/make_bootstrap_leader.rst
+++ b/doc/reference/reference_lua/box_ctl/make_bootstrap_leader.rst
@@ -1,0 +1,20 @@
+.. _box_ctl-make_bootstrap_leader:
+
+box.ctl.make_bootstrap_leader()
+===============================
+
+.. function:: make_bootstrap_leader()
+
+    **Since:** :doc:`3.0.0 </release/3.0.0>`.
+
+    Make the instance a bootstrap leader of a replica set.
+
+    To be able to make the instance a bootstrap leader manually, the :ref:`replication.bootstrap_strategy <configuration_reference_replication_bootstrap_strategy>` configuration option should be set to ``supervised``.
+    In this case, the instances do not choose a bootstrap leader automatically but wait for it to be appointed manually.
+    Configuration fails if no bootstrap leader is appointed during a :ref:`replication.connect_timeout <configuration_reference_replication_connect_timeout>`.
+
+    ..  NOTE::
+
+        When a new instance joins a replica set configured with the ``supervised`` bootstrap strategy,
+        this instance doesn't choose the bootstrap leader automatically but joins to the instance on which
+        ``box.ctl.make_bootstrap_leader()`` was executed last time.
--- a/doc/code_snippets/snippets/replication/instances.enabled/bootstrap_strategy/README.md
+++ b/doc/code_snippets/snippets/replication/instances.enabled/bootstrap_strategy/README.md
@@ -1,0 +1,11 @@
+# replication.bootstrap_strategy
+
+A sample application demonstrating how to use the specified instance to bootstrap a replica set.
+
+## Running
+
+To start all instances, execute the following command in the [replication](../../../replication) directory:
+
+```console
+$ tt start bootstrap_strategy
+```
--- a/doc/reference/configuration/configuration_reference.rst
+++ b/doc/reference/configuration/configuration_reference.rst
@@ -933,9 +933,7 @@ The ``replication`` section defines configuration parameters related to :ref:`re
     *   ``config``: use the specified node to bootstrap a replica set.
         To specify the bootstrap leader, use the :ref:`<replicaset_name>.bootstrap_leader <configuration_reference_replicasets_name_bootstrap_leader>` option.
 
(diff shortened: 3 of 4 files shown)


doc/reference/reference_lua/box_ctl/make_bootstrap_leader.rst
```diff
@@ -0,0 +1,20 @@
+.. _box_ctl-make_bootstrap_leader:
+
+box.ctl.make_bootstrap_leader()
+===============================
+
+.. function:: make_bootstrap_leader()
+
+    **Since:** :doc:`3.0.0 </release/3.0.0>`.
+
+    Make the instance a bootstrap leader of a replica set.
```
sergepetrenko, line 10: I'm not sure how the new configuration works, but it's worth mentioning that `box.ctl.make_bootstrap_leader()` should be called after `box.cfg()` call is issued. I suppose this is the only way to do this with the new configuration anyway.
The node should be listening on a console `box.cfg.listen` or `require'console'.listen()` while running `box.cfg` in order for the user to be able to send the `box.ctl.make_bootstrap_leader()` command. Otherwise there will be a deadlock: `box.cfg()` on the desired bootstrap leader doesn't end until `box.ctl.make_bootstrap_leader()` is issued.

@Totktonada, can you confirm that everything works fine in this case?

Totktonada: I performed several experiments around `supervised` bootstrap mode with config in September. I found no easy way to use it.

Config starts the admin interface after starting the database (`console.listen()` goes after `box.cfg()`). So, we have no way to give a server commands before the bootstrap.

Moreover, the server can't perform some user specified code before `box.cfg()` to, say, pool some bootstrap coordinator or listen on a separate command socket.

This looks like a true deadlock.

I would discuss f2f what is better to do with `replication.bootstrap_strategy: supervised` in the declarative configuration.

Totktonada: We discussed it with Sergey and agreed that the admin console should be started before `box.cfg()`. I've added a subtask to https://github.com/tarantool/tarantool/issues/8862.

andreyaksenov: So, am I correct that these are internal details for a declarative configuration approach and I shouldn't add anything to the `make_bootstrap_leader()` description?

sergepetrenko: > So, am I correct that these are internal details for a declarative configuration approach and I shouldn't add anything to the make_bootstrap_leader() description?

Current description looks good to me, I think we can continue with this PR without waiting for updates in declarative config internals, if @Totktonada doesn't object.

Totktonada: I'm OK.

sergepetrenko (approved):

xuniq (approved):

PR #3985, opened 2024-01-10
Connection settings

### LINKED ISSUE #3665: [Config] Update the 'Security hardening guide' to using a new config
**Previous step:** https://github.com/tarantool/doc/issues/3945
**Feedback:** https://github.com/tarantool/doc/issues/3959

**Product:** Tarantool
**Since:** 3.0
**Root document:** https://www.tarantool.io/en/enterprise_doc/security/
**SME:** @ Totktonada

# Details

Currently, `box.cfg` examples are used.

P.S. Some environment variables should also be updated ([comparison table](https://docs.google.com/spreadsheets/d/1_yEK9N_hBVWIga9mOk7zZzA6Zd3mAq7e5KZsnrbgj30/edit#gid=0)).

# Main steps

- [ ] Create the `Connections` topic in the [Configuration](https://www.tarantool.io/en/doc/latest/concepts/configuration/) section (see also [Connections and Authentication](https://www.postgresql.org/docs/current/runtime-config-connection.html)). The structure of this topic might be:
   - Overview (iproto, connectors, replication, sharding)
   - Listen URI (ip+port, unix socket)
   - Advertise URI (domain names can be used)
   - SSL (`Traffic encryption` will be moved here, mention certificate reloading)
- [ ] Create the `Authentication` topic in the `Configuration` section and move the info from the [Access control](https://www.tarantool.io/en/doc/latest/enterprise/security/#access-control) section.
    - Add corresponding options to reference.

andreyaksenov:
## Main changes

1) Created a new `Connections` topic that describes ``iproto`` settings:
    https://docs.d.tarantool.io/en/doc/3.0-connection-settings/concepts/configuration/configuration_connections/

2) Updated some descriptions in ``iproto`` reference:
     https://docs.d.tarantool.io/en/doc/3.0-connection-settings/reference/configuration/configuration_reference/#iproto

3) Removed the old `Traffic encryption` section from `Security hardening guide`:
     https://docs.d.tarantool.io/en/doc/3.0-connection-settings/enterprise/security/
     Now this topic doesn't include how-to/reference information and describes only general recommendations. We can think about merging `Security hardening guide` with the `Security audit` topic at some point.

4) Removed duplicated information from the main `Configuration` topic and added the link to the new `Connections` topic:
     https://docs.d.tarantool.io/en/doc/3.0-connection-settings/concepts/configuration/#connection-settings
    > Note that the whole `Configuration options overview` section will be removed when corresponding topics for specific subsystems are ready (missing topics: `Access control`, `In-memory storage`, `Persistence`). The `Configuration` topic will be renamed to `Configuration overview`. So, the idea of future structure is smth like this:
    > - Configuration
    >   - Configuration overview
    >     - Storing configuration in etcd
    >     - Storing configuration in Tarantool
    >     - ...
    >   - In-memory storage
    >   - Persistence
    >   - Replication
    >   - Sharding
    >   - Connections
    >   - Access control
    >   - Authentication
    >   - Logging
    >   - ...

5) Added several runnable samples:
    - Single-instance ``iproto.listen`` samples.
    - Advertise URI sample (replica set, uses two listen URIs and one advertise URI).
    - SSL examples (without and with CA).

## Fixed links in other documents

1) Updated a link in the Getting Started guide:
    https://docs.d.tarantool.io/en/doc/3.0-connection-settings/how-to/getting_started_db/#connecting-remotely

2) Updated links in TCM docs:
     https://docs.d.tarantool.io/en/doc/3.0-connection-settings/reference/tooling/tcm/tcm_configuration_reference/#confval-storage.tarantool.ssl.key-file


doc/concepts/configuration/configuration_connections.rst
```diff
@@ -0,0 +1,292 @@
...
+SSL parameters
+~~~~~~~~~~~~~~
+
+You can enable :ref:`traffic encryption <configuration_connections_ssl>` for a connection using the ``params`` section of the specified URI:
+
+..  literalinclude:: /code_snippets/snippets/replication/instances.enabled/ssl_without_ca/config.yaml
+    :language: yaml
+    :start-at: instance001:
+    :end-before: instance002:
+    :dedent:
```
Totktonada: I would highlight that it is available only in Tarantool Enterprise Edition.

doc/concepts/configuration/configuration_connections.rst
```diff
@@ -0,0 +1,292 @@
...
+
+Advertise settings (:ref:`iproto.advertise.* <configuration_reference_iproto_advertise>`) let other cluster members or clients know how to connect to the current Tarantool instance.
+These settings might include the credentials required to connect to this instance, a URI used to listen for incoming requests, and SSL settings.
+
+If an advertise URI is not specified explicitly, a :ref:`listen URI <configuration_connections_listen_uri>` of this instance is used.
+In this case, you need at least to specify credentials for connecting to this instance.
+
+..  NOTE::
+
+    If several listen URIs are specified, the first one is used as an advertise URI.
```
Totktonada: It is about automatic `replication.peers` construction, but there are no words that put a reader into this context.

Totktonada: Strictly speaking, first one *that is suitable to connect*. It excludes URIs with `0.0.0.0` or `[::0]` IP address and URIs with zero TCP/IP port.

doc/concepts/configuration/configuration_connections.rst
```diff
@@ -0,0 +1,292 @@
...
+            uri: 'server001.example.com:3301'
+
+
+.. _configuration_connections_advertise_uri_ssl:
+
+SSL parameters
+~~~~~~~~~~~~~~
+
+:ref:`SSL parameters <configuration_connections_ssl>` for an advertise URI should be set only if this :ref:`advertise URI is specified explicitly <configuration_connections_advertise_explicitly>`.
+Otherwise, SSL parameters of a listen URI are used and no additional configuration is required.
```
Totktonada: These sentences are a bit complicated due to several negations. AFAIU, you want to say that if `uri` or `params` are set in an advertise URI, `login`, `password`, `uri`, `params` are not borrowed from the listen URI. Is it correct?

doc/concepts/configuration/configuration_connections.rst
```diff
@@ -0,0 +1,292 @@
...
+
+Configuring an advertise URI's SSL options depends on whether a trusted certificate authorities (CA) file is set or not.
```
Totktonada: > <...> whether a trusted certificate authorities (CA) file is set or not.

...in `iproto.listen`?

doc/concepts/configuration/configuration_connections.rst
```diff
@@ -0,0 +1,292 @@
...
+Without the CA file, you only need to set ``iproto_advertise.<peer_or_sharding>.params.transport`` to ``ssl`` as shown below:
```
Totktonada: I'm a bit in doubts about 'with/without the CA file' wording. Can we say something like the following.

> Verification of a client certificate on the server is enabled by setting the `ssl_ca_file` parameter in an `iproto.listen` URI. The verification checks that the client certificate issued by the given certificate authority.
>
> Verification of server's certificate is enabled <...>

After this we can just say, whether server/client certificate verification is enabled instead of technical 'with/without the CA file'.

doc/concepts/configuration/configuration_connections.rst
andreyaksenov: Not exactly. Here I want to emphasise that if `iproto_advertise.<peer_or_sharding>.uri` is specified manually, then SSL settings for this URI can't be taken from `iproto.listen` and should be specified manually.

doc/enterprise/security.rst
```diff
@@ -408,275 +408,7 @@ For more information on logging, see the following:
...
-
-.. _enterprise-iproto-encryption-config:
-
-Configuration
-~~~~~~~~~~~~~
-
-To configure traffic encryption, you need to set the special :ref:`URI parameters <index-uri-several-params>` for a particular connection.
-The parameters can be set for the following ``box.cfg`` options and ``nex.box`` method:
-
-*   :ref:`box.cfg.listen <cfg_basic-listen>` -- on the server side.
```
Totktonada, line 430: Shouldn't we add it into the `box.cfg` options description (and `net.box` too)?

Totktonada (approved):
I have no strict objections here. There are thoughts around the URIs topic in my comments above, but I guess that you may want to move forward in smaller iterations and I'm OK with it.

doc/concepts/configuration/configuration_connections.rst
Totktonada: BTW, this logic is also in effect at creating vshard's configuration.

doc/enterprise/security.rst
andreyaksenov: FYI added missing options to the old box.cfg reference: https://github.com/tarantool/doc/pull/4007

doc/concepts/configuration/configuration_connections.rst
```diff
@@ -0,0 +1,295 @@
+..  _configuration_connections:
+
+Connections
+===========
+
+Configuring connections is required for different purposes, for example:
```
p7nov: This intro sounds like a weak writing to me: "we need to start with something, so here's a random general sentence about connections".
Don't have specific ideas what to start with, but this needs rewriting IMO.

p7nov: Maybe, start with the thought that Tarantool connectivity is complex, i.e. includes multiple aspects: cluster internal, admin tools, language connectors etc. And then proceed to explaining what config is needed for different scenarios.

doc/concepts/configuration/configuration_connections.rst
```diff
@@ -0,0 +1,295 @@
...
+
+-   Communicating between cluster members.
+-   Connecting to cluster members for administration using :ref:`tt <tt-cli>`.
+-   Managing clusters using :ref:`Tarantool Cluster Manager <tcm>`.
+-   Connecting to an instance using the :ref:`net.box <net_box-module>` module or :ref:`connectors <index-box_connectors>` for different languages.
+
+To communicate to and between cluster instances, Tarantool uses a :ref:`binary protocol <box_protocol>` called iproto.
```
p7nov: The meaning of `communicate to and between` is unclear to me.

doc/concepts/configuration/configuration_connections.rst
```diff
@@ -0,0 +1,295 @@
...
+The corresponding :ref:`iproto <configuration_reference_iproto>` section in :ref:`YAML configuration <configuration>` lets you configure various connection settings:
+
+-   One or several URIs used to listen for incoming requests.
```
p7nov: Can we explicitly map these settings to connection types (the first list on this page): which settings are used for which connections?

doc/concepts/configuration/configuration_connections.rst
```diff
@@ -0,0 +1,295 @@
...
+
+Configuring connections is required for different purposes, for example:
+
+-   Communicating between cluster members.
+-   Connecting to cluster members for administration using :ref:`tt <tt-cli>`.
+-   Managing clusters using :ref:`Tarantool Cluster Manager <tcm>`.
+-   Connecting to an instance using the :ref:`net.box <net_box-module>` module or :ref:`connectors <index-box_connectors>` for different languages.
+
+To communicate to and between cluster instances, Tarantool uses a :ref:`binary protocol <box_protocol>` called iproto.
+The corresponding :ref:`iproto <configuration_reference_iproto>` section in :ref:`YAML configuration <configuration>` lets you configure various connection settings:
```
p7nov: Maybe invert this sentence to start with user's task rather than config section?
`The connection settings are defined in the iproto ...`
`To define the connection settings, use the iproto...`

doc/concepts/configuration/configuration_connections.rst
```diff
@@ -0,0 +1,295 @@
...
+-   An URI used advertise an instance to other cluster members.
+-   SSL settings used to secure connections.
+
+
+.. _configuration_connections_listen_uri:
+
+Listen URI
+----------
+
+To configure URIs used to listen for incoming requests, use the :ref:`iproto.listen <configuration_reference_iproto_listen>` configuration option.
```
p7nov, line 28: As a non-expert reader, I don't quite understand from this wording, what requests are we talking about. Are these data requests (`select` from spaces)? Internal cluster requests? Or just any network requests from whoever connects to this URI?

doc/concepts/configuration/configuration_connections.rst
```diff
@@ -0,0 +1,295 @@
...
+
+In this case, this port is used for all IP addresses the server listens on.
+
+
+.. _configuration_connections_listen_ssl:
+
+SSL parameters
+~~~~~~~~~~~~~~
+
+In the Enterprise Edition, you can enable :ref:`traffic encryption <configuration_connections_ssl>` for a connection using the ``params`` section of the specified URI:
```
p7nov: Maybe add EE admonition that we typically use in Enterprise-specific sections?

doc/concepts/configuration/configuration_connections.rst
```diff
@@ -0,0 +1,295 @@
...
+
+An advertise URI (:ref:`iproto.advertise.* <configuration_reference_iproto_advertise>`) lets other cluster members or clients know how to connect to the current Tarantool instance:
+
+-   ``iproto.advertise.peer`` specifies how to advertise the instance to other cluster members.
+-   ``iproto.advertise.sharding`` specifies how to advertise the instance to a router and rebalancer.
+-   ``iproto.advertise.client`` accepts a URI used to advertise the instance to clients.
+
+``iproto_advertise.<peer_or_sharding>`` might include the credentials required to connect to this instance, a URI used to listen for incoming requests, and SSL settings.
+
+If ``iproto_advertise.<peer_or_sharding>.uri`` is not specified explicitly, a :ref:`listen URI <configuration_connections_listen_uri>` of this instance is used.
```
p7nov: Maybe mention somewhere explicitly that `listen` is mandatory and `advertise` is not?

doc/concepts/configuration/configuration_connections.rst
```diff
@@ -0,0 +1,295 @@
...
+In this case, you need at least to specify credentials for connecting to this instance.
+
+
+.. _configuration_connections_advertise_credentials:
+
+Connection credentials
+~~~~~~~~~~~~~~~~~~~~~~
+
+In the example below, the ``iproto.advertise.peer`` option is used to inform other replica set members that the ``replicator`` user should be used to connect to the current instance:
```
p7nov, line 126: Above, these were `cluster members` instead of `replica set members`. Does this need a more detailed explanation? If it's obvious for readers - OK.

doc/concepts/configuration/configuration_connections.rst
```diff
@@ -0,0 +1,295 @@
...
+If required, you can specify an advertise URI explicitly by setting up the :ref:`iproto_advertise.\<peer_or_sharding\>.uri <configuration_reference_iproto_advertise.peer_sharding.uri>` option.
+In the example below, ``iproto.listen`` includes two URIs that can be used to connect to ``instance001`` but only the second one is used to advertise this instance to other replica set peers:
+
+..  literalinclude:: /code_snippets/snippets/replication/instances.enabled/advertise_peer/config.yaml
+    :language: yaml
+    :start-at: instance001:
+    :end-before: instance002:
+    :dedent:
+
+The ``iproto_advertise.<peer_or_sharding>.uri`` option can also accept an FQDN instead of an IP address:
```
p7nov: Can a `listen` address be a domain name? If not, I'd highlight this fact. `Unlike iproto.listen.uri, the iproto.advertise.uri can ..`

p7nov: BTW, there is an underscore in `iproto_advertise` instead of period.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -564,9 +564,8 @@ iproto.advertise.*
 
 .. confval:: iproto.advertise.peer
 
-    A URI used to advertise the current instance to other cluster members.
-
-    The ``iproto.advertise.peer`` option accepts a URI in the format described in :ref:`iproto_advertise.\<peer_or_sharding\>.* <configuration_reference_iproto_advertise.peer_sharding>`.
+    Settings used advertise the current instance to other cluster members.
```
p7nov: to advertise
for advertising

doc/reference/configuration/configuration_reference.rst
```diff
@@ -588,9 +587,8 @@ iproto.advertise.*
 
 .. confval:: iproto.advertise.sharding
 
-    An advertise URI used by a router and rebalancer.
-
-    The ``iproto.advertise.sharding`` option accepts a URI in the format described in :ref:`iproto_advertise.\<peer_or_sharding\>.* <configuration_reference_iproto_advertise.peer_sharding>`.
+    Settings used advertise the current instance to a router and rebalancer.
```
p7nov: to advertise
for advertising

doc/reference/configuration/configuration_reference.rst
```diff
@@ -799,15 +790,8 @@ URI parameters that can be used in the following options:
```
p7nov, line 794: Lines 782-783:

-   :ref:`**iproto_advertise**.\<peer_or_sharding\>.params <configuration_reference_iproto_advertise.peer_sharding.params>` - undescore after `iproto`
-   :ref:`iproto.listen <configuration_reference_iproto_listen>` - no `params` in the end

p7nov: Please check for `iproto_advertise` (with underscore) across the page, there are more occurrences.

p7nov (changes requested):
Please see my thoughts about connections.

Skipped the authentication part as we discussed.

doc/concepts/configuration/configuration_connections.rst
andreyaksenov: > we need to start with something, so here's a random general sentence about connections

totally agree, will think how to rewrite

doc/concepts/configuration/configuration_connections.rst
andreyaksenov: Rephrased in the new intro

doc/concepts/configuration/configuration_connections.rst
andreyaksenov: Not exactly. `iproto.listen` is used for all purposes: communicating between instances, administration using tt/TCM, and so on. Advertise URI is only applied to replication and sharding. In the new intro, I've converted the first list to a regular sentence. Looks like the list with connection settings is more important and the second list shouldn't add additional distraction.

doc/concepts/configuration/configuration_connections.rst
andreyaksenov: Rephrased in a new into

doc/concepts/configuration/configuration_connections.rst
andreyaksenov: Good point. I'm not sure that we should give examples of requests right here. From the intro, it should be clear a high-level idea on why connections are required. There are too many request types that can come from other instances and external systems: internal replication requests, sharding requests, data definition and data manipulation requests, and so on. I feel that concrete info should live in corresponding topics. A user can open the `Binary protocol` section linked in a note to see a tons of request types :)

doc/concepts/configuration/configuration_connections.rst
andreyaksenov: I'd keep it as is - an admonition takes too much attention for such a short section.

doc/concepts/configuration/configuration_connections.rst
andreyaksenov: BTW, the reference already tells that all options in `iproto_advertise.<peer_or_sharding>` are optional:

https://www.tarantool.io/en/doc/latest/reference/configuration/configuration_reference/#iproto-advertise-peer-or-sharding

I hope, `is not specified explicitly` implies that `iproto_advertise.<peer_or_sharding>.uri` is optional.

doc/concepts/configuration/configuration_connections.rst
andreyaksenov: The behavior of `iproto.advertise.peer` is a bit complicated. It advertises the instance to other replica set members. If `iproto.advertise.sharding` is not specified, `iproto.advertise.peer` is used to advertise the instance to a router and rebalancer. So, both are true if different contexts.

doc/concepts/configuration/configuration_connections.rst
andreyaksenov: > BTW, there is an underscore in iproto_advertise instead of period.

thanks, fixed in all places (including the Reference page).

> Can a listen address be a domain name?

I'd not add this as using an FQDN as a listen address sounds a bit strange because it is used on the application level only (while a listen address works on the TCP/IP level)

doc/reference/configuration/configuration_reference.rst
andreyaksenov: thanks, fixed in all places

andreyaksenov: > :ref:iproto.listen <configuration_reference_iproto_listen> - no params in the end

Changed to:

```
iproto.listen.<uri>.params
```

doc/concepts/configuration/configuration_connections.rst
p7nov: I don't mean that we should describe all possible requests, but maybe provide general conceptual info in an explicit human-friendly form.
Like, `this is the basic option. It is required on any instance to enable external connections to it`.
Also, we can add a navigating link to `Advertise URI` by writing a sentence about the relation between the two URI types. Why do we need the two at all?

As an idea: we can expand the intro to explain all key entities in it (listen URI, advertise URI, secure connections) and then leave all subsections as is - just showing the usage examples.

doc/concepts/configuration/configuration_connections.rst
p7nov: > If `iproto.advertise.sharding` is not specified, `iproto.advertise.peer` is used

I believe this should be written explicitly, I couldn't ever guess myself)

doc/concepts/configuration/configuration_connections.rst
andreyaksenov: I hope, the current description should work better - it adds the information about how an advertise URI is used and also tells that SSL settings are optional.

image

doc/concepts/configuration/configuration_connections.rst
andreyaksenov: Added a note to the reference page:

image

p7nov (approved):
👍

PR #4024, opened 2024-01-26
Rework the Access control topic

### LINKED ISSUE #3666: [Config] Document the 'credentials' option

andreyaksenov:
Reworked the [Access control](https://docs.d.tarantool.io/en/doc/config-credentials/book/admin/access_control/) topic:

- Rewrote the part about Access Control concepts. Now concepts and how-to instructions are documented separately.
- Added testable samples that show how to work with users and roles.
- Added a separate how-to section that shows how to grant privileges for specific object types.
- Added info from the [Wiki page](https://github.com/tarantool/tarantool/wiki/Privileges-and-Access-Control-explained) about combinations of different object types and privileges. Note that some low-level details might be missing in the updated topic.
- Included new samples to reference:
   - https://docs.d.tarantool.io/en/doc/config-credentials/reference/reference_lua/box_schema/user_create/
   - ...
   - https://docs.d.tarantool.io/en/doc/config-credentials/reference/reference_lua/box_schema/user_create/
   - https://docs.d.tarantool.io/en/doc/config-credentials/reference/reference_lua/box_space/_user/

Next step: https://github.com/tarantool/doc/pull/4035.

### DIFF
--- a/doc/book/admin/access_control.rst
+++ b/doc/book/admin/access_control.rst
@@ -1,262 +1,769 @@
 .. _authentication:
+.. _access_control:
 
-================================================================================
 Access control
-================================================================================
+==============
 
-This section explains how Tarantool makes it possible for administrators
-to prevent unauthorized access to the database and to certain functions.
+Tarantool enables flexible management of access to various database resources.
+The main concepts of Tarantool access control system are as follows:
 
-Briefly:
+*   A *user* is a person or program that interacts with a Tarantool instance.
+*   An *object* is an entity to which access can be granted, for example, spaces, indexes, or functions.
+*   *Privileges* allow a user to perform certain operations on specific objects, for example, creating spaces, reading or updating data.
+*   A *role* is a named collection of privileges that can be granted to a user.
 
-* There is a method to guarantee with password checks that users really are
-  who they say they are (“authentication”).
 
-* There is a :ref:`_user <box_space-user>` system space, where usernames and
-  password-hashes are stored.
 
-* There are functions for saying that certain users are allowed to do certain
-  things (“privileges”).
+.. _access_control_overview:
 
-* There is a :ref:`_priv <box_space-priv>` system space, where privileges are
-  stored. Whenever a user tries to do an operation, there is a check whether
-  the user has the privilege to do the operation (“access control”).
-
-Details follow.
+Overview
+--------
 
 .. _authentication-users:
+.. _access_control_concepts_users:
 
---------------------------------------------------------------------------------
 Users
---------------------------------------------------------------------------------
+~~~~~
 
-There is a **current user** for any program working with Tarantool,
-local or remote.
-If a remote connection is using a :ref:`binary port <admin-security>`,
-the current user, by default, is '**guest**'.
-If the connection is using an :ref:`admin-console port <admin-security>`,
-the current user is '**admin**'.
-When executing a :ref:`Lua initialization script <index-init_label>`,
-the current user is also ‘**admin**’.
+A user is a person or program that interacts with a Tarantool instance.
+There might be different types of users, for example:
 
-The current user name can be found with
-:doc:`/reference/reference_lua/box_session/user`.
+*   A database administrator is responsible for the overall management and administration of a database.
+    An administrator can create other users and grant them specified privileges.
+*   A regular user has limited access to certain data and stored functions. For example, such users can be used to let :ref:`connectors <index-box_connectors>` communicate to a database.
+*   Users used in communications between Tarantool instances. For example, such users can be created to maintain replication and sharding in a Tarantool cluster.
 
-The current user can be changed:
+There are two built-in users in Tarantool:
 
-* For a binary port connection -- with the
-  :ref:`AUTH protocol command <box_protocol-iproto_protocol>`, supported
-  by most clients;
+*   'admin' is a user with all available administrative permissions.
+    If the connection uses an :ref:`admin-console port <admin-security>`, the current user is 'admin'.
+    For example, 'admin' is used when connecting to the instance using :ref:`tt connect <tt-connect>` locally using the instance name:
 
-* For an admin-console connection and in a Lua initialization script --
-  with :doc:`/reference/reference_lua/box_session/su`;
+    ..  code-block:: console
 
-* For a binary-port connection invoking a stored function with the CALL command --
-  if the :doc:`SETUID </reference/reference_lua/box_schema/func_create>`
-  property is enabled for the function,
-  Tarantool temporarily replaces the current user with the
-  function’s creator, with all the creator's privileges, during function execution.
+        $ tt connect app:instance001
+
+    ..  NOTE::
+
+        To allow remote :ref:`binary port <admin-security>` connections using the 'admin' user, you need to :ref:`set a password <access_control_user_changing_passwords>`.
+
+*   'guest' is a user with minimum permissions used by default for remote :ref:`binary port <admin-security>` connections.
+    For example, 'guest' is used when connecting to the instance using :ref:`tt connect <tt-connect>` using the IP address and port without specifying the name of a user:
+
+    ..  code-block:: console
+
+        $ tt connect 192.168.10.10:3301
+
+
+..  NOTE::
+
+    Information about users is stored in the :ref:`_user <box_space-user>` space.
+
+
+
 
 .. _authentication-passwords:
 
---------------------------------------------------------------------------------
 Passwords
---------------------------------------------------------------------------------
+~~~~~~~~~
 
-Each user (except 'guest') may have a **password**.
+Any user (except 'guest') may have a password.
 The password is any alphanumeric string.
+Tarantool :ref:`password hashes <enterprise-authentication-protocol>` are stored in the :ref:`_user <box_space-user>` system space.
+So, if the password is '123456', the stored hash is a string like 'a7SDfrdDKRBe5FaN2n3GftLKKtk='.
 
-Tarantool passwords are stored in the :ref:`_user <box_space-user>`
-system space with a
-`cryptographic hash function <https://en.wikipedia.org/wiki/Cryptographic_hash_function>`_
-so that, if the password is ‘x’, the stored hash-password is a long string
-like ‘lL3OvhkIPOKh+Vn9Avlkx69M/Ck=‘.
-Tarantool supports two protocols for authenticating users:
+Tarantool Enterprise Edition allows you to improve database security by enforcing the use of strong passwords, setting up a maximum password age, and so on.
+Learn more from the :ref:`configuration_authentication` topic.
 
-*   `CHAP <https://en.wikipedia.org/wiki/Challenge-Handshake_Authentication_Protocol>`_ with ``SHA-1`` hashing
 
-    In this case, password hashes are stored in the ``_user`` space `unsalted <https://en.wikipedia.org/wiki/Salt_(cryptography)>`_.
-    If an attacker gains access to the database, they may crack a password using, for example, a `rainbow table <https://en.wikipedia.org/wiki/Rainbow_table>`_.
 
-*   `PAP <https://en.wikipedia.org/wiki/Password_Authentication_Protocol>`_ with ``SHA256`` hashing (Enterprise Edition)
+.. _access_control_concepts_objects:
 
-    For PAP, a password is salted with a user-unique salt before saving it in the ``_user`` space.
-    This keeps the database protected from cracking using a rainbow table.
-    Note that PAP sends a password as plain text, so you need to configure SSL/TLS for a connection.
+Objects
+~~~~~~~
 
-There are two functions for managing passwords in Tarantool:
+An object is a securable entity to which access can be granted.
+Tarantool has a number of objects that enable flexible management of access to data, stored functions, specific actions, and so on.
 
-*   :doc:`/reference/reference_lua/box_schema/user_passwd` allows you to change a user's password.
+Below are a few examples of objects:
 
-*   :doc:`/reference/reference_lua/box_schema/user_password` returns a hash of a user's password.
+*   'universe' represents a database (:ref:`box.schema <box_schema>`) that contains database objects, including spaces, indexes, users, roles, sequences, and functions.
+    Granting privileges to 'universe' gives a user access to any object in a database.
+*   'space' enables granting privileges to user-created or system :ref:`spaces <index-box_space>`.
+*   'function' enables granting privileges to :ref:`functions <box_schema-func_create>`.
 
-Tarantool Enterprise Edition also allows you to improve database security by enforcing the use of strong passwords, setting up a maximum password age, and so on. Learn more from the :ref:`configuration_authentication` topic.
+..  NOTE::
 
+    The full list of object types is available in the :ref:`access_control_list_objects` section.
 
 
 .. _authentication-owners_privileges:
 
---------------------------------------------------------------------------------
-Owners and privileges
---------------------------------------------------------------------------------
+Privileges
+~~~~~~~~~~
 
-Tarantool has one database. It may be called "box.schema" or "universe".
-The database contains database objects, including
-spaces, indexes, users, roles, sequences, and functions.
+The privileges granted to a user determine which operations the user can perform, for example:
 
-The **owner** of a database object is the user who created it.
-The owner of the database itself, and the owner of objects that
-are created initially (the system spaces and the default users)
-is '**admin**'.
+*   The 'read' and 'write' privileges granted to the 'space' :ref:`object <access_control_concepts_objects>` allow a user to select or update data in the specified space.
+*   The 'create' privilege granted to the 'space' object allows a user to create new spaces.
+*   The 'execute' privilege granted to the 'function' object allows a user to execute the specified function.
 
-Owners automatically have **privileges** for what they create.
-They can share these privileges with other users or with roles,
-using :doc:`/reference/reference_lua/box_schema/user_grant` requests.
-The following privileges can be granted:
+Note that some privileges might require read and write access to certain system spaces.
+For example, the 'create' privilege granted to the 'space' object requires 'read' and 'write' privileges to the :ref:`_space <box_space-space>` system space.
+Similarly, granting the ability to create functions requires 'read' and 'write' access to the :ref:`_func <box_space-func>` space.
 
-* 'read', e.g. allow select from a space
-* 'write', e.g. allow update on a space
-* 'execute', e.g. allow call of a function, or (less commonly) allow use of a role
-* 'create', e.g. allow
-  :doc:`box.schema.space.create </reference/reference_lua/box_schema/user_create>`
-  (access to certain system spaces is also necessary)
-* 'alter', e.g. allow
-  :doc:`box.space.x.index.y:alter </reference/reference_lua/box_index/alter>`
-  (access to certain system spaces is also necessary)
-* 'drop', e.g. allow
-  :doc:`box.sequence.x:drop </reference/reference_lua/box_schema_sequence/drop>`
-  (access to certain system spaces is also necessary)
-* 'usage', e.g. whether any action is allowable regardless of other
-  privileges (sometimes revoking 'usage' is a convenient way to
-  block a user temporarily without dropping the user)
-* 'session', e.g. whether the user can 'connect'.
+..  NOTE::
 
-To **create** objects, users need the 'create' privilege and
-at least 'read' and 'write' privileges
-on the system space with a similar name (for example, on the
-:ref:`_space <box_space-space>` if the user needs to create spaces).
+    Information about privileges is stored in the :ref:`_priv <box_space-priv>` space.
 
-To **access** objects, users need an appropriate privilege
-on the object (for example, the 'execute' privilege on function F
-if the users need to execute function F). See below some
-:ref:`examples for granting specific privileges <authentication-owners_privileges-examples-specific>`
-that a grantor -- that is, 'admin' or the object creator -- can make.
 
-To drop an object, a user must be an 'admin' or have the 'super' role.
-Some objects may also be dropped by their creators.
-As the owner of the entire database, any 'admin' can drop any object,
-including other users.
 
-To grant privileges to a user, the object owner says
-:doc:`/reference/reference_lua/box_schema/user_grant`.
-To revoke privileges from a user, the object owner says
-:doc:`/reference/reference_lua/box_schema/user_revoke`.
-In either case, there are up to five parameters:
+.. _access_control_concepts_roles:
+
+Roles
+~~~~~
+A role is a container for :ref:`privileges <authentication-owners_privileges>` that can be granted to users.
+Roles can also be assigned to other roles, creating a role hierarchy.
+
+There are the following built-in roles in Tarantool:
+
+*   'super' has all available administrative permissions.
+*   'public' is automatically granted to new users when they are created.
+*   'replication' can be granted to a user used to maintain replication in a cluster.
+*   'sharding' can be granted to a user used to maintain sharding in a cluster.
+
+Below are a few diagrams that demonstrate how privileges can be granted to a user without and with using roles.
+
+*   In this example, a user gets privileges directly without using roles.
+
+    ..  code-block:: none
+
+        user1 ── privilege1
+            ├─── privilege2
+            └─── privilege3
+
+*   In this example, a user gets all privileges provided by 'role1' and specific privileges assigned directly.
+
+    ..  code-block:: none
+
+        user1 ── role1 ── privilege1
+            │        └─── privilege2
+            ├─── privilege3
+            └─── privilege4
+
+*   In this example, 'role2' is granted to 'role1'.
+    This means that a user with 'role1' subsequently gets all privileges from both roles 'role1' and 'role2'.
+
+    ..  code-block:: none
+
+        user1 ── role1 ── privilege1
+            │        ├─── privilege2
+            │        └─── role2
+            │                 ├─── privilege3
+            │                 └─── privilege4
+            ├─── privilege5
+            └─── privilege6
+
+..  NOTE::
+
+    Information about roles is stored in the :ref:`_user <box_space-user>` space.
+
+
+
+.. _authentication-owners:
+
+Object owners
+~~~~~~~~~~~~~
+
+An owner of a database :ref:`object <access_control_concepts_objects>` is the user who created it.
+The owner of the database and the owner of objects that are created initially (the system spaces and the default users) is the 'admin' :ref:`user <access_control_concepts_users>`.
+
+Owners automatically have :ref:`privileges <authentication-owners_privileges>` for what they create.
+They can :ref:`share these privileges <access_control_granting_privileges>` with other users or roles using ``box.schema.user.grant()`` and ``box.schema.role.grant()``.
+
+..  NOTE::
+
+    Information about users who gave the specified privileges is stored in the :ref:`_priv <box_space-priv>` space.
+
+
+
+.. _authentication-sessions:
+
+Sessions
+~~~~~~~~
+
+A session is the state of a connection to Tarantool.
+The session contains:
+
+*   An integer ID identifying the connection.
+*   The current :ref:`user <authentication-users>` associated with the connection.
+*   The text description of the connected peer.
+*   A session's local state, such as Lua variables and functions.
+
+In Tarantool, a single session can execute multiple concurrent transactions.
+Each transaction is identified by a unique integer ID, which can be queried
+at the start of the transaction using :doc:`/reference/reference_lua/box_session/sync`.
+
+..  NOTE::
+
+    To track all connects and disconnects, you can use :ref:`connection and authentication triggers <triggers>`.
+
+
+
+
+.. _access_control_users:
+
+Users
+-----
+
+.. _access_control_user_creating:
+
+Creating a user
+~~~~~~~~~~~~~~~
+
+To create a new user, call :ref:`box.schema.user.create() <box_schema-user_create>`.
+In the example below, a user is created without a password:
+
+..  literalinclude:: /code_snippets/test/access_control/grant_user_privileges_test.lua
+    :language: lua
+    :start-after: Create a user without a password
+    :end-before: End: Create a user without a password
+    :dedent:
+
+
+In this example, the password is specified in the ``options`` parameter:
+
+..  literalinclude:: /code_snippets/test/access_control/grant_user_privileges_test.lua
+    :language: lua
+    :start-after: Create a user with a password
+    :end-before: End: Create a user with a password
+    :dedent:
+
+
+.. _access_control_user_changing_passwords:
+
+Changing passwords
+~~~~~~~~~~~~~~~~~~
+
+To set or change a user's password, use :ref:`box.schema.user.passwd() <box_schema-user_create>`.
+In the example below, a user password is set for a :ref:`currently <access_control_users_current_user>` logged-in user:
+
+..  literalinclude:: /code_snippets/test/access_control/grant_user_privileges_test.lua
+    :language: lua
+    :start-after: Set a password for the current user
+    :end-before: End: Set a password for the current user
+    :dedent:
+
+To set the password for the specified user, pass a username and password as shown below:
+
+..  literalinclude:: /code_snippets/test/access_control/grant_user_privileges_test.lua
+    :language: lua
+    :start-after: Set a password for the specified user
+    :end-before: End: Set a password for the specified user
+    :dedent:
+
+..  NOTE::
+
+    :doc:`/reference/reference_lua/box_schema/user_password` returns a hash of the specified password.
+
+
+
+.. _access_control_user_granting_privileges:
+
+Granting privileges to a user
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To grant the specified privileges to a user, use the ``box.schema.user.grant()`` function.
+In the example below, 'testuser' gets read privileges to the ``writers`` space and read/write privileges to the ``books`` space:
+
+..  literalinclude:: /code_snippets/test/access_control/grant_user_privileges_test.lua
+    :language: lua
+    :start-after: Grant privileges to the specified user
+    :end-before: End: Grant privileges to the specified user
+    :dedent:
+
+Learn more about granting privileges to different types of objects from :ref:`access_control_granting_privileges`.
+
+
+
+.. _access_control_user_revoking_privileges:
+
+Revoking user's privileges
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To revoke the specified privileges, use the :ref:`box.schema.user.revoke() <box_schema-user_revoke>` function.
+In the example below, write access to the ``books`` space is revoked:
+
+..  literalinclude:: /code_snippets/test/access_control/grant_user_privileges_test.lua
+    :language: lua
+    :start-after: Revoke space reading
+    :end-before: End: Revoke space reading
+    :dedent:
+
+Revoking the 'session' privilege from 'universe' can be used to disallow a user to connect to a Tarantool instance:
+
+..  literalinclude:: /code_snippets/test/access_control/grant_user_privileges_test.lua
+    :language: lua
+    :start-after: Revoke session
+    :end-before: End: Revoke session
+    :dedent:
+
+
+.. _access_control_user_info:
+
+Getting a user's information
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To check whether the specified user exists, call :ref:`box.schema.user.exists() <box_schema-user_exists>`:
+
+..  code-block:: lua
+
+    box.schema.user.exists('testuser')
+    --[[
+    - true
+    --]]
+
+To get information about privileges granted to a user, call :ref:`box.schema.user.info() <box_schema-user_info>`:
+
+..  code-block:: lua
+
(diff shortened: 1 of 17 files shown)


doc/book/admin/access_control.rst
```diff
@@ -1,270 +1,777 @@
...
+    ..  code-block:: console
+
+        $ tt connect app:instance001
+
+    ..  NOTE::
+
+        To allow remote :ref:`binary port <admin-security>` connections using the 'admin' user, you need to :ref:`set a password <access_control_user_changing_passwords>`.
+
+*   'guest' is a user with minimum permissions used by default for remote :ref:`binary port <admin-security>` connections.
+    For example, 'guest' is used when connecting to the instance using :ref:`tt connect <tt-connect>` using the IP address and port without specifying the name of a user:
```
Lord-KA: I would like to note that even default guest user with just "execute universe" grant (it is required for `tt connect`) can do Remote Code Execution on an instance. So it is very unsafe to have guest enabled with any non-default grants in a internet-accessible server, because it doesn't support passwords.

Lord-KA: It is actually a quite common mistake for CE users to leave server without firewall rules and with enabled guest.

doc/book/admin/access_control.rst
```diff
@@ -1,270 +1,854 @@
...
 .. _authentication-passwords:
 
---------------------------------------------------------------------------------
 Passwords
---------------------------------------------------------------------------------
+~~~~~~~~~
 
-Each user (except 'guest') may have a **password**.
+Any user (except 'guest') may have a password.
 The password is any alphanumeric string.
```
Lord-KA: We actually support other symbols too, e.g. `"!"`,`"/"`,`"]"` etc.

doc/book/admin/access_control.rst
```diff
@@ -1,270 +1,854 @@
...
+So, if the password is '123456', the stored hash is a string like 'a7SDfrdDKRBe5FaN2n3GftLKKtk='.
 
-Tarantool passwords are stored in the :ref:`_user <box_space-user>`
-system space with a
-`cryptographic hash function <https://en.wikipedia.org/wiki/Cryptographic_hash_function>`_
-so that, if the password is ‘x’, the stored hash-password is a long string
-like ‘lL3OvhkIPOKh+Vn9Avlkx69M/Ck=‘.
-Tarantool supports two protocols for authenticating users:
+Tarantool Enterprise Edition allows you to improve database security by enforcing the use of strong passwords, setting up a maximum password age, and so on.
+Learn more from the :ref:`configuration_authentication` topic.
```
Lord-KA, line 83: Tarantool EE also supports another authentication protocol that stores salted hashes. It would be nice to mention it here.

doc/book/admin/access_control.rst
Lord-KA: > may have a password

If the password is not set, the user can't be connected to.

doc/book/admin/access_control.rst
```diff
@@ -1,270 +1,854 @@
...
+~~~~~
+A role is a container for :ref:`privileges <authentication-owners_privileges>` that can be granted to users.
+Roles can also be assigned to other roles, creating a role hierarchy.
+
+There are the following built-in roles in Tarantool:
+
+*   'super' has all available administrative permissions.
+*   'public' is automatically granted to new users when they are created.
+*   'replication' can be granted to a user used to maintain replication in a cluster.
+*   'sharding' can be granted to a user used to maintain sharding in a cluster.
```
Lord-KA: The 'sharding' role is created only on instance managed with config.

doc/book/admin/access_control.rst
```diff
@@ -1,270 +1,854 @@
...
+
+.. _authentication-owners:
+
+Object owners
+~~~~~~~~~~~~~
+
+An owner of a database :ref:`object <access_control_concepts_objects>` is the user who created it.
+The owner of the database and the owner of objects that are created initially (the system spaces and the default users) is the 'admin' :ref:`user <access_control_concepts_users>`.
+
+Owners automatically have :ref:`privileges <authentication-owners_privileges>` for what they create.
```
Lord-KA: > for what they create

Nit: "for *objects* they create"?

doc/book/admin/access_control.rst
```diff
@@ -1,270 +1,854 @@
...
+
+..  NOTE::
+
+    Information about users who gave the specified privileges is stored in the :ref:`_priv <box_space-priv>` space.
+
+
+
+.. _authentication-sessions:
+
+Sessions
```
Lord-KA, line 204: Maybe note here, that 'session' is also a privilege that can be granted to a user (not roles).

doc/book/admin/access_control.rst
```diff
@@ -1,270 +1,854 @@
...
+        - universe
+        -
+      - - alter
+        - user
+        - testuser
+    --]]
+
+In the example above, 'testuser' has the following privileges:
+
+*   The 'execute' privilege to the 'public' role means that this role is :ref:`assigned to a user <access_control_roles_granting_user>`.
```
Lord-KA: Nit:
> assigned to *the* user

(We are talking about 'testuser' here, or some other specific user that has this grant)

Lord-KA: Maybe note here, that 'public' is granted to all users on creation?
(I know, that this is stated in the Roles section, but it could be helpful in describing a users privileges too)

doc/book/admin/access_control.rst
```diff
@@ -1,270 +1,854 @@
...
+    :end-before: End: Revoke space reading
+    :dedent:
+
+Revoking the 'session' privilege from 'universe' can be used to disallow a user to connect to a Tarantool instance:
+
+..  literalinclude:: /code_snippets/test/access_control/grant_user_privileges_test.lua
+    :language: lua
+    :start-after: Revoke session
+    :end-before: End: Revoke session
+    :dedent:
```
Lord-KA, line 375: This is a really cool feature!

doc/book/admin/access_control.rst
```diff
@@ -295,83 +879,179 @@ no-password user becomes this function's creator. Finally, another user
...
+
+..  container:: table
+
+    ..  list-table::
+        :header-rows: 1
+        :widths: 15 15 15 55
+
+        *   -   Privilege
+            -   Object type
+            -   Applied to roles
```
Lord-KA: Nit:
Maybe "Granted to roles"? I would interpret this, as the permission is ignored, but it actually can't be granted. Feel free to ignore.

Lord-KA (commented):
Thank you for the patchset! Now access control documentation looks much better overall. See my minor comments below

doc/book/admin/access_control.rst
andreyaksenov: Yep, a new Docs-as-tests approach :)

doc/book/admin/access_control.rst
andreyaksenov: > assigned to the user

Exactly! Thanks for the catch

doc/book/admin/access_control.rst
andreyaksenov: Good point, thanks. Added a warning:

image

doc/book/admin/access_control.rst
andreyaksenov: > The password is any alphanumeric string.

Decided to completely delete this sentence. Looks like mentioning special characters is useful in the `Password policy` section (which is linked below).

andreyaksenov: > If the password is not set, the user can't be connected to.

Changed to:

```
Any user (except ‘guest’) may have a password. 
If a password is not set, a user cannot connect to Tarantool instances.
```

doc/book/admin/access_control.rst
andreyaksenov: Thanks, added:

image

doc/book/admin/access_control.rst
andreyaksenov: Added to the example list of privileges as it looks important: 

image

doc/book/admin/access_control.rst
andreyaksenov: > Maybe note here, that 'public' is granted to all users on creation?
(I know, that this is stated in the Roles section, but it could be helpful in describing a users privileges too)

I'd keep it as is because it is a how-to section and maintaining such info in several places might be hard if some privileges/roles change for the default user.

doc/book/admin/access_control.rst
andreyaksenov: Added a note:

image

doc/book/admin/access_control.rst
Lord-KA: > Exactly! Thanks for the catch

It seems like there is the same issue in other bullet-points below

andreyaksenov: > It seems like there is the same issue in other bullet-points below

Thanks, fixed

doc/book/admin/access_control.rst
```diff
@@ -1,270 +1,865 @@
...
+.. _access_control:
 
-================================================================================
 Access control
-================================================================================
+==============
 
-This section explains how Tarantool makes it possible for administrators
-to prevent unauthorized access to the database and to certain functions.
+Tarantool enables flexible management of access to various database resources.
```
p7nov, line 7: `database resources` - does this wording cover all entities we can manage in Tarantool? I mean, in scope of the _middleware_/_platform_ positioning.

doc/book/admin/access_control.rst
```diff
@@ -1,270 +1,865 @@
...
 
-There is a **current user** for any program working with Tarantool,
-local or remote.
-If a remote connection is using a :ref:`binary port <admin-security>`,
-the current user, by default, is '**guest**'.
-If the connection is using an :ref:`admin-console port <admin-security>`,
-the current user is '**admin**'.
-When executing a :ref:`Lua initialization script <index-init_label>`,
-the current user is also ‘**admin**’.
+A user is a person or program that interacts with a Tarantool instance.
```
p7nov: This definition looks too broad and vague without details.
I'd provide a more detailed explanation: 
- Technically, a user we talk about is an _entity in Tarantool_, not a person or external program.
- But it corresponds (== _can be used by_) either to a person, or to a program, enabling them to do something

doc/book/admin/access_control.rst
```diff
@@ -1,270 +1,865 @@
...
+There might be different types of users, for example:
 
-The current user name can be found with
-:doc:`/reference/reference_lua/box_session/user`.
+*   A database administrator is responsible for the overall management and administration of a database.
+    An administrator can create other users and grant them specified privileges.
+*   A regular user has limited access to certain data and stored functions. For example, such users can be used to let :ref:`connectors <index-box_connectors>` communicate to a database.
```
p7nov: I don't think that _regular user_ is definitive enough. Maybe write a more specific example?

p7nov: Additionally, this example confused me: from the start, I expected this to be a _person_ that works on data in specific spaces, but then it turned out that it's about connectors.

doc/book/admin/access_control.rst
```diff
@@ -1,270 +1,865 @@
...
+*   Users used in communications between Tarantool instances. For example, such users can be created to maintain replication and sharding in a Tarantool cluster.
```
p7nov, line 35: These users: are they people or programs? I think a deeper explanation is needed here.

doc/book/admin/access_control.rst
```diff
@@ -1,270 +1,865 @@
...
 
-The current user can be changed:
+There are two built-in users in Tarantool:
+
+*   'admin' is a user with all available administrative permissions.
```
p7nov: Let's use the code style (double backticks) for all user, role, privilege, etc. names on this page.

doc/book/admin/access_control.rst
```diff
@@ -1,270 +1,865 @@
...
+    If the connection uses an :ref:`admin-console port <admin-security>`, the current user is 'admin'.
+    For example, 'admin' is used when connecting to the instance using :ref:`tt connect <tt-connect>` locally using the instance name:
+
+    ..  code-block:: console
+
+        $ tt connect app:instance001
+
+    To allow remote :ref:`binary port <admin-security>` connections using the 'admin' user, you need to :ref:`set a password <access_control_user_changing_passwords>`.
+
+*   'guest' is a user with minimum permissions used by default for remote :ref:`binary port <admin-security>` connections.
```
p7nov: Maybe write explicitly that guest cannot read/modify any data or execute code? _Minimum permissions_ is pretty vague.

doc/book/admin/access_control.rst
```diff
@@ -1,270 +1,865 @@
...
-:doc:`/reference/reference_lua/box_schema/user_revoke`.
-In either case, there are up to five parameters:
+Privileges
+~~~~~~~~~~
 
-.. code-block:: lua
+The privileges granted to a user determine which operations the user can perform, for example:
 
-    (user-name, privilege, object-type [, object-name [, options]])
+*   The 'read' and 'write' privileges granted to the 'space' :ref:`object <access_control_concepts_objects>` allow a user to select or update data in the specified space.
```
p7nov: I'd use `read and modify` instead of `select and update` -- these can be read as exact function names, while `read and modify` covers sets of operations.

doc/book/admin/access_control.rst
```diff
@@ -1,270 +1,865 @@
...
+~~~~~
+A role is a container for :ref:`privileges <authentication-owners_privileges>` that can be granted to users.
+Roles can also be assigned to other roles, creating a role hierarchy.
+
+There are the following built-in roles in Tarantool:
+
+*   'super' has all available administrative permissions.
+*   'public' is automatically granted to new users when they are created.
+*   'replication' can be granted to a user used to maintain replication in a cluster.
+*   'sharding' can be granted to a user used to maintain sharding in a cluster.
```
p7nov: I think the two last roles may need more explanation: If I get it right, their usage is rather specific.
Some links to replication and sharding docs would also help.

doc/book/admin/access_control.rst
```diff
@@ -1,270 +1,865 @@
...
+Managing users
+--------------
+
+.. _access_control_user_creating:
+
+Creating a user
+~~~~~~~~~~~~~~~
+
+To create a new user, call :ref:`box.schema.user.create() <box_schema-user_create>`.
+In the example below, a user is created without a password:
```
p7nov, line 237: A note about inability to connect for such users would be useful here.

doc/book/admin/access_control.rst
```diff
@@ -1,270 +1,865 @@
...
+    :dedent:
+
+
+.. _access_control_user_changing_passwords:
+
+Changing passwords
+~~~~~~~~~~~~~~~~~~
+
+To set or change a user's password, use :ref:`box.schema.user.passwd() <box_schema-user_create>`.
+In the example below, a user password is set for a :ref:`currently <access_control_users_current_user>` logged-in user:
```
p7nov: A link under `currently` looks weird to me
image

doc/book/admin/access_control.rst
```diff
@@ -1,270 +1,865 @@
...
+    :doc:`/reference/reference_lua/box_schema/user_password` returns a hash of the specified password.
+
+
+
+.. _access_control_user_granting_privileges:
+
+Granting privileges to a user
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To grant the specified privileges to a user, use the ``box.schema.user.grant()`` function.
```
p7nov: Link to the function reference?

doc/book/admin/access_control.rst
```diff
@@ -1,270 +1,865 @@
...
+To check whether the specified user exists, call :ref:`box.schema.user.exists() <box_schema-user_exists>`:
+
+..  code-block:: lua
+
+    box.schema.user.exists('testuser')
+    --[[
+    - true
+    --]]
+
+To get information about privileges granted to a user, call :ref:`box.schema.user.info() <box_schema-user_info>`:
```
p7nov, line 315: We can reuse the example you added in the [box.schema.user.info()](https://docs.d.tarantool.io/en/doc/config-credentials/reference/reference_lua/box_schema/user_info/#box-schema-user-info) reference. It looks more practical than the list of `admin`'s privileges shown now.

doc/book/admin/access_control.rst
```diff
@@ -1,270 +1,865 @@
...
+*   The 'alter' privilege lets 'testuser' modify its own settings, for example, a password.
+
+
+
+.. _access_control_user_revoking_privileges:
+
+Revoking user's privileges
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To revoke the specified privileges, use the :ref:`box.schema.user.revoke() <box_schema-user_revoke>` function.
```
p7nov, line 360: Can we revoke roles from users?

doc/book/admin/access_control.rst
```diff
@@ -1,270 +1,865 @@
...
+To grant the specified privileges to a role, use the ``box.schema.role.grant()`` function.
+In the example below, the 'books_space_manager' role gets read and write privileges to the ``books`` space:
+
+..  literalinclude:: /code_snippets/test/access_control/grant_roles_test.lua
+    :language: lua
+    :start-after: Grant read/write privileges to a role
+    :end-before: Grant write privileges to a role
+    :dedent:
+
+The 'writers_space_reader' role gets read privileges to the ``books`` space:
```
p7nov: `writers` space

doc/book/admin/access_control.rst
p7nov: Found this. Let's add a link)

doc/book/admin/access_control.rst
```diff
@@ -1,270 +1,865 @@
...
+
+To revoke role's privileges, use :ref:`box.schema.role.revoke() <box_schema-role_revoke>`.
+
+
+.. _access_control_roles_dropping:
+
+Dropping roles
+~~~~~~~~~~~~~~
+
+To drop the specified role, call :ref:`box.schema.role.drop() <box_schema-role_drop>`:
```
p7nov, line 576: Are there conditions for role deletion? Typically for such cases, the existence of user with this role is checked.

doc/book/admin/access_control.rst
```diff
@@ -1,270 +1,865 @@
...
-      box.schema.user.grant('U','read','space','_func_index')
-    -- So that 'U' can drop any space (ignore if the privilege exists already)
-      box.schema.user.grant('U','drop','space',nil,{if_not_exists=true})
-
-**Example for creating users and objects then granting privileges**
-
-Here a Lua function is created that will be executed under the user ID of its
+    box.schema.user.grant(user-name, privileges, object-type, object-name[, {options}])
+
+*   ``user-name``: the name of the user that gets the specified privileges.
```
p7nov: Hyphenated `user-name` looks grammatically incorrect (although I get why it is used here). Let's maybe change to either `user` or `username`?

doc/book/admin/access_control.rst
```diff
@@ -1,270 +1,865 @@
...
+
+Spaces
+~~~~~~
+
+.. _access_control_grant_spaces_create:
+
+Creating and altering spaces
+****************************
+
+In the example below, 'testuser' gets privileges allowing them to create :ref:`spaces <index-box_space>`:
```
p7nov: The need to grant privileges to implicitly connected system objects is really hardcore. Can we maybe find a place to explain this in general before showing these lists of calls?

doc/book/admin/access_control.rst
```diff
@@ -1,270 +1,865 @@
...
+
+.. _creating_users_and_objects_granting_privileges:
+
+Example
+~~~~~~~
+
+In the example below, the created Lua function is executed under the user ID of its
 creator, even if called by another user.
 
 First, the two spaces ('u' and 'i') are created, and a no-password user ('internal')
```
p7nov: Let's use more self-explanatory and realistic entity names than `u`, `i`, and `internal`.

doc/book/admin/access_control.rst
```diff
@@ -295,83 +890,179 @@ no-password user becomes this function's creator. Finally, another user
...
+Object types and privileges
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+..  container:: table
+
+    ..  list-table::
+        :header-rows: 1
+        :widths: 15 85
+
+        *   -   Object type
```
p7nov, line 1009: Wouldn't this look better in a three-column table?

doc/code_snippets/test/access_control/grant_user_privileges_test.lua
```diff
@@ -0,0 +1,138 @@
...
+        box.schema.user.create('testuser')
+        -- End: Create a user without a password --
+        t.assert_equals(box.space._user.index.name:select { 'testuser' }[1][5]['chap-sha1'], nil)
+    end)
+end
+
+g.test_user_with_password_created = function(cg)
+    cg.server:exec(function()
+        -- Create a user with a password --
+        box.schema.user.create('testuser', { password = 'foobar' })
```
p7nov, line 69: Nit: most style guides recommend avoiding `foo`/`bar` anywhere in examples.

p7nov (changes requested):
Please see my comments

doc/book/admin/access_control.rst
Lord-KA: > a password is salted with a user-unique salt
Maybe:
a password is salted with randomly generated user-unique data


Actually, I don't think we need to get into details of implementation, password salt is always random and unique, so maybe just say we support it, it may be a required criteria for some organization.

andreyaksenov: Okay, let's remove this sentence - the reader can find this info in the linked `Authentication` page.

doc/book/admin/access_control.rst
andreyaksenov: I hope, yes. Tarantool is used as the base for many systems but still remains the _database_ under the hood.

doc/book/admin/access_control.rst
andreyaksenov: I feel that _identify_ should work better here:

> A user identifies a person or program that interacts with a Tarantool instance.

And this is eliminates the need of using the _entity_ word that might add some confusion :)

doc/book/admin/access_control.rst
andreyaksenov: Good point. Removed information about connectors because this is a specific case. Both admin and regular user might work directly or using a connector.
Also, added info that users can get privileges from dbadmin.

doc/book/admin/access_control.rst
andreyaksenov: Strictly speaking, guest cannot do anything without additional permissions (including reading and modifying any data). The _vague_ word describes exactly theirs permissions :) Guest can do smth that is hard to convey in one sentence.

doc/book/admin/access_control.rst
andreyaksenov: Added links to corresponding docs.

> I think the two last roles may need more explanation

Let's keep it as is for now. I believe that the detailed info about these roles should be described in topics covering replication and sharding details. I can say smth about replication but how the `sharding` role helps maintaining communications requires investigation.

doc/book/admin/access_control.rst
andreyaksenov: Strictly speaking, this user can connect but cannot execute Lua code after connecting using the admin console:

image

So, this depends on whether we talk about a dbadmin (needs such access) or a user with limited privileges to certain data.

doc/book/admin/access_control.rst
andreyaksenov: Hmm.. Thought about this but cell merging in the first column is required for better readability. Does Sphinx support this?

doc/book/admin/access_control.rst
andreyaksenov: Okay :) just copy-pasted this example from old docs as is and was a bit lazy to fix this.

doc/book/admin/access_control.rst
andreyaksenov: Hmm.. In this case we need to fix this in all reference pages related to privileges (because I just copy-pasted this from the `box.schema.user.grant` page).

doc/book/admin/access_control.rst
andreyaksenov: Maybe, I will update all `box.schema.grant/revoke` descriptions and include all new examples there.

doc/book/admin/access_control.rst
andreyaksenov: As I see, Tarantool silently removes a role even if it is assigned to a user.

doc/book/admin/access_control.rst
andreyaksenov: I hope, this should be enough:

image

doc/book/admin/access_control.rst
Lord-KA: Ok

Lord-KA (approved):

doc/code_snippets/test/access_control/grant_user_privileges_test.lua
Lord-KA: Usually password is "secret", or something similar.

doc/book/admin/access_control.rst
p7nov: AFAIR, no: tables in Sphinx/RST are a bit limited.

doc/book/admin/access_control.rst
p7nov: Yes, but I have two concerns:
1. This text is too far from the section where this question came up. Hardly somebody will read the whole page of this size.
2. ``Note that some`` looks like some rare or specific thing. To me, this whole point seems like a general concept of the Tarantool access model. I'd explain it in a stronger and more exact wording. Just a personal opinion, you decide.

doc/book/admin/access_control.rst
```diff
@@ -1,377 +1,1069 @@
 .. _authentication:
+.. _access_control:
 
-================================================================================
 Access control
-================================================================================
+==============
```
p7nov, line 5: Idea: use the words `role-based access control system/model` explicitly somewhere in the intro paragraph. This is a familiar term, it helps readers understand general concepts easier.

doc/book/admin/access_control.rst
```diff
@@ -1,377 +1,1069 @@
...
 
-This section explains how Tarantool makes it possible for administrators
-to prevent unauthorized access to the database and to certain functions.
+Tarantool enables flexible management of access to various database resources.
+The main concepts of Tarantool access control system are as follows:
 
-Briefly:
+*   A *user* is a person or program that interacts with a Tarantool instance.
+*   An *object* is an entity to which access can be granted, for example, spaces, indexes, or functions.
```
p7nov: Singular `a space`, `an index`, `a function` would be more precise, since we grant permissions per object, not for all spaces (functions) at once.

doc/book/admin/access_control.rst
```diff
@@ -1,377 +1,1069 @@
...
+*   *Privileges* allow a user to perform certain operations on specific objects, for example, creating spaces, reading or updating data.
```
p7nov: Maybe `A privilege` for consistency with other list items?

doc/book/admin/access_control.rst
```diff
@@ -1,377 +1,1069 @@
...
+Roles can also be assigned to other roles, creating a role hierarchy.
 
-**Examples for granting privileges for specific operations**
+There are the following built-in roles in Tarantool:
 
-In these examples an administrator grants strictly
-the minimal privileges necessary for particular operations,
-to user 'U'.
+*   ``super`` has all available administrative permissions.
+*   ``public`` is automatically granted to new users when they are created.
```
p7nov: Can we say anything specific or give a usage advice about this role? The current definition gives me no idea why is it needed and how to use it...

doc/book/admin/access_control.rst
```diff
@@ -1,377 +1,1069 @@
...
+*   ``replication`` can be granted to a user used to maintain :ref:`replication <replication_overview>` in a cluster.
+*   ``sharding`` can be granted to a user used to maintain :ref:`sharding <sharding>` in a cluster.
```
p7nov, line 141: I'm still confused about the two last roles. 
Are they purely internal/technical by purpose? Should a person ever use them manually or change their privileges?
I mean, if I need to administer a cluster, do I connect to an instance using these role and call some functions by hand? Or Tarantool only uses them under the hood?

doc/book/admin/access_control.rst
```diff
@@ -1,377 +1,1069 @@
...
+
+
+.. _access_control_users:
+
+Managing users
+--------------
+
+.. _access_control_user_creating:
+
+Creating a user
```
p7nov, line 233: An idea for this an similar task-oriented subsections: mention a privilege that needed for doing this.

```
Creating a user

Required privilege: `write` on the `_users` space.

To create a new user ...
```

doc/book/admin/access_control.rst
```diff
@@ -1,377 +1,1069 @@
...
+    box.schema.user.grant('testuser','read','space','_ck_constraint')
+    box.schema.user.grant('testuser','read','space','_func_index')
+
+
+.. _access_control_grant_spaces_indexes:
+
+Creating and altering indexes
+*****************************
+
+In the example below, ``testuser`` gets privileges allowing them to create :ref:`indexes <index-box_index>` in the 'writers' space:
```
p7nov, line 673: The space name `writers` isn't highlighted as code, and there are more such names below (books, id_seq, sum). Please recheck the whole page.

doc/book/admin/access_control.rst
```diff
@@ -1,377 +1,1069 @@
...
+
+
+
+
+.. _creating_users_and_objects_granting_privileges:
+
+Example
+~~~~~~~
+
+In the example below, the created Lua function is executed under the user ID of its
```
p7nov: `on behalf of` sounds more natural than than `under the user ID`.

doc/book/admin/access_control.rst
```diff
@@ -1,377 +1,1069 @@
...
+      local fiber = require('fiber')
+      local t = space1:get{key}
+      if t ~= nil then
+        space1:put{key, box.session.uid()}
+        space2:put{key, fiber.time()}
+      end
+    end
+
+    box.session.su('private_user')
+    box.schema.func.create('read_and_modify', {setuid= true})
```
p7nov, line 890: I believe, this is what should be explained explicitly: `setuid = true`.
Most other code is pretty obvious, but this option is
1) new and unknown to readers
2) key - it is what sets the user for function execution

doc/book/admin/access_control.rst
```diff
@@ -1,377 +1,1069 @@
...
+
+.. _creating_users_and_objects_granting_privileges:
+
+Example
+~~~~~~~
+
+In the example below, the created Lua function is executed under the user ID of its
+creator, even if called by another user.
+
+First, the two spaces (``space1`` and ``space2``) are created, and a no-password user (``private_user``)
```
p7nov, line 860: nit: `no-password user` sounds weird, especially when used three times in a row as a user's identifier in text. If it's really important, let's say it once and not repeat. Although I'd rephrase this completely.

doc/book/admin/access_control.rst
```diff
@@ -1,377 +1,1069 @@
...
+        space1:put{key, box.session.uid()}
+        space2:put{key, fiber.time()}
+      end
+    end
+
+    box.session.su('private_user')
+    box.schema.func.create('read_and_modify', {setuid= true})
+    box.session.su('admin')
+    box.schema.user.create('public_user', {password = 'secret'})
+    box.schema.user.grant('public_user', 'execute', 'function', 'read_and_modify')
```
p7nov, line 893: Let's maybe write a conclusion sentence, smth like: `now whenever public_user calls the function, it is executed on behalf of its creator, private_user`

doc/reference/reference_lua/box_schema/user_create.rst
```diff
@@ -21,7 +21,7 @@ box.schema.user.create()
     * ``password`` (default = '') - string; the ``password`` = *password*
       specification is good because in a :ref:`URI <index-uri>`
       (Uniform Resource Identifier) it is usually illegal to include a
-      user-name without a password.
+      username without a password.
 
     .. NOTE::
```
p7nov, line 26: Wow, this surprised me a lot! 
Isn't it important enough to mention in the **Users** section of **Access control**?

image

p7nov (approved):
LGTM in general, just some more ideas and qeustion.

doc/book/admin/access_control.rst
andreyaksenov: > This text is too far from the section where this question came up. Hardly somebody will read the whole page of this size.

Will add a note after the first example to remind a reader about this concept.

doc/code_snippets/test/access_control/grant_user_privileges_test.lua
andreyaksenov: > Usually password is "secret", or something similar.

Thanks for your suggestion. Honestly, I try not to use meaningful words as password examples in docs because they make a reader think about the word meaning. IMO, passwords should look like some gibberish, well-known weak passwords (`123456`), or smth like `P@$$word`) to allow a reader quickly scan it und understand it is a _password_, not smth else. Not sure I'm right here :)

doc/book/admin/access_control.rst
andreyaksenov: > Are they purely internal/technical by purpose? 

AFAIU, yes. They are granted to corresponding _service_ users in a cluster config.

> Should a person ever use them manually or change their privileges?

No, they designed only for specific internal operations.

doc/book/admin/access_control.rst
andreyaksenov: Added a link to the `box.schema.func.create` page where `setuid` is described in details.

doc/book/admin/access_control.rst
andreyaksenov: Replaced `no-password user` with `private_user` in the next sentences.

doc/reference/reference_lua/box_schema/user_create.rst
andreyaksenov: This is also described in Limitations: https://www.tarantool.io/en/doc/latest/book/box/limitations/.
I'd keep it as is because it might be hard to maintain such reference information is multiple places.

doc/book/admin/access_control.rst
andreyaksenov: Privileges can be assigned directly to a user. So, `role-based access control` doesn't cover all possible capabilities of Tarantool access control system.

PR #4042, opened 2024-02-13
Sharding get started

### LINKED ISSUE #3659: [Config] How-to: sharding configuration
**Related doc issues:** https://github.com/tarantool/doc/issues?q=is%3Aissue+is%3Aopen+config+label%3A3.0+label%3Aconfig+label%3Avshard

**Product:** Tarantool
**Since:** 3.0
**Root document:** https://www.tarantool.io/en/doc/latest/how-to/vshard_quick/
**Sample:** [sharding/instances.enabled/sharded_cluster](https://github.com/tarantool/doc/tree/latest/doc/code_snippets/snippets/sharding/instances.enabled/sharded_cluster)
**SME:** @ Totktonada

# Details

Other examples: 

- https://gist.github.com/Totktonada/6339517a32cd7df4ac7df700fc3fa6b8
- https://github.com/tarantool/tarantool-cluster-manager/pull/1061/files
- https://github.com/tarantool/tt/commit/06bdb55869b826f91541223087a5c76957c45a10

Note from @ filonenko-mikhail: document UUID for configuring vshard topology

See also: https://github.com/tarantool/doc/issues/3660

andreyaksenov:
- Updated the existing How-to topic to using a new config: https://docs.d.tarantool.io/en/doc/sharding-get-started/how-to/vshard_quick/.

   > Note that there are still no links to configuration options related to sharding. They will be documented later.

- Updated `README` - removed most of the steps that duplicate content from the new topic: https://github.com/tarantool/doc/tree/sharding-get-started/doc/code_snippets/snippets/sharding/instances.enabled/sharded_cluster.

- Some cosmetic changes in the app: updated storage function names to distinguish them from router functions.

- Updated `vshard` version to `0.1.26`.

### DIFF
--- a/doc/how-to/vshard_quick.rst
+++ b/doc/how-to/vshard_quick.rst
@@ -95,10 +95,16 @@ Add the :ref:`credentials <configuration_reference_credentials>` configuration s
     :end-at: roles: [sharding]
     :dedent:
 
-In this section, two users are created:
+In this section, two users with the specified passwords are created:
 
 *   The ``replicator`` user with the ``replication`` role.
 *   The ``storage`` user with the ``sharding`` role.
+
+..  WARNING::
+
+    It is recommended to load passwords from safe storage such as external files or environment variables.
+    You can learn how to do this from :ref:`configuration_credentials_loading_secrets`.
+
 
 
 ..  _vshard-quick-start-configuring-cluster-advertise:
@@ -170,9 +176,9 @@ Here is a schematic view of the cluster's topology:
         :end-before: routers:
         :dedent:
 
-    The main top-level options here are:
+    The main group-level options here are:
 
-    *   ``app``: The ``app.module`` option specifies that code specific to storages should be loaded from the ``storage`` module. See also: :ref:`vshard-quick-start-storage-code`.
+    *   ``app``: The ``app.module`` option specifies that code specific to storages should be loaded from the ``storage`` module. This is explained below in the :ref:`vshard-quick-start-storage-code` section.
     *   ``sharding``: The ``sharding.roles`` option specifies that all instances inside this group act as storages.
         A rebalancer is selected automatically from two master instances.
     *   ``replication``: The :ref:`replication.failover <configuration_reference_replication_failover>` option specifies that a leader in each replica set should be specified manually.
@@ -187,9 +193,9 @@ Here is a schematic view of the cluster's topology:
         :end-at: 127.0.0.1:3300
         :dedent:
 
-    The main top-level options here are:
+    The main group-level options here are:
 
-    *   ``app``: The ``app.module`` option specifies that code specific to a router should be loaded from the ``router`` module. See also: :ref:`vshard-quick-start-router-code`.
+    *   ``app``: The ``app.module`` option specifies that code specific to a router should be loaded from the ``router`` module. This is explained below in the :ref:`vshard-quick-start-router-code` section.
     *   ``sharding``: The ``sharding.roles`` option specifies that an instance inside this group acts as a router.
     *   ``replicasets``: This section configures one replica set with one router instance.
 
@@ -312,7 +318,7 @@ Open the ``sharded_cluster-scm-1.rockspec`` file and add the following content:
     :language: none
     :dedent:
 
-In the ``dependencies`` section, you can see the specified version of the ``vshard`` module.
+The ``dependencies`` section includes the specified version of the ``vshard`` module.
 To install dependencies, you need to :ref:`build the application <vshard-quick-start-building-app>`.
 
 
@@ -432,14 +438,12 @@ To check the cluster's status, execute :ref:`vshard.router.info() <router_api-in
     ...
 
 
-.. _vshard-quick-start-working-adding-data:
+.. _vshard-quick-start-working-adding-selecting-data:
 
-Adding data
-~~~~~~~~~~~
+Adding and selecting data
+~~~~~~~~~~~~~~~~~~~~~~~~~
 
-To check how data is distributed across the cluster's nodes, follow the steps below:
-
-1.  On the router, call the ``insert_data()`` function:
+1.  To insert sample data, call the :ref:`insert_data() <vshard-quick-start-router-code>` function on the router:
 
     .. code-block:: console
 
@@ -447,7 +451,36 @@ To check how data is distributed across the cluster's nodes, follow the steps be
         ---
         ...
 
-2.  Connect to any storage in the ``storage-a`` replica set:
+    Calling this function :ref:`distributes data <vshard-quick-start-working-adding-data>` evenly across the cluster's nodes.
+
+2.  To get a tuple by the specified ID, call the ``get()`` function:
+
+    .. code-block:: console
+
+        sharded_cluster:router-a-001> get(4)
+        ---
+        - [4, 'The Beatles', 1960]
+        ...
+
+3.  To insert a new tuple, call the ``put()`` function:
+
+    .. code-block:: console
+
+        sharded_cluster:router-a-001> put(11, 'The Who', 1962)
+        ---
+        ...
+
+
+
+
+.. _vshard-quick-start-working-adding-data:
+
+Checking data distribution
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To check how data is distributed across the cluster's nodes, follow the steps below:
+
+1.  Connect to any storage in the ``storage-a`` replica set:
 
     ..  code-block:: console
 
@@ -461,15 +494,16 @@ To check how data is distributed across the cluster's nodes, follow the steps be
 
         sharded_cluster:storage-a-001> box.space.bands:select()
         ---
-        - - [1, 614, 'Roxette', 1986]
-          - [2, 986, 'Scorpions', 1965]
-          - [5, 755, 'Pink Floyd', 1965]
-          - [7, 998, 'The Doors', 1965]
-          - [8, 762, 'Nirvana', 1987]
+        - - [3, 11, 'Ace of Base', 1987]
+          - [4, 42, 'The Beatles', 1960]
+          - [6, 55, 'The Rolling Stones', 1962]
+          - [9, 299, 'Led Zeppelin', 1968]
+          - [10, 167, 'Queen', 1970]
+          - [11, 70, 'The Who', 1962]
         ...
 
 
-3.  Connect to any storage in the ``storage-b`` replica set:
+2.  Connect to any storage in the ``storage-b`` replica set:
 
     ..  code-block:: console
 
@@ -483,9 +517,9 @@ To check how data is distributed across the cluster's nodes, follow the steps be
 
         sharded_cluster:storage-b-001> box.space.bands:select()
         ---
-        - - [3, 11, 'Ace of Base', 1987]
-          - [4, 42, 'The Beatles', 1960]
-          - [6, 55, 'The Rolling Stones', 1962]
-          - [9, 299, 'Led Zeppelin', 1968]
-          - [10, 167, 'Queen', 1970]
+        - - [1, 614, 'Roxette', 1986]
+          - [2, 986, 'Scorpions', 1965]
+          - [5, 755, 'Pink Floyd', 1965]
+          - [7, 998, 'The Doors', 1965]
+          - [8, 762, 'Nirvana', 1987]
         ...
--- a/doc/how-to/vshard_quick.rst
+++ b/doc/how-to/vshard_quick.rst
@@ -43,17 +43,17 @@ In this tutorial, the application layout is prepared manually:
 
 1.  Create a tt environment in the current directory by executing the :ref:`tt init <tt-init>` command.
 
-2.  Inside the ``instances.enabled`` directory of the created tt environment, create the ``sharded_cluster`` directory.
+2.  Inside the empty ``instances.enabled`` directory of the created tt environment, create the ``sharded_cluster`` directory.
 
 3.  Inside ``instances.enabled/sharded_cluster``, create the following files:
 
     -   ``instances.yml`` specifies instances to run in the current environment.
-    -   The ``config.yaml`` file is intended to store the cluster's :ref:`configuration <configuration_overview>`.
-    -   ``storage.lua`` is intended to store code specific for :ref:`storages <vshard-architecture-storage>`.
-    -   ``router.lua`` is intended to store code specific for a :ref:`router <vshard-architecture-router>`.
-    -   ``sharded_cluster-scm-1.rockspec`` includes external dependencies required by the application.
+    -   ``config.yaml`` specifies the cluster's :ref:`configuration <configuration_overview>`.
+    -   ``storage.lua`` contains code specific for :ref:`storages <vshard-architecture-storage>`.
+    -   ``router.lua`` contains code specific for a :ref:`router <vshard-architecture-router>`.
+    -   ``sharded_cluster-scm-1.rockspec`` specifies external dependencies required by the application.
 
-    The next :ref:`vshard-quick-start-developing-app` section shows how to configure the cluster and write code specific for a router and storages.
+    The next :ref:`vshard-quick-start-developing-app` section shows how to configure the cluster and write code for routing read and write requests to different storages.
 
 
 ..  _vshard-quick-start-developing-app:
@@ -100,10 +100,12 @@ In this section, two users with the specified passwords are created:
 *   The ``replicator`` user with the ``replication`` role.
 *   The ``storage`` user with the ``sharding`` role.
 
-..  WARNING::
+These users are intended to maintain replication and sharding in the cluster.
 
-    It is recommended to load passwords from safe storage such as external files or environment variables.
-    You can learn how to do this from :ref:`configuration_credentials_loading_secrets`.
+..  IMPORTANT::
+
+    It is not recommended to store passwords as plain text in a YAML configuration.
+    Learn how to load passwords from safe storage such as external files or environment variables from :ref:`configuration_credentials_loading_secrets`.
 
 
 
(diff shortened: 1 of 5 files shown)


ImeevMA (approved):

doc/how-to/vshard_quick.rst
```diff
@@ -1,160 +1,491 @@
...
+
+1.  To configure storages, add the following code inside the ``groups`` section:
+
+    ..  literalinclude:: /code_snippets/snippets/sharding/instances.enabled/sharded_cluster/config.yaml
+        :language: yaml
+        :start-at: storages:
+        :end-before: routers:
+        :dedent:
+
+    The main top-level options here are:
```
Totktonada: Maybe it is better to say 'group-level' instead of 'top-level' to avoid possible misunderstanding.

doc/how-to/vshard_quick.rst
```diff
@@ -1,160 +1,491 @@
...
+
+2.  To configure a router, add the following code inside the ``groups`` section:
+
+    ..  literalinclude:: /code_snippets/snippets/sharding/instances.enabled/sharded_cluster/config.yaml
+        :language: yaml
+        :start-at: routers:
+        :end-at: 127.0.0.1:3300
+        :dedent:
+
+    The main top-level options here are:
```
Totktonada: Ditto.

doc/how-to/vshard_quick.rst
```diff
@@ -1,160 +1,491 @@
...
+
+    ..  literalinclude:: /code_snippets/snippets/sharding/instances.enabled/sharded_cluster/config.yaml
+        :language: yaml
+        :start-at: storages:
+        :end-before: routers:
+        :dedent:
+
+    The main top-level options here are:
+
+    *   ``app``: The ``app.module`` option specifies that code specific to storages should be loaded from the ``storage`` module. See also: :ref:`vshard-quick-start-storage-code`.
```
Totktonada: 'See also' is in imperative mood and motivates to click on the link. We can say 'explained below in the <...> section', so a reader knows that he/she doesn't miss this information if just continuous read the artice.

doc/how-to/vshard_quick.rst
```diff
@@ -1,160 +1,491 @@
...
+
+    ..  literalinclude:: /code_snippets/snippets/sharding/instances.enabled/sharded_cluster/config.yaml
+        :language: yaml
+        :start-at: routers:
+        :end-at: 127.0.0.1:3300
+        :dedent:
+
+    The main top-level options here are:
+
+    *   ``app``: The ``app.module`` option specifies that code specific to a router should be loaded from the ``router`` module. See also: :ref:`vshard-quick-start-router-code`.
```
Totktonada: Ditto.

doc/how-to/vshard_quick.rst
```diff
@@ -1,160 +1,491 @@
...
-
-``vshard.storage.cfg()`` automatically calls ``box.cfg()`` and configures the listen
-port and replication parameters.
-
-For a sample configuration, see ``router.lua`` and ``storage.lua`` in the
-``example/`` directory of the `vshard repository <https://github.com/tarantool/vshard>`__.
+
+.. _vshard-quick-start-working-adding-data:
+
+Adding data
```
Totktonada: Shouldn't we add a last section like 'access data', where we show an example of put/get requests to the router? It is OK to perform it from a console connected to the router.

Totktonada (approved):

doc/how-to/vshard_quick.rst
andreyaksenov: Agree, thanks

doc/how-to/vshard_quick.rst
andreyaksenov: Good point, will fix

doc/how-to/vshard_quick.rst
andreyaksenov: Yep, this is absolutely necessary. Totally forgot about this, will do.

andreyaksenov: Two sections instead of `Adding data`:
- [Adding and selecting data](https://docs.d.tarantool.io/en/doc/sharding-get-started/how-to/vshard_quick/#adding-and-selecting-data)
- [Checking data distribution](https://docs.d.tarantool.io/en/doc/sharding-get-started/how-to/vshard_quick/#checking-data-distribution)

Totktonada: Thanks for the update!

doc/how-to/vshard_quick.rst
```diff
@@ -1,160 +1,525 @@
...
 
-* ``router_1`` – a ``router`` instance
-* ``storage_1_a`` – a ``storage`` instance, the **master** of the **first** replica set
-* ``storage_1_b`` – a ``storage`` instance, the **replica** of the **first** replica set
-* ``storage_2_a`` – a ``storage`` instance, the **master** of the **second** replica set
-* ``storage_2_b`` – a ``storage`` instance, the **replica** of the **second** replica set
+The cluster created in this tutorial includes 5 instances: one router and 4 storages, which constitute two replica sets.
 
-All instances are managed using the :ref:`tt <tt-cli>` administrative utility.
+.. image:: /book/admin/admin_instances_dev.png
```
p7nov, line 13: It's better to make a separate copy for this tutorial. If we change the pic in the admin guide, we'll definitely forget about this usage.

doc/how-to/vshard_quick.rst
```diff
@@ -1,160 +1,525 @@
...
-:ref:`public API <vshard_api_reference-router_public_api>`:
+..  _vshard-quick-start-creating-app:
 
-..  code-block:: tarantoolsession
+Creating a cluster application
+------------------------------
 
-    unix/:./data/router_1.control> vshard.router.info()
+The :ref:`tt create <tt-create>` command can be used to create an application from a predefined or custom template.
+For example, the built-in ``vshard_cluster`` template enables you to create a ready-to-run sharded cluster application.
```
p7nov, line 40: I'd make this an admonition for better visibility. And use a stronger wording without `for example`: something like `tt provides a built-in template vshard_router, which enables`

doc/how-to/vshard_quick.rst
```diff
@@ -1,160 +1,525 @@
...
+
+In this tutorial, the application layout is prepared manually:
+
+1.  Create a tt environment in the current directory by executing the :ref:`tt init <tt-init>` command.
+
+2.  Inside the ``instances.enabled`` directory of the created tt environment, create the ``sharded_cluster`` directory.
+
+3.  Inside ``instances.enabled/sharded_cluster``, create the following files:
```
p7nov, line 48: Maybe say explicitly that they are empty for now? I've started clicking the links in file descriptions to find out about their content before reading the sentence after the list.

doc/how-to/vshard_quick.rst
```diff
@@ -1,160 +1,525 @@
...
+
+    -   ``instances.yml`` specifies instances to run in the current environment.
```
p7nov, line 50: The list look a bit inconsistent
- Articles: with and without `the` in the beginning
- Verb form: `specifies` and `is intended to store`

doc/how-to/vshard_quick.rst
```diff
@@ -1,160 +1,525 @@
...
+    -   The ``config.yaml`` file is intended to store the cluster's :ref:`configuration <configuration_overview>`.
+    -   ``storage.lua`` is intended to store code specific for :ref:`storages <vshard-architecture-storage>`.
+    -   ``router.lua`` is intended to store code specific for a :ref:`router <vshard-architecture-router>`.
+    -   ``sharded_cluster-scm-1.rockspec`` includes external dependencies required by the application.
```
p7nov: `specifies` looks more accurate here. I mean, you declare a name, not download the module here :)

p7nov: `specifies application dependencies` is also shorter and pretty complete)

doc/how-to/vshard_quick.rst
```diff
@@ -1,160 +1,525 @@
...
+
+    The next :ref:`vshard-quick-start-developing-app` section shows how to configure the cluster and write code specific for a router and storages.
```
p7nov: `write code specific for a router and storages` - I feel like more conceptual meaning would be nice here. Not just specific code for this and that, but code that enables automatic distribution of data/calls/whatever based on (what?).

doc/how-to/vshard_quick.rst
```diff
@@ -1,160 +1,525 @@
...
+
+Add the :ref:`credentials <configuration_reference_credentials>` configuration section:
+
+..  literalinclude:: /code_snippets/snippets/sharding/instances.enabled/sharded_cluster/config.yaml
+    :language: yaml
+    :start-at: credentials:
+    :end-at: roles: [sharding]
+    :dedent:
+
+In this section, two users with the specified passwords are created:
```
p7nov, line 98: Maybe say that these are purely technical/internal users? readers can think that these are users they'll use for connecting and executing some calls in the console.

doc/how-to/vshard_quick.rst
```diff
@@ -1,160 +1,525 @@
...
+
+*   The ``replicator`` user with the ``replication`` role.
+*   The ``storage`` user with the ``sharding`` role.
+
+..  WARNING::
```
p7nov: This isn't a warning. Warnings should notify about potentially dangerous consequences

doc/how-to/vshard_quick.rst
```diff
@@ -1,160 +1,525 @@
...
+In this section, the following options are configured:
+
+*   ``iproto.advertise.peer`` specifies how to advertise the current instance to other cluster members.
+    In particular, this option informs other replica set members that the ``replicator`` user should be used to connect to the current instance.
+*   ``iproto.advertise.sharding`` specifies how to advertise the current instance to a router and rebalancer.
+
+
+..  _vshard-quick-start-configuring-cluster-bucket-count:
+
+Step 3: Configuring bucket count
```
p7nov, line 134: Shoudn't we introduce buckets somehow or provide a link? They just come out if the blue, never introduced before.

doc/how-to/vshard_quick.rst
```diff
@@ -1,160 +1,525 @@
...
+
+    *   ``app``: The ``app.module`` option specifies that code specific to a router should be loaded from the ``router`` module. This is explained below in the :ref:`vshard-quick-start-router-code` section.
+    *   ``sharding``: The ``sharding.roles`` option specifies that an instance inside this group acts as a router.
+    *   ``replicasets``: This section configures one replica set with one router instance.
+
+
+Resulting configuration
+***********************
+
+The resulting cluster configuration should look as follows:
```
p7nov: Maybe add the name `config.yaml` name in this sentence or the example itself for those readers who just scroll and quickly scan the page?

doc/how-to/vshard_quick.rst
```diff
@@ -1,160 +1,525 @@
...
+
+1.  Open the ``router.lua`` file and load the ``vshard`` module as follows:
+
+    ..  literalinclude:: /code_snippets/snippets/sharding/instances.enabled/sharded_cluster/router.lua
+        :language: lua
+        :start-at: local vshard
+        :end-at: local vshard
+        :dedent:
+
+2.  Define the ``put`` function used to write data to a storage:
```
p7nov: I'd like more context here: say explicitly that this function defines how the router selects the storage to write the data. And the same for reading.

doc/how-to/vshard_quick.rst
```diff
@@ -1,160 +1,525 @@
...
+The ``dependencies`` section includes the specified version of the ``vshard`` module.
+To install dependencies, you need to :ref:`build the application <vshard-quick-start-building-app>`.
+
+
+.. _vshard-quick-start-building-app:
+
+Building the application
+------------------------
+
+In the terminal, open a directory where the :ref:`tt environment is created <vshard-quick-start-creating-app>`.
```
p7nov: `directory where the tt environment is created` > `the tt environment directory`?

doc/how-to/vshard_quick.rst
```diff
@@ -1,160 +1,525 @@
...
+       • Starting an instance [sharded_cluster:storage-b-002]...
+       • Starting an instance [sharded_cluster:router-a-001]...
+
+
+.. _vshard-quick-start-working-bootstrap:
+
+Bootstrapping a cluster
+~~~~~~~~~~~~~~~~~~~~~~~
+
+To bootstrap the cluster, follow the steps below:
```
p7nov: Maybe add an intro sentence that bootstrap is required? It's not obvious.

doc/how-to/vshard_quick.rst
```diff
@@ -1,160 +1,525 @@
...
+        - true
+        ...
+
+
+.. _vshard-quick-start-working-status:
+
+Checking status
+~~~~~~~~~~~~~~~
+
+To check the cluster's status, execute :ref:`vshard.router.info() <router_api-info>` on the router:
```
p7nov, line 398: Any hints what to look at? Now it's a big unfamiliar console output that hardly helps newcomers.

p7nov (changes requested):
Please see some comments from my side.

doc/how-to/vshard_quick.rst
andreyaksenov: Currently, the `Application environment` topic explicitly says that the `sharded_cluster` is used as a demo app:

image

So, having a separate copy of this image requires updating both images in a case the application is updated. I agree that finding usages of this image becomes complicated, but looks that this is another issue. IMO, it is better to store all images in one directory in our future docs repo, so the path will be the same for all usages.

doc/how-to/vshard_quick.rst
andreyaksenov: Changed to `Inside the empty ``instances.enabled`` directory`

doc/how-to/vshard_quick.rst
andreyaksenov: Updated

doc/how-to/vshard_quick.rst
andreyaksenov: Changed to `specifies external dependencies` because Tarantool also includes "internal" dependencies.

stores code -> contains code

doc/how-to/vshard_quick.rst
andreyaksenov: Looks like this sounds better:

```
The next :ref:`vshard-quick-start-developing-app` section shows how to 
configure the cluster and write code for routing read and write requests to different storages.
```

doc/how-to/vshard_quick.rst
andreyaksenov: I feel that `technical/internal` doesn't exactly convey the purpose of these users (but can't come up with better wording). Let's add one sentence after the list:

```
These users are intended to maintain replication and sharding in the cluster.
```

doc/how-to/vshard_quick.rst
andreyaksenov: Rephrased to:

```
It is not recommended to store passwords as plain text in a YAML configuration.
Learn how to load passwords ...
```

Not sure that we should clarify the exact consequences here.

doc/how-to/vshard_quick.rst
andreyaksenov: Added a link to https://www.tarantool.io/en/doc/latest/concepts/sharding/vshard_architecture/#virtual-buckets

doc/how-to/vshard_quick.rst
andreyaksenov: Changed to:

```
The resulting ``config.yaml`` file should look as follows:
```

doc/how-to/vshard_quick.rst
andreyaksenov: > this function defines how the router selects the storage to write the data

Not exactly, it does two things:
1) Calculates a bucket ID used to write data to the correct storage.
2) Uses this bucket ID to write data to the storage.

This is what the list after the function definition describes. I'd keep the first sentence short because too much context before the function definition makes it hard to read because of too much technical details.

doc/how-to/vshard_quick.rst
andreyaksenov: Not sure - all the steps in this tutorial are required :) I've created a separate section for such a simple operation to emphasise this.

doc/how-to/vshard_quick.rst
andreyaksenov: Added the description for each top-level section

doc/how-to/vshard_quick.rst
andreyaksenov: Rephrased:

```
After starting instances, you need to bootstrap the cluster as follows:
```

doc/how-to/vshard_quick.rst
p7nov: I still don't particularly like the `for example` wording. It's a definite an recommended way to create an exact app structure, not an random example.

doc/how-to/vshard_quick.rst
p7nov: I mean creating empty files. The emptiness of instances directory is pretty obvious :)
When I see an instruction to create a file, I expect next steps with its content right away. Now we're leaving them empty, so it's better to say that it's ok for now, the content will appear later.

doc/how-to/vshard_quick.rst
p7nov: I mean, the admonition type can be a note, tip, or something similar. We're just informing that there are more secure options for real life, nothing to **warn** about.

doc/how-to/vshard_quick.rst
```diff
@@ -1,160 +1,535 @@
...
+*   ``replicasets``: contains information about storages and their availability.
+*   ``bucket``: displays the total number of read-write and read-only buckets that are currently available for this router.
+*   ``status``: the number from 0 to 3 that indicates whether there are any issues with the cluster.
+    0 means that there are no issues.
+*   ``alerts``: might describe the exact issues related to bootstrapping a cluster, for example, connection issues, failover events, or unidentified buckets.
+
+
+.. _vshard-quick-start-working-adding-selecting-data:
+
+Adding and selecting data
```
p7nov:
```suggestion
Writing and selecting data
```

p7nov (approved):

PR #4044, opened 2024-02-14
Update EE changelogs

xuniq:
https://github.com/tarantool/sdk/tags
Added: 605, 606, 609, 613, 616

Staging:


andreyaksenov (approved):
LGTM, but we need to think about adding 3.0 SDKs to this page (in particular, `*-r23.linux.x86_64.tar.gz`)

Issue #4081, opened 2024-03-07
[Config] Rework the main 'Configuration' topic


andreyaksenov:
**Related doc issue(s):** https://github.com/tarantool/doc/issues/4190

**Product:** Tarantool
**Since:** 3.0/3.1
**Root document:** https://www.tarantool.io/en/doc/latest/concepts/configuration/
**SME:** @ Totktonada

# Details

Below are some things that can be improved:

- ~~Remove the [Configuration options overview](https://www.tarantool.io/en/doc/latest/concepts/configuration/#configuration-options-overview) section after fixing https://github.com/tarantool/doc/issues/4012.~~
- Update the structure of the `Configuration` section:
   - Cluster configuration
     - Centralized configuration storages
     - Configuration in code (legacy)
- Emphasize the difference between two configuration approaches: in a file and config storage. For example, the diagrams from [this topic](https://www.tarantool.io/en/doc/latest/concepts/configuration/configuration_etcd/) can be moved to the main topic.
- Describe some examples of maintaining consistency in a file config: network folder, external tools for delivering the config file to different machines.
- Consider adding diagrams that help a reader understand main concepts, for example:
  - Config and one instance
  - Config and several instances on one machine
  - Config and several instances on different machines
  - Other ideas: https://github.com/tarantool/doc/issues/3849#issuecomment-1809456024
- Add information about the config module API: https://github.com/tarantool/doc/issues/3662.
- Take Roles into account: https://github.com/tarantool/doc/issues/3702.

## Possible topic structure

- Configuration sources
  - Local
  - Сentralized
  Env vars.
- Configuring a cluster topology
  - Instance
  - Replica set
  - Group
  - Global
  - Scopes (merging)
- Developing applications
   - app/app_cfg
   - roles/roles_cfg
- Configuration values
  - Value type (string, number, array, dynamic (sync quorum), ...)
  - Default values
  - Predefined variables
- Environment variables
- Loading config from external storage (`config.context`)
- Configuration precedence
- Reloading configuration

Issue #3920, opened 2023-12-08
Update the 'Developing applications' section to using a new approach to develop apps


andreyaksenov:
**TODO:** Make one more iteration after fixing https://github.com/tarantool/doc/issues/3817
**Related doc issues:** https://github.com/tarantool/doc/issues/3518, https://github.com/tarantool/doc/issues/3179

**Product:** Tarantool
**Since:** 3.0
**Root document:** https://www.tarantool.io/en/doc/latest/how-to/app/
**SME:** @ Totktonada, @ filonenko-mikhail

## Details

The entire section should be updated to using a new approach to develop apps. Possible section structure:

https://docs.google.com/document/d/1El_8nAjguzijVkOijmt7v1Jt615lirkSUqZIcG5nBzE/edit?usp=sharing

This structure is based on the tt usage scenarios and the [Developer’s guide](https://www.tarantool.io/en/doc/2.11/book/cartridge/cartridge_dev/) for Cartridge.

PR #4221, opened 2024-05-23
Docs for roles/role_cfg

### LINKED ISSUE #3670: [Config] Document 'app' configuration settings
**Related doc issue:** https://github.com/tarantool/doc/issues/3702

**Product:** Tarantool
**Since:** 3.0
**Root document:** 
- A new `app` section in the [Configuration reference](https://www.tarantool.io/en/doc/latest/reference/configuration/configuration_reference/).
- https://www.tarantool.io/en/doc/latest/concepts/configuration/#loading-an-application
- Mention [roles](https://www.tarantool.io/en/doc/latest/reference/configuration/configuration_reference/#roles)

**SME:** @ Totktonada

# Details

Options:
- https://github.com/tarantool/tarantool/blob/f58bfc97c627f1838696697339fb3e443c05767a/src/box/lua/config/instance_config.lua#L1630

### LINKED ISSUE #3702: Roles
**Related dev. issue(s):** https://github.com/tarantool/tarantool/issues/9078
**Related doc. issue(s):** https://github.com/tarantool/doc/issues/3883, https://github.com/tarantool/doc/issues/3670

**Product:** Tarantool
**Since:** 3.0
**Root document:** 
- [roles reference](https://www.tarantool.io/en/doc/latest/reference/configuration/configuration_reference/#configuration-reference-roles-options)
- [Tarantool config storage](https://www.tarantool.io/en/doc/latest/concepts/configuration/configuration_etcd/#tarantool-based-storage)
- Somewhere in [Developing applications](https://github.com/tarantool/doc/issues/3920)
- https://github.com/tarantool/doc/issues/4152

**SME:** @ ImeevMA

# Details

Two new options have been added: "roles" and "roles_cfg". The first one
is an array and the second one is a map. Each of these can be defined
per instance, replica set, group, and globally. As with almost all other
options, with the exception of those defined as 'map', the 'roles'
option for the lower scope will replace the roles for the higher scope.
Value roles_cfg however defined as "map", so it will be merged.

The "roles" option defines the roles for each instance. A role is a
program that runs when a configuration is loaded or reloaded. If a role
is defined more than once on an instance, it will still only be run
once. Three functions must be defined in the role: validate(), apply()
and stop(). Each of these functions should throw an error if it occurs.

The "roles_cfg" option specifies the configuration for each role. In
this option, the role name is the key and the role configuration is the
value.

On each run, all roles will be loaded (if necessary) in the order in
which they were specified; the configuration for each role will then be
validated using the corresponding validate() function in the same order;
and then they will all be run with apply() function in the same order.
If some roles have been removed from the instance, they will be stopped
in reverse order using the stop() function.

Example of a role structure:
```lua
local M = {}

-- Validates configuration of the role.
--
-- Called on initial configuration apply at startup and on
-- configuration reload if the role is enabled for the given instance.
--
-- The cfg argument may have arbitrary user provided value,
-- including nil.
--
-- Must raise an error if the validation fail.
function M.validate(cfg)
    -- <...>
end

-- Applies the given configuration of the role.
--
-- Called on initial configuration apply at startup and on
-- configuration reload if the role is enabled for the given instance.
--
-- The cfg argument may have arbitrary user provided value,
-- including nil.
--
-- Must raise an error if the given configuration can't be applied.
function M.apply(cfg)
    -- <...>
end

-- Stops the role.
--
-- Called on configuration reload if the role was enabled before
-- and removed now from the list of roles of the given instance.
--
-- Should cancel all background fibers and clean up hold
-- resources.
--
-- Must raise an error if this action can't be performed.
function M.stop()
    -- <...>
end

return M
```
Requested by @ ImeevMA in https://github.com/tarantool/tarantool/commit/52884400993d1bca1c490e3a739f5528ce06ed3b.

# Example

Config:

```yaml
credentials:
  users:
    guest:
      roles: [super]

app:
  file: 'init.lua'

groups:
  routers:
    roles_cfg:
      router:
        init_size: 100
    roles: [router]
    replicasets:
      router-001:
        sharding:
          roles: [router]
        instances:
          instance-000:
            iproto:
              listen: 'localhost:3300'
              advertise:
                client: 'localhost:3300'
                sharding: 'localhost:3300'
  storages:
    roles: [storage]
    replicasets:
      storage-001:
        database:
          replicaset_uuid: 'aaaaaaaa-0000-4000-a000-000000000000'
        instances:
          instance-001:
            database:
              instance_uuid: 'aaaaaaaa-0000-4000-a000-000000000011'
              mode: rw
            sharding:
              roles: [storage]
            iproto:
              listen: 'localhost:3301'
              advertise:
                client: 'localhost:3301'
                sharding: 'localhost:3301'
          instance-002:
            database:
              instance_uuid: 'aaaaaaaa-0000-4000-a000-000000000012'
            sharding:
              roles: [storage]
            iproto:
              listen: 'localhost:3302'
              advertise:
                client: 'localhost:3302'
                sharding: 'localhost:3302'
      storage-002:
        database:
          replicaset_uuid: 'bbbbbbbb-0000-4000-a000-000000000000'
        instances:
          instance-003:
            database:
              instance_uuid: 'bbbbbbbb-0000-4000-a000-000000000021'
              mode: rw
            sharding:
              roles: [storage]
            iproto:
              listen: 'localhost:3303'
              advertise:
                client: 'localhost:3303'
                sharding: 'localhost:3303'
          instance-004:
            database:
              instance_uuid: 'bbbbbbbb-0000-4000-a000-000000000022'
            sharding:
              roles: [storage]
            iproto:
              listen: 'localhost:3304'
              advertise:
                client: 'localhost:3304'
                sharding: 'localhost:3304'
```

`init.lua`:

```lua
local fiber = require("fiber")
require('log').info("lua: example started")

local version = require('tarantool').version

assert(version:sub(1, 2) == '3.', "tarantool version have to be >= 3, got " .. version)

_G.hello = function(name)
    return "Hello "..tostring(name)
end


fiber.create(function()
    box.ctl.wait_rw(30)
    box.schema.func.create('prometheus_collect_http', {if_not_exists=true})
    box.schema.func.create('hello', {if_not_exists=true})
end)
```

Router code:

```lua
local vshard = require('vshard')
local crud = require('crud')
local uuid = require('uuid')
local datetime = require('datetime')
local log = require('log')
local fiber = require('fiber')

local M = {}

function M.validate(cfg)
    if cfg.init_size then
        assert(type(cfg.init_size) == "number", "init_size must be number")
        assert(cfg.init_size > 0, "init_size must be positive")
    end
end

function M.bootstrap_f(cfg)
    for i = 1,10 do
        local rc, err = vshard.router.bootstrap()
        log.info("Attempt to bootstrap vshard cluster")
        log.info(rc)
        if rc == nil then
            if string.find(tostring(err), "is already bootstrapped") ~=nil then
                break
            end
            log.info(err)
        end
        if rc ~= nil then
            break
        end
        fiber.sleep(1)
    end
    if cfg.init_size then
        for i = 1, cfg.init_size do
            crud.insert('sales', {
                uuid.new(),
                box.NULL,
                ("product_%s"):format(i % 100),
                ("buyer_%s"):format(i % 10),
                i,
                true,
                datetime.new{timestamp=os.time()}
            })
        end


        for i = 1, cfg.init_size do
            crud.replace('counter', {
                i,
                box.NULL,
                i,
                i,
            })
        end
        for i = 1, cfg.init_size do 
            crud.replace('complex', {
                i, 
                "BBBB", 
                box.NULL, 
                false, 
                false, i, 30, i + 1.1111,
                {"string1",  "string2",  "string3" }, 
                "Something", 
                "Category", 
                uuid.new(),
                "Key2", 
                datetime.new{timestamp=os.time()}
            })
        end
    end
end



function M.apply(cfg)
    M.bootstrap_fiber = fiber.create(M.bootstrap_f, cfg)
    
    crud.init_router()
end

function M.stop()
    crud.stop_router()
end

return M
```

Storage code:

```lua
local crud = require('crud')

local M = {}

function M.validate()
end

function M.apply()
    crud.init_storage()
    if box.info.ro ~= true then
        box.schema.space.create('sales', {if_not_exists=true})

        box.space.sales:format({
            {name='id', type='uuid'},
            {name='bucket_id', type='unsigned'},
            {name='product_id', type='string'},
            {name='buyer_id', type='string'},
            {name='price', type='double'},
            {name='direct', type='boolean'},
            {name='time', type='datetime'},
        })
        box.space.sales:create_index('primary',{ parts = {{'id'}}, if_not_exists=true})
        box.space.sales:create_index('bucket_id', {
            parts={{'bucket_id'}}, unique=false, if_not_exists=true,
        })


        box.schema.space.create('counter', {if_not_exists=true})

        box.space.counter:format({
            {name='id1', type='unsigned'},
            {name='bucket_id', type='unsigned'},
            {name='id2', type='unsigned'},
            {name='id3', type='unsigned'},
        })
        box.space.counter:create_index('primary',{ parts = {{'id1'}}, if_not_exists=true})
        box.space.counter:create_index('bucket_id', {
            parts={{'bucket_id'}}, unique=false, if_not_exists=true,
        })


        box.schema.space.create('complex', {if_not_exists=true})
        box.space.complex:format({
            {name='key1', type='integer'},
            {name='val', type='string'},
            {name="bucket_id", type="unsigned"},
            {name='flag', type='boolean'},
            {name='flag sort', type='boolean', is_nullable=true},
            {name='integer_number', type='integer'},
            {name='age', type='unsigned'},
            {name='int_or_float', type='number'},
            {name="arr", type='array'},
            {name='something', type='scalar'},
            {name='minstr', type='string'},
            {name='id', type='uuid'},
            {name='key2', type='string'},
            {name='time', type='datetime'},
        })
        box.space.complex:create_index('pkey', {parts={{field='key1', type='integer'}, {field='key2', type='string'}}, if_not_exists=true})
    
        box.space.complex:create_index('secondary', {parts={{field='key1', type='integer'}, {field='flag', type='boolean'}, {field='int_or_float', type='number'}}, if_not_exists=true})
    
        box.space.complex:create_index('otherindex', {parts={{field='something', type='scalar'}, {field='flag', type='boolean'}, {field='int_or_float', type='number'}}, if_not_exists=true})
    
        box.space.complex:create_index('ageindex', {parts={{field='something', type='scalar'}}, unique=false, if_not_exists=true})
    
        box.space.complex:create_index('agefifers', {parts={{field='something', type='scalar'}, {field='integer_number', type='integer'}}, if_not_exists=true})
    
        box.space.complex:create_index('nullableTrue', {parts={{field='something', type='scalar'}, {field='flag sort', type='boolean', is_nullable=true} }, if_not_exists=true, unique=false})
    
        box.space.complex:create_index('ageArray', {parts={{field='arr', type="string", path="[*]"}, {field='integer_number', type='integer'}}, if_not_exists=true, unique=false})
    
        box.space.complex:create_index('bucket_id', {
            parts={{'bucket_id'}}, unique=false, if_not_exists=true,
        })
    end
end

function M.stop()
    crud.stop_storage()
end

return M
```

### LINKED ISSUE #3883: dependencies for roles
**Related dev. issue(s):** https://github.com/tarantool/tarantool/issues/9078
**Related doc. issue(s):** https://github.com/tarantool/doc/issues/3702
**Related Cartridge docs:** [Defining role dependencies](https://www.tarantool.io/en/doc/2.11/book/cartridge/cartridge_dev/#defining-role-dependencies)

**Product:** Tarantool
**Since:** 3.0
**Root document:** 
- [Tarantool config storage](https://github.com/tarantool/enterprise_doc/issues/265)

**SME:** @ ImeevMA

# Details

Roles can now have dependencies. This means that the verify() and
apply() methods will be executed for these roles, taking into account
the dependencies. Dependencies should be written in the "dependencies"
field of the array type. Note, the roles will be loaded (not applied!)
in the same order in which they were specified, i.e. not taking
dependencies into account.

Roles dependencies are specified in a role's code - not in a config:

```lua

return {
    dependencies = {'four', 'two'},
    validate = function() end,
    apply = apply,
    stop = function() end,
}
```

Example:

Dependencies of role A: B, C
Dependencies of role B: D
No other role has dependencies.

Order in which roles were given: [E, C, A, B, D, G]
They will be loaded in the same order: [E, C, A, B, D, G]
The order, in which functions verify() and apply() will be executed:
[E, C, D, B, A, G].
Requested by @ ImeevMA in https://github.com/tarantool/tarantool/commit/7fbc1d040e0e0eb6008f138c82c3ee1b0846d651.

### LINKED ISSUE #4152: Update CRUD sample to using built-in roles
**Related dev. issue(s):** https://github.com/tarantool/crud/issues/415

**Product:** Tarantool, crud
**Since:** Tarantool 3.1, [crud 1.5.0+](https://github.com/tarantool/crud/releases)
**SME:** @ DifferentialOrange

# Details

Sample: https://github.com/tarantool/doc/tree/latest/doc/code_snippets/snippets/sharding/instances.enabled/sharded_cluster_crud

CRUD docs: https://github.com/tarantool/crud?tab=readme-ov-file#tarantool-3-roles

andreyaksenov:
Docs for roles.

- Created several samples for demonstrating built-in and custom roles: 
   - Updated the `sharded_cluster_crud` sample to using built-in CRUD roles.
   - Added a `hello world` role without config.
   - Added a `hello world` role with config.
   - Added a sample that demonstrates how to create an HTTP API role.
- Updated the main `Configuration` topic:
   - Replaced a section about `app` with roles: [Enabling and configuring roles](https://docs.d.tarantool.io/en/doc/roles-samples/concepts/configuration/#enabling-and-configuring-roles)
   - Added a note about option precedence for different scopes: [Configuration scopes](https://docs.d.tarantool.io/en/doc/roles-samples/concepts/configuration/#configuration-scopes)
   - Updated the `TT_APP_CFG` example to `TT_ROLES_CFG`: [Environment variables | Map](https://docs.d.tarantool.io/en/doc/roles-samples/concepts/configuration/#map)
   - Removed the `Configuration options overview` section. Existing topics already cover this info.
- Created a new [Application roles](https://docs.d.tarantool.io/en/doc/roles-samples/how-to/app/app_roles/) topic inside `Developing applications with Tarantool`.
- Config reference:
   - `roles`/`roles_cfg` options: [roles](https://docs.d.tarantool.io/en/doc/roles-samples/reference/configuration/configuration_reference/#roles)
   - `app` options: [app](https://docs.d.tarantool.io/en/doc/roles-samples/reference/configuration/configuration_reference/#app)


ImeevMA (approved):

doc/reference/configuration/configuration_reference.rst
```diff
@@ -2877,6 +2878,120 @@ This section describes configuration parameters related to roles.
...
 
+    Specify the roles of an instance.
+    To specify a role's configuration, use the :ref:`roles_cfg <configuration_reference_roles_cfg>` option.
+
+    There are two types of application roles:
+
+    -   Built-in roles. For example, the `CRUD <https://github.com/tarantool/crud>`__ module provides the ``roles.crud-storage`` and ``roles.crud-router`` roles that enable CRUD operations in a sharded cluster.
+    -   Custom roles. For example, you create a custom role to define a stored procedure or implement a supplementary service, such as an e-mail notifier or a replicator.
+
+    To create a custom role, you need to define the following methods and fields:
```
Totktonada: Nit: I would use the term 'function' instead of 'method', because these functions **don't** accept `self` as a first argument.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -2877,6 +2878,120 @@ This section describes configuration parameters related to roles.
...
+
+    -   ``validate()``
+
+        Validate a role's configuration.
+        This method is called when the initial configuration is applied or the configuration is reloaded for the instance with this role.
+        ``validate()`` should throw an error if the validation fails.
```
Totktonada: 'When the initial configuration is applied' means the same what 'on instance startup' means, but the latter looks simpler.

First time I read it, my thought was about a first `require` of the module (because I thought about a role's configuration, not the whole configuration).

doc/reference/configuration/configuration_reference.rst
```diff
@@ -2877,6 +2878,120 @@ This section describes configuration parameters related to roles.
...
+    To specify a role's configuration, use the :ref:`roles_cfg <configuration_reference_roles_cfg>` option.
+
+    There are two types of application roles:
+
+    -   Built-in roles. For example, the `CRUD <https://github.com/tarantool/crud>`__ module provides the ``roles.crud-storage`` and ``roles.crud-router`` roles that enable CRUD operations in a sharded cluster.
+    -   Custom roles. For example, you create a custom role to define a stored procedure or implement a supplementary service, such as an e-mail notifier or a replicator.
+
+    To create a custom role, you need to define the following methods and fields:
+
+    -   ``validate()``
```
Totktonada: The idea of this function is to provide a validation feedback earlier, so all the `validate` functions for all the enabled roles are executed first and only after that the `apply()` functions are executed. I think that it worth to mention it, because otherwise it is not clear why not to just call some `myvalidate()` directly from `apply()`.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -2877,6 +2878,120 @@ This section describes configuration parameters related to roles.
...
+
+        Validate a role's configuration.
+        This method is called when the initial configuration is applied or the configuration is reloaded for the instance with this role.
+        ``validate()`` should throw an error if the validation fails.
+
+    -   ``apply()``
+
+        Apply a role's configuration.
+        This method is called when the initial configuration is applied or the configuration is reloaded for the instance with this role.
+        ``apply()`` should throw an error if the specified configuration can't be applied.
```
Totktonada: Maybe it worth to highlight that it doesn't matter, whether the role's configuration or any field in the whole configuration is changed. If a configuration reload occurs (due to changing of a remote storage configuration revision or because of manual `config:reload()` call), all the `validate` and `apply` functions are called for the enabled roles.

If some action should be taken once during data lifetime or once during process lifetime, a role author should care about it.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -2877,6 +2878,120 @@ This section describes configuration parameters related to roles.
...
+
+    -   ``stop()``
+
+        Stop a role.
+        This method is called on configuration reload if the role is removed from ``roles`` for the given instance.
+        ``stop()`` should cancel all background fibers and clean up used resources.
+        If these actions can't be performed, ``stop()`` should throw an error.
```
Totktonada: I'm not sure that we should push role authors to throw an error from `stop()`.

In my understanding, we should describe the contract, so the author can choose.

Something like the following:

> All these functions (`validate`, `apply` and `stop`) report an unrecoverable error by throwing an error object (usually just a string). An error in any phase of configuration applying means stop of the configuration applying. If role's stop or start raises an error, no roles are stopped or started atferwards. The error is catched and saved. It is shown in alerts reported by the `config:info()` call.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -2877,6 +2878,120 @@ This section describes configuration parameters related to roles.
...
+        This method is called when the initial configuration is applied or the configuration is reloaded for the instance with this role.
+        ``validate()`` should throw an error if the validation fails.
+
+    -   ``apply()``
+
+        Apply a role's configuration.
+        This method is called when the initial configuration is applied or the configuration is reloaded for the instance with this role.
+        ``apply()`` should throw an error if the specified configuration can't be applied.
+
+    -   ``stop()``
```
Totktonada: I think that it worth to mention that all the `<role>.stop()` calls are performed before any the `<role>.validate()` or `<role>.apply()` calls.

So, old roles are stopped first and only then new roles are started.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -2877,6 +2878,120 @@ This section describes configuration parameters related to roles.
...
+
+        Stop a role.
+        This method is called on configuration reload if the role is removed from ``roles`` for the given instance.
+        ``stop()`` should cancel all background fibers and clean up used resources.
+        If these actions can't be performed, ``stop()`` should throw an error.
+
+    -   (Optional) ``dependencies``
+
+        Define a role's dependencies as an array.
+        For roles that depend on each other, their ``validate()`` and ``apply()`` methods are executed taking into account the dependencies.
```
Totktonada: We also guarantee config's order for independent roles. See details [here](https://github.com/tarantool/tarantool/pull/9087#discussion_r1317938242) (or ask me for details if this comment is too algorithmic :)).

Totktonada: What is possibly more important than the order: we forbid to start a role without its dependencies. All the dependencies of a role should be in `roles`, otherwise an error is reported. There is no automatic loading of roles that are not listed in the configuration explicitly (it is a design decision).

doc/reference/configuration/configuration_reference.rst
```diff
@@ -2861,11 +2861,12 @@ The ``replication`` section defines configuration parameters related to :ref:`re
 roles
 -----
 
-This section describes configuration parameters related to roles.
+This section describes configuration parameters related to application roles.
+An application role is a Lua module that implements specific functions or logic.
```
Totktonada: While it is true that a role implements specific functions or logic, it doesn't differentiate the term 'role' from just 'module'.

If we say that a role is a managed code that can be enabled or disabled (started/stopped), it is closer to the idea of roles.

It is also important to say that a role is a Lua module and that role's name is a value passed to `require()` to obtain the module (unlike cartridge, where a role defines its own name in the code, which is different from module's name).

doc/reference/configuration/configuration_reference.rst
```diff
@@ -2861,11 +2861,12 @@ The ``replication`` section defines configuration parameters related to :ref:`re
...
 
 .. NOTE::
 
-    Configuration parameters related to roles can be defined in any :ref:`scope <configuration_scopes>`.
+    Configuration parameters related to application roles can be defined in any :ref:`scope <configuration_scopes>`.
```
Totktonada: While the configuration merging rules are simple and the same for all the sections, it seems important to highlight them here in context of transforming a cluster configuration (with global/group/replicaset/instance options) to an instance configuration.

It is the point, where behavior is not very intuitive.

First, the rules:

1. Records and maps are deeply merged.
2. Arrays and scalars are *chosen*, not merged. (Note: `any` is a scalar even if its value is a table.)

Instance's option has highest precedence, then replicaset's ones, then group's ones, then globals.

See details about the merging algorithm [here](https://github.com/tarantool/tarantool/blob/8de22969aeeed85428aefb4d608a38c86cd3c3ea/src/box/lua/config/utils/schema.lua#L1299-L1337).

Back to roles.

`roles` is an array. It means that if there is `roles: [foo]` on the global level and `roles: [bar]` on the instance's level, the result is `roles: [bar]`. Not `roles: [foo, bar]`!

It also means that if `roles: [foo, bar, baz]` are defined in globals, in a group or in a replicaset and we need to disable `bar` for a specific instance `i-001` (seems quite usual admin's task), we need to list all the remaining roles for `i-001`: `roles: [foo, baz]`.

Something similar with `roles_cfg`: it is `map(string -> any)`. So, `roles_cfg: {foo: {myfoo: 5}}` + `roles_cfg: {bar: {mybar: 6}}` gives `roles_cfg: {foo: {myfoo: 5},  {bar: {mybar: 6}}`, but `roles_cfg: {foo: {myfoo: 5}}` + `roles_cfg: {foo: {myfoo_2: 6}}` gives `roles_cfg: {foo: {myfoo_2: 6}}`. `any` is a scalar.

I would like to discuss how to make things better (I have some ideas around), but now it is important to document constraints and peculiars of the existing implementation.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -2877,6 +2878,120 @@ This section describes configuration parameters related to roles.
 
     **Since:** :doc:`3.0.0 </release/3.0.0>`.
 
+    Specify the roles of an instance.
+    To specify a role's configuration, use the :ref:`roles_cfg <configuration_reference_roles_cfg>` option.
+
+    There are two types of application roles:
+
+    -   Built-in roles. For example, the `CRUD <https://github.com/tarantool/crud>`__ module provides the ``roles.crud-storage`` and ``roles.crud-router`` roles that enable CRUD operations in a sharded cluster.
+    -   Custom roles. For example, you create a custom role to define a stored procedure or implement a supplementary service, such as an e-mail notifier or a replicator.
```
Totktonada: AFAIK, we use the term 'built-in' for things that are built into tarantool executable. `crud` is an external module in this sense.

I would draw the line in a bit different way.

* Roles that are shipped by a Lua rock or provided by tarantool itself. The idea here is to write them once and use in different applications.
* Roles that are part of an application. The idea is to highlight that the given code  is a kind of service and, also, give a way to enable/disable this specific code for an administrator.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -2877,6 +2878,120 @@ This section describes configuration parameters related to roles.
...
+
+        ..  literalinclude:: /code_snippets/snippets/sharding/instances.enabled/sharded_cluster_crud/config.yaml
+            :language: yaml
+            :start-at: storages
+            :end-at: roles.crud-storage
+            :dedent:
+
+    You can find the full example here: `sharded_cluster_crud <https://github.com/tarantool/doc/tree/latest/doc/code_snippets/snippets/sharding/instances.enabled/sharded_cluster_crud>`_.
+
+    **Example: Custom role**
```
Totktonada: For me, it feels beyond the configuration reference scope. Can we create a separate 'how to write a role' document and link it from here?

Totktonada: Questions that I was asked by early adopters:

* Where to place an initialization code (cartridge's `<role>.init()`)?[^1]
* How to create a space?[^2]

It is beyond the reference scope, but these questions are important and we definitely should highlight them in examples/recipes/howtos.

[^1]: On the top level of a module. One may wrap it to a function and call on the top level:

      ```lua
      local M = {}

      local function init()
          <...>
      end

      function M.validate(cfg)
          <...>
      end

      function M.apply(cfg)
          <...>
      end

      function M.stop()
          <...>
      end

      init()

      return M
      ```
[^2]: We should do it on an RW instance (using a fiber plus `box.ctl.wait_rw()` or using `box.watch('box.status', <...>)` or using `box.ctl.on_election()`, but the latter is applicable only for `replication.failover: election`).

      Note: `apply()` is not invoked if an instance goes to RW as part of a Raft election or appointed by a failover coordinator. Unlike cartridge, there is no `is_leader` property in the `apply` function, its purpose is different. If some actions should be performed, when an instance goes to RW, a code should catch **this** event, not a reconfiguration.

      Note 2: But the schema initialization code (that creates a space) should be called from `apply` too, because a role may be enabled, when an instance is already RW.

      We should create a space once, so `if_not_exists` should be used (or `box.once`, but there were some problems around it with timeouts and error handling).

doc/reference/configuration/configuration_reference.rst
```diff
@@ -2877,6 +2878,120 @@ This section describes configuration parameters related to roles.
...
+
+    ..  literalinclude:: /code_snippets/snippets/config/instances.enabled/application_role/greeter.lua
+        :language: lua
+        :dedent:
+
+    You can find the full example here: `application_role <https://github.com/tarantool/doc/tree/latest/doc/code_snippets/snippets/config/instances.enabled/application_role>`_.
+
+    **Example: Role dependency**
+
+    The example below implements the ``greeter`` role:
```
Totktonada: Just idea: how about somethings that creates an HTTP+JSON endpoint as an example? It feels much more interesting than logging to a file, while it can be written in a short way using the `http` module[^1].

We anyway lean on the `crud` module in this document, so why not?

[^1]: If you'll go this way, you may find cartridge's hack for removing routes useful. See https://github.com/tarantool/http/issues/150.

doc/code_snippets/snippets/config/instances.enabled/application_role/greeter.lua
```diff
@@ -0,0 +1,21 @@
...
+    end
+end
+
+function M.apply(cfg)
+    log.info("%s from the 'greeter' role!", cfg.greeting)
+end
+
+function M.stop(cfg)
+    log.info("The 'greeter' role is stopped")
+end
```
Totktonada: AFAIR, `stop()` doesn't accept a configuration as an argument.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -2889,6 +3004,40 @@ This section describes configuration parameters related to roles.
...
+        :end-at: greeting
+        :dedent:
+
+    You can define the ``validate()``, ``apply()``, and ``stop()`` functions in the role's code as follows:
+
+    ..  literalinclude:: /code_snippets/snippets/config/instances.enabled/application_role/greeter.lua
+        :language: lua
+        :dedent:
+
+    You can find the full example here: `application_role <https://github.com/tarantool/doc/tree/latest/doc/code_snippets/snippets/config/instances.enabled/application_role>`_.
```
Totktonada: It seems, it duplicates the same examples from the `roles` description. The size of this block is close to a viewport height in my case, so I feels this duplication as undesirable.

I would highlight here my idea to move these examples to a 'how to write a role' guide and link it from here.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -2889,6 +3004,40 @@ This section describes configuration parameters related to roles.
 
     **Since:** :doc:`3.0.0 </release/3.0.0>`.
 
+    Specify a role's configuration.
+    This option accepts a role name as the key and a role's configuration as the value.
+    To specify the roles of an instance, use the :ref:`roles <configuration_reference_roles>` option.
```
Totktonada, line 3106: We can also say that `validate()` and `apply()` functions accept **this** value[^1] as an argument.

We can also say that this argument is for convenience, but any value from a currently applying configuration may be received from `validate()` and `apply()` using the  `config:get(<...>)` call.

We can also say that a user is not obligated to use only `roles_cfg.<role name>` to store role's configuration. It is not a mistake if `roles_cfg` has a key that doesn't correspond to any existing role name. A role may read anything from `roles_cfg` or from the configuration at all.

However, `roles_cfg.<role name>` is good default choice.

[^1]: To be precise: `roles_cfg.<role name>`.

Totktonada, line 3106: I've described peculiars of the globals->group->replicaset->instance options merging above. It is relevant for `roles_cfg` as well: the behavior is not very intuitive here. We should fairly describe it.

doc/reference/configuration/configuration_reference.rst
andreyaksenov: Extended the description of the `apply` function:

```
    -   ``apply()``

        Apply a role's configuration.
        ``apply()`` is called after ``validate()`` is executed for all the enabled roles.
        As the ``validate()`` function, ``apply()`` is called on instance startup or when the configuration is reloaded for the instance with this role.
```

doc/reference/configuration/configuration_reference.rst
andreyaksenov: Added this info and also added the link to the [Reloading configuration](https://www.tarantool.io/en/doc/latest/concepts/configuration/configuration_etcd/#reloading-configuration) section to clarify this:

```
        This function is called on instance startup or when the :ref:`configuration is reloaded <etcd_reloading_configuration>` for the instance with this role.
        Note that the ``validate()`` function is called regardless of the role's configuration or any field in the whole configuration is changed.
```

Note that the `Reloading configuration` section will be moved to the main Configuration topic at some moment (https://github.com/tarantool/doc/issues/4081) because it is applied to a file config, too.

doc/reference/configuration/configuration_reference.rst
andreyaksenov: If I'm correct, adding this note should be enough for now:

```
Note that independent roles are executed in the order they defined in a configuration.
```

doc/reference/configuration/configuration_reference.rst
andreyaksenov: As a temporary workaround, this example will be placed in the configuration reference but outside the `roles/roles_cfg` descriptions. As some moment, all such content will be moved to the `Developing applications` section (https://github.com/tarantool/doc/issues/3920).

doc/reference/configuration/configuration_reference.rst
Totktonada: I've updated the comment, it was a bit incorrect regarding disabling of a role.

Totktonada: The last bullet in https://docs.d.tarantool.io/en/doc/roles-samples/concepts/configuration/#roles-and-configuration-scopes doesn't correspond tarantool's current behavior.

We consider changing it from the tarantool side. But until it I think that the documentation should match tarantool's behavior.

Leaving the discussion open.

Totktonada: Now the description matches tarantool's behavior. Closing.

Totktonada (approved):
Thank you for the decent work!

doc/concepts/configuration.rst
```diff
@@ -171,55 +178,128 @@ You can learn more about configuring replication from :ref:`Replication tutorial
 
 
 .. _configuration_application:
+.. _configuration_application_roles:
+
+Enabling and configuring roles
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+An application role is a Lua module that implements specific functions or logic.
+You can enable or turn off a particular role for certain instances in a configuration without restarting these instances.
```
p7nov: "enable or turn off" sounds inconsistent. Let's use one word for the two states (`enable/disable` or `turn on/off`)

doc/concepts/configuration.rst
```diff
@@ -171,55 +178,128 @@ You can learn more about configuring replication from :ref:`Replication tutorial
...
 
-Loading an application
-**********************
+This section describes how to enable and configure roles.
+To learn how to develop custom roles, see :ref:`application_roles`.
```
p7nov, line 191: You write about custom roles, but there there is no explanation of _non custom_ roles. Maybe add a sentence about them to the first paragraph?

doc/concepts/configuration.rst
```diff
@@ -171,55 +178,128 @@ You can learn more about configuring replication from :ref:`Replication tutorial
...
+
+.. _configuration_application_roles_scopes:
+
+Roles and configuration scopes
+******************************
+
+As the most of configuration options, roles and their configurations can be defined at :ref:`different levels <configuration_scopes>`.
+Given that the ``roles`` option has the ``array`` type and ``roles_cfg`` has the ``map`` type, there are some specifics of applying the configuration:
+
+-   For ``roles``, an instance's role takes precedence over roles defined at another level.
```
p7nov, line 248:
```suggestion
-   For ``roles``, an instance's role overwrites all roles defined at another levels.
```
Nit: `overwrites` seems a more straightforward verb to explain that other roles are discarded.

doc/how-to/app/app_roles.rst
```diff
@@ -0,0 +1,476 @@
...
+Application roles
+=================
+
+An application role is a Lua module that implements specific functions or logic.
+You can enable or turn off a particular role for certain instances in a :ref:`configuration <configuration>` without restarting these instances.
+A role is run when a configuration is loaded or reloaded.
+
+Roles can be divided into the following groups:
+
+-   Roles provided by Tarantool.
```
p7nov:
```suggestion
-   Tarantool's built-in roles.
```
Nit: this sounds more natural to me (if this wording is correct in our terminology).

doc/how-to/app/app_roles.rst
```diff
@@ -0,0 +1,476 @@
...
+
+This section describes how to develop custom roles.
+To learn how to enable and configure roles, see :ref:`configuration_application_roles`.
+
+..  NOTE::
+
+    Don't confuse application roles with other role types:
+
+    -   A role is a container for privileges that can be granted to users. Learn more in :ref:`access_control_concepts_roles`.
+    -   A role of a replica set in regard to sharding. Learn more in :ref:`vshard_config_sharding_roles`.
```
p7nov, line 27:
```suggestion
    -   A user role is a container for privileges that can be granted to users. Learn more in :ref:`access_control_concepts_roles`.
    -   A sharding role of a replica set in regard to sharding. Learn more in :ref:`vshard_config_sharding_roles`.
```
Just an idea, to decrease readers' confusion a bit

doc/how-to/app/app_roles.rst
```diff
@@ -0,0 +1,476 @@
...
+This example shows how to enable and configure the ``greeter`` role, which is implemented in the next section:
+
+..  literalinclude:: /code_snippets/snippets/config/instances.enabled/application_role_cfg/config.yaml
+    :language: yaml
+    :start-at: instance001
+    :dedent:
+
+The role's configuration provided in ``roles_cfg`` can be accessed when :ref:`validating <roles_create_custom_role_validate>` and :ref:`applying <roles_create_custom_role_apply>` this configuration.
+
+Given that a role is a :ref:`Lua module <app_server-modules>`, a role's name is passed to ``require()`` to obtain the module.
```
p7nov, line 49: I'm confused with passive voice here: does it explain what Tarantool does under the hood or instructs me to pass the role's name when I write the app code?

doc/how-to/app/app_roles.rst
```diff
@@ -0,0 +1,476 @@
...
+----------------------
+
+.. _roles_create_custom_role_overview:
+
+Overview
+~~~~~~~~
+
+Creating a custom role includes the following steps:
+
+1.  Define a function that validates a role's configuration.
```
p7nov, line 66: I'd provide the function names explicitly in this list.

doc/how-to/app/app_roles.rst
```diff
@@ -0,0 +1,476 @@
...
+
+
+
+.. _roles_create_custom_role_validate:
+
+Validating a role configuration
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To validate a role's configuration, you need to define the :ref:`validate([cfg]) <roles_api_reference_validate>` function.
+The ``cfg`` argument allows you to get access to the provided :ref:`role's configuration <roles_create_custom_role_config>` and check its validity.
```
p7nov:
```suggestion
The ``cfg`` argument provides access to the :ref:`role's configuration <roles_create_custom_role_config>` and check its validity.
```
Seems shorter and easier to read

doc/how-to/app/app_roles.rst
```diff
@@ -0,0 +1,476 @@
...
+
+For example, you can :ref:`create spaces <box_schema-space_create>`, define :ref:`indexes <concepts-data_model_indexes>`, or :ref:`grant privileges <configuration_credentials>` to specific users or roles.
+
+See also: :ref:`roles_create_space`.
+
+
+
+.. _roles_create_space:
+
+Specifics of creating spaces
```
p7nov, line 201:
```suggestion
Creating spaces in roles
```

doc/how-to/app/app_roles.rst
```diff
@@ -0,0 +1,476 @@
...
+    ..  NOTE::
+
+        Dependencies do not affect the order in which roles are loaded.
+        However, the ``validate()``, ``apply()``, and ``stop()`` functions are executed taking dependencies into account.
+        Learn more in :ref:`roles_life_cycle_dependencies_specifics`.
+
+
+..  _roles_life_cycle_stopping_roles:
+
+2)  *Stopping roles*
```
p7nov, line 249: Maybe `Spotting removed roles` or smth like that? the current wording surprised me and took some time to understand which roles we're talking about

doc/how-to/app/app_roles.rst
```diff
@@ -0,0 +1,476 @@
...
+
+.. _roles_life_cycle_dependencies_specifics:
+
+Executing functions for dependent roles
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+For roles that :ref:`depend <roles_create_custom_role_dependencies>` on each other, their ``validate()``,  ``apply()``, and ``stop()`` functions are executed taking into account the dependencies.
+Suppose, there are two independent roles (``role1``, ``role2``) and three roles that depend on each other as follows:
+
+..  code-block:: none
```
p7nov, line 282: Maybe add a verbal explanation? It's unclear to me now.

doc/how-to/app/app_roles.rst
```diff
@@ -0,0 +1,476 @@
...
+    Stop a role.
+    This function is called on configuration reload if the role is removed from ``roles`` for the given instance.
+
+    See also: :ref:`roles_create_custom_role_stop`
+
+.. _roles_api_reference_dependencies:
+
+..  data:: dependencies
+
+    (Optional) Define a role's dependencies.
```
p7nov, line 476:
```suggestion
    (Optional) Role's dependencies.
```
Nit: IMO, verbs are not needed in option/parameter/variable descriptions in references.

p7nov (approved):
Great addition, I finally understood what are these application roles :)

One general question: should we maybe provide more cross-links to doc pages where built-in roles and module roles are explained or used? I mean, you say `there are built-in roles`, and the reader wonders `What built-in roles? Is there a list? How to use them?`

doc/concepts/configuration.rst
andreyaksenov: Good point, updated the intro of this section to tell a bit about different types of roles.

doc/how-to/app/app_roles.rst
andreyaksenov: > A user role

Sounds better but a role can also has a role :) So, it's better to keep it as is.

> A sharding role of a replica set in regard to sharding.

`sharding` meets two times in this case and add some confusion.

doc/how-to/app/app_roles.rst
andreyaksenov: explains what Tarantool does under the hood. Maybe, this should be a note...

doc/how-to/app/app_roles.rst
andreyaksenov: This applies to creating spaces for applications in general. So, I'm going to move this section outside this topic at some moment.

andreyaksenov:
> One general question: should we maybe provide more cross-links to doc pages where built-in roles and module roles are explained or used? I mean, you say `there are built-in roles`, and the reader wonders `What built-in roles? Is there a list? How to use them?`

It's a good idea. AFAIK, Tarantool provides only one role for now. For roles provided by external modules, this question goes beyond roles. Currently, I know two external modules that provide its own roles: CRUD and expirationd (and the docs gives a link to CRUD as an example). At the same time, any information for these modules are missing in our docs. I'll create an issue to think how to deal with this.

doc/how-to/app/app_roles.rst
andreyaksenov: This list give a general idea about role's functions. The code snippet after this lists the exact names.

PR #4240, opened 2024-05-27
Config reference for console and process

lenkis:
Add/update reference for console and prices config sections

### DIFF
--- a/doc/reference/configuration/configuration_reference.rst
+++ b/doc/reference/configuration/configuration_reference.rst
@@ -756,7 +756,6 @@ Tarantool output can be written to a file or Unix socket.
     | Type: boolean
     | Default: true
     | Environment variable: TT_CONSOLE_ENABLED
-    | Dynamic: yes
 
 .. _config_console_socket:
 
@@ -771,7 +770,6 @@ Tarantool output can be written to a file or Unix socket.
     | Type: string
     | Default: 'var/run/{{ instance_name }}/tarantool.control'
     | Environment variable: TT_CONSOLE_SOCKET
-    | Dynamic: yes
 
 ..  _configuration_reference_credentials:
 
@@ -2446,10 +2444,9 @@ The ``process`` section defines configuration parameters of the Tarantool proces
     **Since:** :doc:`3.0.0 </release/3.0.0>`.
 
     Run the server as a background task.
-    The :ref:`pid_file <configuration_reference_process_pid_file>` parameter
-    must be non-null for this to work.
-
-    ??? another non-null param for this to work was `log` (TT_LOG). Now irrelevant?
+    The :ref:`process.pid_file <configuration_reference_process_pid_file>`
+    and :ref:`log <configuration_reference_log>`
+    parameters must be non-null for this to work.
 
     .. important::
 
@@ -2460,7 +2457,6 @@ The ``process`` section defines configuration parameters of the Tarantool proces
     | Type: boolean
     | Default: false
     | Environment variable: TT_PROCESS_BACKGROUND
-    | Dynamic: no
 
 .. _configuration_reference_process_coredump:
 
@@ -2474,7 +2470,6 @@ The ``process`` section defines configuration parameters of the Tarantool proces
     | Type: boolean
     | Default: false
     | Environment variable: TT_PROCESS_COREDUMP
-    | Dynamic: no
 
 .. _configuration_reference_process_title:
 
@@ -2483,9 +2478,10 @@ The ``process`` section defines configuration parameters of the Tarantool proces
     **Since:** :doc:`3.0.0 </release/3.0.0>`.
 
     Add the given string to the server's process title
-    (what’s shown in the COMMAND column for ``ps -ef`` and ``top -c`` commands).
+    (e.g. it's shown in the COMMAND column for the Linux commands
+    ``ps -ef`` and ``top -c``).
 
-    For example, if you set ``title=''``, :samp:`ps -ef` shows
+    For example, if you set ``title: ''``, :samp:`ps -ef` shows
     the Tarantool server process thus:
 
     .. code-block:: console
@@ -2493,7 +2489,7 @@ The ``process`` section defines configuration parameters of the Tarantool proces
         $ ps -ef | grep tarantool
         1000     14939 14188  1 10:53 pts/2    00:00:13 tarantool <running>
 
-    But if you set ``title='sessions'``, then the output looks like:
+    But if you set ``title: 'sessions'``, then the output looks like:
 
     .. code-block:: console
 
@@ -2504,7 +2500,6 @@ The ``process`` section defines configuration parameters of the Tarantool proces
     | Type: string
     | Default: 'tarantool - {{ instance_name }}'
     | Environment variable: TT_PROCESS_TITLE
-    | Dynamic: yes
 
 .. _configuration_reference_process_pid_file:
 
@@ -2520,7 +2515,6 @@ The ``process`` section defines configuration parameters of the Tarantool proces
     | Type: string
     | Default: 'var/run/{{ instance_name }}/tarantool.pid'
     | Environment variable: TT_PROCESS_PID_FILE
-    | Dynamic: no
 
 .. _configuration_reference_process_strip_core:
 
@@ -2536,7 +2530,6 @@ The ``process`` section defines configuration parameters of the Tarantool proces
     | Type: boolean
     | Default: true
     | Environment variable: TT_PROCESS_STRIP_CORE
-    | Dynamic: no
 
 .. _configuration_reference_process_username:
 
@@ -2550,7 +2543,6 @@ The ``process`` section defines configuration parameters of the Tarantool proces
     | Type: string
     | Default: box.NULL
     | Environment variable: TT_PROCESS_USERNAME
-    | Dynamic: no
 
 .. _configuration_reference_process_work_dir:
 
@@ -2562,13 +2554,12 @@ The ``process`` section defines configuration parameters of the Tarantool proces
     switches to ``process.work_dir`` with :manpage:`chdir(2)` after start. Can be
     relative to the current directory. If not specified, defaults to
     the current directory. Other directory parameters may be relative to ``process.work_dir``,
-    for example :ref:`directories for storing snapshots and write-ahead logs <configuration_options_directories>`
+    for example, :ref:`directories for storing snapshots and write-ahead logs <configuration_options_directories>`
 
     |
     | Type: string
     | Default: box.NULL
     | Environment variable: TT_PROCESS_WORK_DIR
-    | Dynamic: no
 
 ..  _configuration_reference_replication:
 
--- a/doc/reference/configuration/configuration_reference.rst
+++ b/doc/reference/configuration/configuration_reference.rst
@@ -2482,11 +2482,12 @@ The ``process`` section defines configuration parameters of the Tarantool proces
     (or set corresponding options in systemd's unit file)
     before running a Tarantool process to get core dumps.
     If ``process.coredump`` is enabled, Tarantool sets the corresponding
-    resource limit by itself and the administrator doesn't need to call
-    ``ulimit -c unlimited``.
+    resource limit by itself
+    and the administrator doesn't need to call ``ulimit -c unlimited``
+    (see :href:`man 3 setrlimit <https://man7.org/linux/man-pages/man3/setrlimit.3p.html>`).
 
-    This option also :href:`sets <https://man7.org/linux/man-pages/man3/setrlimit.3p.html>`
-    the state of the ``dumpable`` attribute, which is enabled by default,
+    This option also sets the state of the ``dumpable`` attribute,
+    which is enabled by default,
     but may be dropped in some circumstances (according to
     :href:`man 2 prctl <https://man7.org/linux/man-pages/man2/prctl.2.html>`, see PR_SET_DUMPABLE).
 
--- a/doc/reference/configuration/configuration_reference.rst
+++ b/doc/reference/configuration/configuration_reference.rst
@@ -735,7 +735,7 @@ This section describes options related to providing connection settings to a :re
 console
 -------
 
-Configure the administration console. A client to the console is ``tt connect``.
+Configure the administrative console. A client to the console is ``tt connect``.
 
 ..  NOTE::
 
@@ -744,36 +744,36 @@ Configure the administration console. A client to the console is ``tt connect``.
 * :ref:`console.enabled <config_console_enabled>`
 * :ref:`console.socket <config_console_socket>`
 
-.. _config_console_enabled:
+.. _configuration_reference_console_enabled:
 
 .. confval:: console.enabled
 
     **Since:** :doc:`3.0.0 </release/3.0.0>`.
 
     Whether to listen on the Unix socket provided in the
-    :ref:`console.socket <config_console_socket>` option.
+    :ref:`console.socket <configuration_reference_console_socket>` option.
 
-    If the option is set to false, the admininstration console is disabled.
+    If the option is set to ``false``, the administrative console is disabled.
 
     |
     | Type: boolean
     | Default: true
     | Environment variable: TT_CONSOLE_ENABLED
 
-.. _config_console_socket:
+.. _configuration_reference_console_socket:
 
 .. confval:: console.socket
 
     **Since:** :doc:`3.0.0 </release/3.0.0>`.
 
-    The Unix socket for administration console.
+    The Unix socket for the administrative console.
 
     Mind the following nuances:
 
     * Only a Unix domain socket is allowed. A TCP socket can't be configured this way.
     * ``console.socket`` is a file path, without any ``unix:`` or ``unix/:`` prefixes.
     * If the file path is a relative path, it is interpreted relative to
-      :ref:`process.work_dir <config_process_work_dir>`.
+      :ref:`process.work_dir <configuration_reference_process_work_dir>`.
 
     |
     | Type: string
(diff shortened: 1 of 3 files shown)


doc/reference/configuration/configuration_reference.rst
```diff
@@ -2400,7 +2443,13 @@ The ``process`` defines configuration parameters of the Tarantool process in the
 
 .. confval:: process.background
 
+    **Since:** :doc:`3.0.0 </release/3.0.0>`.
+
     Run the server as a background task.
+    The :ref:`pid_file <configuration_reference_process_pid_file>` parameter
+    must be non-null for this to work.
+
+    ??? another non-null param for this to work was `log` (TT_LOG). Now irrelevant?
```
andreyaksenov: Looks like [log](https://www.tarantool.io/en/doc/latest/reference/configuration/configuration_reference/#log) is required:

```
% tarantool --name instance001 --config instances.enabled/application/config.yaml 
started
'background' requires 'log' configuration option to be set
```

doc/reference/configuration/configuration_reference.rst
```diff
@@ -730,8 +730,48 @@ This section describes options related to providing connection settings to a :re
...
+
+    **Since:** :doc:`3.0.0 </release/3.0.0>`.
+
+    Whether to enable Tarantool output to a file or Unix socket.
+
+    |
+    | Type: boolean
+    | Default: true
+    | Environment variable: TT_CONSOLE_ENABLED
+    | Dynamic: yes
```
andreyaksenov: Let's remove `Dynamic` fields for all options documented in this PR. In a new configuration approach, we don't change options "dynamically" in the console but reload the configuration to use new config values.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -2458,28 +2536,39 @@ The ``process`` defines configuration parameters of the Tarantool process in the
...
 .. confval:: process.work_dir
 
-    A directory where Tarantool working files are stored.
+    **Since:** :doc:`3.0.0 </release/3.0.0>`.
+
+    A directory where database working files will be stored. The server instance
+    switches to ``process.work_dir`` with :manpage:`chdir(2)` after start. Can be
+    relative to the current directory. If not specified, defaults to
+    the current directory. Other directory parameters may be relative to ``process.work_dir``,
+    for example :ref:`directories for storing snapshots and write-ahead logs <configuration_options_directories>`
```
andreyaksenov:
```suggestion
    for example, :ref:`directories for storing snapshots and write-ahead logs <configuration_options_directories>`.
```

doc/reference/configuration/configuration_reference.rst
```diff
@@ -2411,45 +2460,74 @@ The ``process`` defines configuration parameters of the Tarantool process in the
...
 .. _configuration_reference_process_title:
 
 .. confval:: process.title
 
-    Sets the name of the server's process title (what’s shown in the COMMAND column for
-    ``ps -ef`` and ``top -c`` commands).
+    **Since:** :doc:`3.0.0 </release/3.0.0>`.
+
+    Add the given string to the server's process title
+    (what’s shown in the COMMAND column for ``ps -ef`` and ``top -c`` commands).
```
andreyaksenov: > what’s shown in the COMMAND column for ``ps -ef`` and ``top -c`` commands

I'd generalized this description a bit because this look

- There is no COMMAND column on macOS.
- The `-c` flag doesn't work on macOS for the `top` command.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -2411,45 +2460,74 @@ The ``process`` defines configuration parameters of the Tarantool process in the
...
+
+        $ ps -ef | grep tarantool
+        1000     14939 14188  1 10:53 pts/2    00:00:13 tarantool <running>
+
+    But if you set ``title='sessions'``, then the output looks like:
+
+    .. code-block:: console
+
+        $ ps -ef | grep tarantool
+        1000     14939 14188  1 10:53 pts/2    00:00:16 tarantool <running>: sessions
```
andreyaksenov: `ps -ef` shows the following info on my machine. Not sure that we need such output in the config reference - looks like they might differ on different platforms. IMO, configuration examples can be more useful here.

```
% ps -ef | grep tarantool
  503  4330  4328   0 10:20AM ttys000    0:00.52 tarantool <running>: sessio -n instance001 -c /Users/john.doe/PycharmProjects/doc/doc/code_snippets/snippets/config/instances.enabled/application/config.yaml
```

If we still want to show the output, we can tell which platform is used for this example.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -2411,45 +2460,74 @@ The ``process`` defines configuration parameters of the Tarantool process in the
...
+
+    For example, if you set ``title=''``, :samp:`ps -ef` shows
+    the Tarantool server process thus:
+
+    .. code-block:: console
+
+        $ ps -ef | grep tarantool
+        1000     14939 14188  1 10:53 pts/2    00:00:13 tarantool <running>
+
+    But if you set ``title='sessions'``, then the output looks like:
```
andreyaksenov: > ``title='sessions'``

Let's use YAML syntax for config examles:

```yaml
process:
  title: 'session'
```

andreyaksenov (changes requested):

doc/reference/configuration/configuration_reference.rst
lenkis: Thx

doc/reference/configuration/configuration_reference.rst
lenkis: Thx

doc/reference/configuration/configuration_reference.rst
lenkis: My suggestion:

    Add the given string to the server's process title
    (e.g. it's shown in the COMMAND column for the Linux commands
    ``ps -ef`` and ``top -c``).

doc/reference/configuration/configuration_reference.rst
lenkis: Yes, ps and top have Mac-specific nuances. I guess saying "Linux commands" above would be enough to avoid confusion.

andreyaksenov (approved):

doc/reference/configuration/configuration_reference.rst
```diff
@@ -730,8 +730,46 @@ This section describes options related to providing connection settings to a :re
     | Default: 3
     | Environment variable: TT_CONFIG_STORAGE_TIMEOUT
 
+..  _configuration_reference_console:
 
+console
+-------
+
+Tarantool output can be written to a file or Unix socket.
```
Totktonada: These two options (`console.{enabled,socket}`) are in fact a wrapper[^1] around the [console.listen](https://www.tarantool.io/en/doc/latest/reference/reference_lua/console/#console-listen) call with the following nuances:

* Only a Unix domain socket is allowed. A TCP socket can't be configured this way[^2].
* `console.socket` is a file path, without any `unix:` or `unix/:` prefixes[^3].
* If the file path is a relative path, it is interpreted relative to `process.work_dir` -- consistently with all the file paths in the declarative configuration.
* Enabled by default.

I don't like 'Tarantool output' in the current wording, because it is unclear what it is (logs, stored data, network response?). I think that 'a file or Unix socket' is incorrect, because we have only one file path here and it is a Unix socket -- one entity, not two different ones.

So, I propose to adjust the existing description with a goal of the admin socket. For example,

> Configure the administration console.
>
> `tt connect` is a client to the console.
>
> <..details, if needed..>

[^1]: The code is more or less straightforward, you can glance on it [here](https://github.com/tarantool/tarantool/blob/01915e6d34bbe6c6bdd5042f8a3d1f4333c08975/src/box/lua/config/applier/console.lua).
[^2]: `console.listen()` supports it, but we decided to forbid it due to security concerns. The admin console doesn't support authorization and always execute commands as the `admin` user (with all the permissions). Since there is no authorization, it shouldn't be accessible remotely. So, listening on a TCP port is forbidden.
[^3]: It is not very consistent with all the other listening socket options, but Mons asked for a simple interface here and I agree that if we support only a Unix socket here, there is no reason to obligate a user to write `unix:` or `unix:/`.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -730,8 +730,46 @@ This section describes options related to providing connection settings to a :re
...
+* :ref:`console.enabled <config_console_enabled>`
+* :ref:`console.socket <config_console_socket>`
 
+.. _config_console_enabled:
+
+.. confval:: console.enabled
+
+    **Since:** :doc:`3.0.0 </release/3.0.0>`.
+
+    Whether to enable Tarantool output to a file or Unix socket.
```
Totktonada: Whether to listen on the Unix socket provided in the `console.socket` option. If the option is set to `false`, the admin console is disabled.

See the comment above for details.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -730,8 +730,46 @@ This section describes options related to providing connection settings to a :re
...
+    | Environment variable: TT_CONSOLE_ENABLED
+
+.. _config_console_socket:
+
+.. confval:: console.socket
+
+    **Since:** :doc:`3.0.0 </release/3.0.0>`.
+
+    The file or Unix socket for Tarantool output.
+    The value is a file path, with or without ``unix:``.
```
Totktonada: 1. AFAIK this option doesn't allow `unix:` or `unix/:`.
2. I don't like 'Tarantool output' (it is vague) and I think that 'file or socket' is incorrect (see my first comment above).

doc/reference/configuration/configuration_reference.rst
```diff
@@ -2400,7 +2441,12 @@ The ``process`` defines configuration parameters of the Tarantool process in the
 
 .. confval:: process.background
 
+    **Since:** :doc:`3.0.0 </release/3.0.0>`.
+
     Run the server as a background task.
+    The :ref:`process.pid_file <configuration_reference_process_pid_file>`
+    and :ref:`log <configuration_reference_log>`
+    parameters must be non-null for this to work.
```
Totktonada: This sentence is correct in context of `box.cfg` options, but it is a bit imprecise in context of the options from the declarative configuration. Let's look on them separately.

`process.pid_file` has a default value (`var/run/{{ instance_name }}/tarantool.pid`). A missing value or `null` means that the effective value is the default value. We can assume that the PID file is always created and consider this requirement as always met. There is no reason to highlight it to a user.

If you see the opposite behavior, I would like to know details, because it is not what I expect.

`log` is a section and its options have defaults. It means that the effective `log` value can't be `null`.

However, that's true that we should configure logs to `file`, `pipe` or `syslog` (anything other than `stderr`). The `log.to` option is for that. Its default is `stderr` and it actually requires an attention from a user, when `process.background` is set to `true`.

The reason why logging to `stderr` can't be used with a daemonization is that a daemon process is detached from a terminal and it can't write to the terminal's stderr. See also [man 3 daemon](https://man7.org/linux/man-pages/man3/daemon.3.html)[^1].

[^1]: Tarantool implements this functionality by itself, but the idea is the same. We close terminal's stdin/stdout/stderr too.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -2400,7 +2441,12 @@ The ``process`` defines configuration parameters of the Tarantool process in the
 
 .. confval:: process.background
 
+    **Since:** :doc:`3.0.0 </release/3.0.0>`.
+
     Run the server as a background task.
```
Totktonada: We can call it 'daemonization' (daemon process, daemonized process and so on). This is a well know term.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -2412,11 +2458,13 @@ The ``process`` defines configuration parameters of the Tarantool process in the
...
     | Environment variable: TT_PROCESS_BACKGROUND
 
-
 .. _configuration_reference_process_coredump:
 
 .. confval:: process.coredump
 
+    **Since:** :doc:`3.0.0 </release/3.0.0>`.
+
+    Whether to create coredump files.
```
Totktonada: Just tested it. It actually works on my system, wow!

At least, it works if I just run tarantool without daemonization and without changing an effective user.

Usually an administrator needs to call `ulimit -c unlimited` before running tarantool process to get coredumps (or set corresponding options in systemd's unit file). If `process.coredump` is enabled, tarantool set the corresponding resource limit by itself and the admin don't need to call `ulimit -c unlimited`.

As I see from the [code](https://github.com/tarantool/tarantool/blob/01915e6d34bbe6c6bdd5042f8a3d1f4333c08975/src/main.cc#L381-L396), the option also sets the `dumpable` attribute, which is enabled by default, but may be dropped in some circumstances (according to [man 2 prctl](https://man7.org/linux/man-pages/man2/prctl.2.html), see `PR_SET_DUMPABLE`).

It is interesting that it is [marked](https://www.tarantool.io/en/doc/latest/reference/configuration/#cfg-basic-coredump-deprecated) as deprecated in the box.cfg reference. I don't know why. We found it useful and included into the new declarative configuration.

You can include some of these details into the option description. I think it may be useful.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -2427,8 +2475,26 @@ The ``process`` defines configuration parameters of the Tarantool process in the
...
+    (e.g. it's shown in the COMMAND column for the Linux commands
+    ``ps -ef`` and ``top -c``).
+
+    For example, if you set ``title: ''``, :samp:`ps -ef` shows
+    the Tarantool server process thus:
+
+    .. code-block:: console
+
+        $ ps -ef | grep tarantool
+        1000     14939 14188  1 10:53 pts/2    00:00:13 tarantool <running>
```
Totktonada: I think that an empty value is not the best example to place it first.

How about something like `myservice - {{ instance_name }}`?

doc/reference/configuration/configuration_reference.rst
```diff
@@ -2439,7 +2505,11 @@ The ``process`` defines configuration parameters of the Tarantool process in the
 
 .. confval:: process.pid_file
 
+    **Since:** :doc:`3.0.0 </release/3.0.0>`.
+
     Store the process id in this file.
+    Can be relative to :ref:`process.work_dir <configuration_reference_process_work_dir>`.
+    A typical value is “:file:`tarantool.pid`”.
```
Totktonada: I think that a typical value is the default (`var/run/{{ instance_name }}/tarantool.pid`) and we don't need to mention it specifically.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -2439,7 +2505,11 @@ The ``process`` defines configuration parameters of the Tarantool process in the
 
 .. confval:: process.pid_file
 
+    **Since:** :doc:`3.0.0 </release/3.0.0>`.
+
     Store the process id in this file.
+    Can be relative to :ref:`process.work_dir <configuration_reference_process_work_dir>`.
```
Totktonada: Nit: Technically speaking, the option may contain a relative file path and in this case it is interpreted as relative to `process.work_dir`.

It is a bit more precise comparing to 'can be relative to `process.work_dir`'.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -2474,7 +2548,13 @@ The ``process`` defines configuration parameters of the Tarantool process in the
...
 .. confval:: process.work_dir
 
-    A directory where Tarantool working files are stored.
+    **Since:** :doc:`3.0.0 </release/3.0.0>`.
+
+    A directory where database working files will be stored. The server instance
+    switches to ``process.work_dir`` with :manpage:`chdir(2)` after start. Can be
+    relative to the current directory. If not specified, defaults to
+    the current directory. Other directory parameters may be relative to ``process.work_dir``,
+    for example, :ref:`directories for storing snapshots and write-ahead logs <configuration_options_directories>`
```
Totktonada: A couple of nits.

> A directory where database working files will be stored.

Not only the database files, but also logs, a PID file, a console Unix socket and other files if an application generates them in the current directory. 'Tarantool working files' seems a good general term if we don't want to go into details here.

> Can be relative to the current directory.

If we agree to follow my proposal above about more precise speaking about relative paths, I would replace it with 'if set as a relative path, it is relative to the current working directory, from where tarantool is started'.

> Other directory parameters may be relative to ``process.work_dir`` <...>

To be precise: directory and file parameters.

If we agree to follow the more precise wording about relative paths, I would say 'other directory and file parameters are interpreted as relative to `process.work_dir` if set as relative paths' or something of this kind.

Don't know, if it is too formal or not.

Totktonada (commented):
@lenkis Welcome back!

Thank you for the updates!

I've added a couple of technical details: some of them may be useful for a reader, some are just for context. I would like if you look over them and decide.

I also wrote my opinion about the used terms and proposed other ones, where it looks appropriate. Here I would also lean on your view about proper balance between a formal language and the conciseness.

doc/reference/configuration/configuration_reference.rst
Totktonada: > It is interesting that it is [marked](https://www.tarantool.io/en/doc/latest/reference/configuration/#cfg-basic-coredump-deprecated) as deprecated in the box.cfg reference. I don't know why. We found it useful and included into the new declarative configuration.

I guess that we should undeprecate it in the box.cfg reference for consistency with the declarative configuration reference.

If we meet the problem that was the reason for the deprecation again, then we return here back. Now it seems working and useful.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -2427,9 +2500,19 @@ The ``process`` defines configuration parameters of the Tarantool process in the
...
-    Sets the name of the server's process title (what’s shown in the COMMAND column for
-    ``ps -ef`` and ``top -c`` commands).
+    **Since:** :doc:`3.0.0 </release/3.0.0>`.
+
+    Add the given string to the server's process title
+    (it is shown in the COMMAND column for the Linux commands
+    ``ps -ef`` and ``top -c``).
+
+    For example, if you set ``title: 'myservice - instance1'``,
+    :samp:`ps -ef` shows the Tarantool server process like this:
```
Totktonada: I suggest to use `title: myservice - {{ instance_name }}` instead of `title: myservice - instance1`, because it works for all the instance names. See the [predefined variables](https://www.tarantool.io/en/doc/latest/concepts/configuration/#predefined-variables) section.

Maybe it is better to show the option example as a separate block, don't inline it into the prose text.

```yaml
process:
  title: myservice - {{ instance_name }}
```

Totktonada (approved):
Thanks for the review fixes!

PR #4265, opened 2024-06-05
Supervised failover

### LINKED ISSUE #4249: `tt cluster failover` commands
**Product**: tt CLI
**Since**: 2.3.0
**Root document**: https://www.tarantool.io/en/doc/latest/reference/tooling/tt_cli/cluster/

**SME**: @ 0x501D  

## Details

This patch adds new commands:
* tt cluster failover switch - switch master instance
* tt cluster failover switch-status - show master switching status

This commands works only with etcd based failover configuration.

Closes #822
Requested by @0x501D in https://github.com/tarantool/tt/commit/43b9f7ea11b7f83902066cdc4c616e96d363fb39.

andreyaksenov:
1) Created a sample app with the `supervised` failover: [supervised_failover](https://github.com/tarantool/doc/tree/supervised-failover/doc/code_snippets/snippets/replication/instances.enabled/supervised_failover).
2) Created a new [Supervised failover](https://docs.d.tarantool.io/en/doc/supervised-failover/concepts/replication/supervised_failover/) topic.
3) Updated the `Configuration reference`:
   - Added a link to a new topic in the [replication.failover: supervised](https://docs.d.tarantool.io/en/doc/supervised-failover/reference/configuration/configuration_reference/#configuration-reference-replication-failover) description.
   - Added a new [failover](https://docs.d.tarantool.io/en/doc/supervised-failover/reference/configuration/configuration_reference/#failover) section.
   - Mentioned one more use case for [config.etcd.*](https://docs.d.tarantool.io/en/doc/supervised-failover/reference/configuration/configuration_reference/#config-etcd).
4) Added the [--failover](https://docs.d.tarantool.io/en/doc/supervised-failover/reference/tarantool_cli_options/#cmdoption-failover) option to the `tarantool` CLI reference. Note that this reference is moved from the existing [Starting and stopping instances](https://docs.d.tarantool.io/en/doc/supervised-failover/book/admin/start_stop_instance/) topic.
5) Added the `tt cluster failover` command reference:
   - A new [failover](https://docs.d.tarantool.io/en/doc/supervised-failover/reference/tooling/tt_cli/cluster/#failover) subsection in the `tt cluster` command reference.
   - New `timeout` and `wait` options (applicable to `tt cluster failover`): [Options](https://docs.d.tarantool.io/en/doc/supervised-failover/reference/tooling/tt_cli/cluster/#options)


doc/concepts/replication/supervised_failover.rst
```diff
@@ -0,0 +1,182 @@
...
+
+..  admonition:: Enterprise Edition
+    :class: fact
+
+    Supervised failover is supported by the `Enterprise Edition <https://www.tarantool.io/compare/>`_ only.
+
+**Example on GitHub**: `supervised_failover <https://github.com/tarantool/doc/tree/latest/doc/code_snippets/snippets/replication/instances.enabled/supervised_failover>`_
+
+Tarantool provides the ability to control leadership in a replica set using an external failover coordinator.
+A failover coordinator reads a cluster configuration from a file or an :ref:`etcd-based configuration storage <configuration_etcd>`, polls instances for their statuses, and appoints a leader for each replica set depending on availability and health or instances.
```
Totktonada: Typo: health or instances -> health of instances.

doc/concepts/replication/supervised_failover.rst
```diff
@@ -0,0 +1,182 @@
...
+To increase fault tolerance, you can run two or more failover coordinators and store their state in etcd.
```
Totktonada: The 'run <...> coordinators and store their state in etcd' sounds like some user action is needed to store something in etcd to run several coordinators (because 'run coordinators' is performed by user's hands and the phrase is 'run and store').

We can say something like 'an etcd cluster is needed for synchronization between the coordinators if several ones are run'.

doc/concepts/replication/supervised_failover.rst
```diff
@@ -0,0 +1,182 @@
...
+
+3.  :ref:`Start a failover coordinator <supervised_failover_start_coordinator>`.
+    You can start two or more failover coordinators to increase :ref:`fault tolerance <supervised_failover_overview_fault_tolerance>`.
+    In this case, one coordinator is active and others are passive.
+
+When a cluster and failover coordinators are up and running, a failover coordinator :ref:`appoints one instance to be a master <supervised_failover_overview_appoint_master>` if there is no master instance in a replica set.
+Then, the following events may occur:
+
+-   If a master instance fails, a failover coordinator performs an :ref:`automated failover <supervised_failover_overview_perform_failover>`.
+-   If an active failover coordinator fails, another coordinator becomes active and performs read-write appointments.
```
Totktonada: I use 'RW appointment' term internally, but here I would use 'the automatic failover' instead, because it is already known for a reader from the previous sentence.

doc/concepts/replication/supervised_failover.rst
```diff
@@ -0,0 +1,182 @@
...
+
+
+.. _supervised_failover_overview_appoint_master:
+
+Appointing a master instance
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A failover coordinator uses so-called appointments to switch an instance to read-write mode.
+For each appointment, there is a read-write deadline, which is renewed periodically to prevent split-brain.
```
Totktonada: 1. Regarding 'read-write deadline': maybe it would be better to say 'read-write mode deadline'? Without the 'mode' word it sounds like a deadline for a read or write operation.

Totktonada: > renewed periodically to prevent split-brain

This clause may mislead someone who is not deeply in context of such fencing algorithms.

Someone can wrongly interpret it as follows.

> If periodical renews are not performed, a split-brain situation occurs.

However, things are different.

If periodical renews are not performed, an instance goes to RO (when the deadline is reached). A knowledge about the deadline is used by the coordinator to implement automatic failover algorithm that doesn't lead to appearance of two RW instances in a replicaset at the same time (IOW, to don't step into the split-brain situation).

doc/concepts/replication/supervised_failover.rst
```diff
@@ -0,0 +1,182 @@
...
+
+A failover coordinator appoints a master instance as follows:
+
+-   If a replica set has no read-write instance, a failover coordinator appoints one to be a master instance.
+-   If a replica set has one read-write instance and there is no previous appointment, a failover coordinator creates an appointment for this instance and sets a read-write deadline.
+-   If a replica set has one read-write instance and there is a previous appointment, a failover coordinator waits for its deadline.
```
Totktonada: We can describe these algorithms in two ways.

1. Define logical components 'failover' and 'appointment' and describe theirs behavior separately.

   In this case, the 'failover' component just waits, while the 'appointment' component performs RW deadline renewals.
2. Describe a coordinator's behavior at whole.

   In this case we can't say that the coordinator just waits for the deadline. It renews the deadline in a cycle.

It seems that we implicitly assume 'the coordinator' and 'the failover component' behind the 'failover coordinator' words.

Totktonada: We miss the 'there is an appointment, but no RW instances' case. It sounds like an oxymoron, but if we say the same in a bit more expicit way -- 'there is an appointment, but no **alive** RW instances' -- things become clear.

I would separate the situations in the following way.

1. All the instances are alive and RO -> choose master -> create appointment.
2. There is an alive RW instance.
   * No appointment -> (in some cases we wait here until an appointment from a previously active coordinator expires) -> create appointment
   * Has appointment -> renew it in background
3. An appointed instance is unavailable -> wait for RW deadline (IOW, until the existing appointment expires) -> choose new master -> create appointment.

doc/concepts/replication/supervised_failover.rst
```diff
@@ -0,0 +1,182 @@
...
+
+
+.. _supervised_failover_overview_perform_failover:
+
+Performing a failover
+~~~~~~~~~~~~~~~~~~~~~
+
+A failover coordinator monitors the statuses of instances by sending requests to instances each ``probe_interval`` seconds.
+While an appointment for a master instance exists and a read-write deadline has not expired, a failover coordinator does nothing.
+If all attempts to renew the deadline are failed during some time interval, a leader switches to read-only mode.
```
Totktonada: Similarly to a remark above I wouldn't say that a coordinator does nothing if there is an appointed master. At least, it renews the appointment in background. Also, it monitors the cluster.

Technically speaking, the coordinator also renews its lock in etcd, renews etcd lease, renews monitoring information (`<...>/info/by-uuid`), polls for new commands in etcd.

doc/concepts/replication/supervised_failover.rst
```diff
@@ -0,0 +1,182 @@
...
+
+.. _supervised_failover_overview_fault_tolerance:
+
+Active and passive coordinators
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To increase fault tolerance, you can :ref:`run <supervised_failover_start_coordinator>` two or more failover coordinators.
+In this case, only one coordinator is active and used to control leadership in a replica set.
+Other coordinators are passive.
+Passive coordinators don't perform any read-write appointments and decline all the existing appointments.
```
Totktonada: There is no process of declining an appointment from another coordinators. A coordinator may discard its own appointment as part of the manual switchover process. However, it seems, it is not what we say about here. Possibly a mistake.

doc/reference/tooling/tt_cli/cluster.rst
```diff
@@ -271,6 +272,74 @@ to ``ro`` and reloads the configuration.
...
+
+
+.. _tt-cluster-failover-switch:
+
+switch
+~~~~~~
+
+.. code-block:: console
+
+    $ tt cluster failover switch URI INSTANCE_NAME [OPTION ...]
```
Totktonada: Maybe it worth to mention `-w` here explicitly, because this option is likely often used together with this command.

Totktonada (commented):
Thank you for the links in the pull request description. It is really convenient to look over the parts that actually needs a review.

I highlighted several points in the failover algorithm overview. I think it worth to discuss f2f how to better describe it in a consice way.

At least, I see that we can discuss the following points:

* How to better divide it into subsections.
* Should we divide the service into more-or-less simple blocks and describe them separately or describe how the service works at whole?
* Should we highlight which actions are performed by which actor (coordinator, instance) and when exactly (on appoint request, on switching to active mode and so on).

I would also like to provide some in-depth materials like ones that are placed in https://github.com/tarantool/enterprise_doc/issues/253 (plus pictures from my presentations). I open a pull request with some drafts, but I likely need your assistance to make it ready for the website.

doc/concepts/replication/supervised_failover.rst
andreyaksenov: Yep, bad phrasing. At least, "... run two or more failover coordinators that will store their state in etcd." sounds better. Will fix.

doc/concepts/replication/supervised_failover.rst
andreyaksenov: > I use 'RW appointment' term internally

Thanks for the notice, I also had doubts regarding this term.

doc/concepts/replication/supervised_failover.rst
andreyaksenov: I got your point, will rephrase.

doc/concepts/replication/supervised_failover.rst
andreyaksenov: > Define logical components 'failover' and 'appointment' and describe theirs behavior separately.

Good idea, thanks. Will think how to incorporate this to algorithm description.

Totktonada (approved):
Thank you for the concise description and the reference!

doc/reference/tarantool_cli_options.rst
```diff
@@ -0,0 +1,191 @@
+.. _tarantool_cli:
+.. _configuration_command_options:
+
+tarantool command-line options
```
p7nov, line 4: There is no introduction/definition of the `tarantool` executable.
I'd add a sentence (possibily a note) explaining that `tarantool` is an executable for running a single instance and there is `tt` that covers much more scenarios.

doc/reference/tooling/tt_cli/cluster.rst
```diff
@@ -271,6 +272,74 @@ to ``ro`` and reloads the configuration.
...
+
+``tt cluster failover switch`` appoints the specified instance to be a master.
+This command accepts the following arguments and options:
+
+-   ``URI``: A :ref:`URI <tt-cluster-uri>` of the cluster configuration storage.
```
p7nov: I found a possible issue that applies to the whole page.
In the `tt cluster` reference, the `URI` arguments refers to the _config storage URI_, while in the rest of the `tt` reference it's the _instance URI_. This can confuse readers. Perhaps we should change the argument name on this page to `CONFIG_URI` or smth like that.

doc/reference/tooling/tt_cli/cluster.rst
```diff
@@ -271,6 +272,74 @@ to ``ro`` and reloads the configuration.
...
+-   ``INSTANCE_NAME``: An instance name.
+-   ``[OPTION ...]``: :ref:`Options <tt-cluster-options>` to pass to the command.
+
+Example:
+
+.. code-block:: console
+
+    $ tt cluster failover switch http://localhost:2379/myapp storage-a-002
+    To check the switching status, run:
```
p7nov, line 315: Looks like the output needs an explanation. At first, I decided that it's a part of the `switch-status` reference pasted here by mistake.

doc/reference/tooling/tt_cli/cluster.rst
```diff
@@ -271,6 +272,74 @@ to ``ro`` and reloads the configuration.
...
+
+.. code-block:: console
+
+    $ tt cluster failover switch-status URI TASK_ID
+
+``tt cluster failover switch-status`` shows the status of switching a master instance.
+This command accepts the following arguments:
+
+-   ``URI``: A :ref:`URI <tt-cluster-uri>` of the cluster configuration storage.
+-   ``TASK_ID``: An identifier of the task used to switch a master instance.
```
p7nov: It's never introduced. Let's explain it here or above (see the comment on `failover switch` example).

doc/reference/tooling/tt_cli/cluster.rst
```diff
@@ -271,6 +272,74 @@ to ``ro`` and reloads the configuration.
...
+.. _tt-cluster-failover-switch-status:
+
+switch-status
+~~~~~~~~~~~~~
+
+.. code-block:: console
+
+    $ tt cluster failover switch-status URI TASK_ID
+
+``tt cluster failover switch-status`` shows the status of switching a master instance.
```
p7nov, line 331: What statuses do we have?

doc/reference/tooling/tt_cli/cluster.rst
```diff
@@ -378,12 +447,24 @@ Options
 
     Skip validation when publishing. Default: `false` (validation is enabled).
 
+..  option:: --t, --timeout UINT
```
p7nov:
```suggestion
..  option:: -t, --timeout UINT
```
`-t` with one hyphen?

p7nov (approved):
Checked the reference part. Looks good, just a couple comments.

doc/reference/tooling/tt_cli/cluster.rst
andreyaksenov: Renamed the `URI` argument to `CONFIG_URI`

doc/reference/tooling/tt_cli/cluster.rst
andreyaksenov: Not sure that we need to list possible statuses. Looks like the command output is useful and doesn't require additional explanations:

image

xuniq (approved):
Looks good to me!

Issue #4238, opened 2024-05-24
 Configuration status is shown in `box.info.config` now


TarantoolBot:
**Product:** Tarantool
**Since:** 3.2
**Root document:** https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_info/
**SME:** @ Totktonada

There is the `config:info([version])` method, but in order to access it
over iproto an application developer should add something like the
following into the application code:

```lua
_G.config = require('config')
```

It is not convenient, at least because it requires an attention from the
application developer and it can't be solved solely by an administrator.

Now, the `config:info('v2')` result is reported in the `config` field of
the `box.info` table. It is accessible over iproto if appropriate
privileges are granted for a calling user.

Requested by @Totktonada in https://github.com/tarantool/tarantool/commit/a1544d3bbc029c6fb2a148e580afe2b20e269b8d (https://github.com/tarantool/tarantool/issues/10044, https://github.com/tarantool/tarantool/pull/10051)

PR #4331, opened 2024-07-02
Document new 'config' API

### LINKED ISSUE #4094: Expansion of `config:info()`
**Related dev. issue(s):** https://github.com/tarantool/tarantool/issues/9780

**Product:** Tarantool
**Since:** 3.1
**Root document:** https://www.tarantool.io/en/doc/latest/reference/reference_lua/config/#config-api-reference-info
**SME:** @ ImeevMA

# Details

Before this patch, the `meta` field of the `config:info()` function
contained the meta data of the last load, even if it was unsuccessful
reload. This can be inconvenient because if the reload fails, we expect
that the last configuration that was successfully applied is used, but
its metadata was lost.

Now `config:info()` can take one argument — the version of the
information that should be returned by the function. If no arguments are
specified or the argument is the string `v1`, the return value has the
same structure as before. If the argument is the string `v2`, the `meta`
field in the return value has fields - `active` (meta for the last
successfully applied configuration) and `last` (meta of the last loaded
configuration). Currently only these values are accepted.
Requested by @ ImeevMA in https://github.com/tarantool/tarantool/commit/699775b565269961656244431bf5ea7f05ace4bf.

### LINKED ISSUE #4105: config: access configuration of other cluster members
**Related dev. issue(s):** https://github.com/tarantool/tarantool/issues/9680

**Product:** Tarantool
**Since:** 3.1
**Root document:** 
- https://www.tarantool.io/en/doc/latest/reference/reference_lua/config/
- See also: https://www.tarantool.io/en/doc/latest/release/3.1.0/#accessing-configuration-of-other-cluster-members

**SME:** @Totktonada

# Details

### `config:instances()`

Lists all instances of the cluster.

Returns a table of the following format.

```lua
{
    [<instance_name>] = {
        instance_name = <...>,
        replicaset_name = <...>,
        group_name = <...>,
    },
    <...>
}
```

If an action should be performed for each instance of the given cluster,
it can be written this way:

```lua
for instance_name in pairs(config:instances()) do
    action(instance_name)
end
```

If replicaset/group names matter, the instances may be filtered this
way:

```lua
-- Perform an action for all instances of the given replicaset.
for instance_name, def in pairs(config:instances()) do
    if def.replicaset_name == box.info.replicaset.name then
        action(instance_name)
    end
end
```

### `config:get(<...>, {instance = '<...>'})`

The new `instance` option of the `config:get()` method allows to
retrieve a configuration value of another instance from the given
cluster.

Example:

```lua
-- Collect all the enabled roles within the cluster.
local enabled_roles = {}
for instance_name in pairs(config:instances()) do
    local roles = config:get('roles', {instance = instance_name})
    for _, role in ipairs(roles) do
        enabled_roles[role] = true
    end
end
```

Note: There is a difference between a missing `instance` option and the
`instance` option that is equal to the given instance name. The former
returns an instance configuration taking into account instance
configuration sources (environment variables). The latter takes into
account only cluster configuration, so the environment variables are
ignored.

Requested by @ Totktonada in https://github.com/tarantool/tarantool/commit/30a9c1ca7e25d9393cca35551aee65f8f53ec9fe.

### LINKED ISSUE #4120: instance_uri() method for config module
**Related dev. issue(s):** https://github.com/tarantool/tarantool/issues/9842

**Product:** Tarantool
**Since:** 3.1
**Root document:** https://www.tarantool.io/en/doc/latest/reference/reference_lua/config/
**SME:** @ ImeevMA

# Details

The configuration module's `instance_uri()` method returns the URI of
the instance that is used to create a replicaset or sharding cluster.
This method takes two arguments:
1) `peer` or null to return the URI that is used to create the
   replicaset, or `sharding` to return the URI that is used to create
   the sharding cluster;
2) `options`, which only support `options.instance`, to specify the
   instance whose URI should be returned; by default, the URI of the
   instance on which the function is executed is returned.
Requested by @ ImeevMA in https://github.com/tarantool/tarantool/commit/1826ca7cdb4f04e36ce23be103b0b6c98f3b9a64.

andreyaksenov:
Updated reference for the `config` module. Fixed doc issues:

- https://github.com/tarantool/doc/issues/4105
- https://github.com/tarantool/doc/issues/4120
- https://github.com/tarantool/doc/issues/4094

Staging:

- A new `opts` argument in [config:get()](https://docs.d.tarantool.io/en/doc/config_module_new_api/reference/reference_lua/config/#lua-function.config.get) (in particular, `opts.instance`)
- A new `version` argument in [config:info()](https://docs.d.tarantool.io/en/doc/config_module_new_api/reference/reference_lua/config/#config-api-reference-info)
- A new [config:instance_uri()](https://docs.d.tarantool.io/en/doc/config_module_new_api/reference/reference_lua/config/#config-api-reference-instance-uri) function
- A new [config:instances()](https://docs.d.tarantool.io/en/doc/config_module_new_api/reference/reference_lua/config/#config-api-reference-instances) function

> [!NOTE]
> New code examples are created using the [sharded_cluster_crud_docker](https://github.com/andreyaksenov/sharded_cluster_crud_docker) demo application.

### DIFF
--- a/doc/reference/reference_lua/config.rst
+++ b/doc/reference/reference_lua/config.rst
@@ -8,7 +8,7 @@ Module config
 The ``config`` module provides the ability to work with an instance's configuration.
 For example, you can determine whether the current instance is up and running without errors after applying the :ref:`cluster's configuration <configuration_overview>`.
 
-By using the ``config.storage`` :ref:`role <configuration_reference_roles_options>`, you can set up a Tarantool-based :ref:`centralized configuration storage <configuration_etcd>` and interact with this storage using the ``config`` module API.
+By using the ``config.storage`` :ref:`role <configuration_application_roles>`, you can set up a Tarantool-based :ref:`centralized configuration storage <configuration_etcd>` and interact with this storage using the ``config`` module API.
 
 
 ..  _config_module_loading:
@@ -46,10 +46,16 @@ API Reference
             -
 
         *   -   :ref:`config:get() <config_api_reference_get>`
-            -   Get a configuration applied to the current instance
+            -   Get a configuration applied to the current or remote instance
 
         *   -   :ref:`config:info() <config_api_reference_info>`
             -   Get the current instance's state in regard to configuration
+
+        *   -   :ref:`config:instance_uri() <config_api_reference_instance_uri>`
+            -   Get a URI of the current or remote instance
+
+        *   -   :ref:`config:instances() <config_api_reference_instances>`
+            -   List all instances of the cluster
 
         *   -   :ref:`config:reload() <config_api_reference_reload>`
             -   Reload the current instance's configuration
@@ -83,12 +89,19 @@ config API
 
     .. _config_api_reference_get:
 
-    ..  method:: get([param])
+    ..  method:: get([param, opts])
 
-        Get a configuration applied to the current instance.
-        Optionally, you can pass a configuration option name to get its value.
+        Get a configuration applied to the current or remote instance.
+        Note that there is the following difference between getting a configuration for the current and remote instance:
+
+        -   For the current instance, ``get()`` returns an instance configuration considering :ref:`environment variables <configuration_environment_variable>`.
+        -   For a remote instance, ``get()`` only considers a cluster configuration and ignores environment variables.
 
         :param string param: a configuration option name
+        :param table opts: options to pass. The following options are available:
+
+                           -   ``instance`` (since :doc:`3.1.0 </release/3.1.0>`) -- the name of a remote instance whose configuration should be obtained
+
         :return: an instance configuration
 
         **Examples:**
@@ -104,7 +117,7 @@ config API
                 too_long_threshold: 0.5
                 top:
                   enabled: false
-              # other configuration values
+              # Other configuration values
               # ...
 
         This example shows how to get an ``iproto.listen`` option value:
@@ -121,9 +134,18 @@ config API
 
     .. _config_api_reference_info:
 
-    ..  method:: info()
+    ..  method:: info([version])
 
         Get the current instance's state in regard to configuration.
+
+        :param string version: (since :doc:`3.1.0 </release/3.1.0>`) the version of the information that should be returned. The ``version`` argument can be one of the following values:
+
+                               * ``v1`` (default): the ``meta`` field returned by ``info()`` includes information about the last loaded configuration
+                               * ``v2``: the ``meta`` field returned by ``info()`` includes two fields:
+
+                                   * the ``last`` field includes information about the last loaded configuration
+                                   * the ``active`` field includes information for the last successfully applied configuration
+
 
         :return: a table containing an instance's state
 
@@ -155,7 +177,7 @@ config API
                 ...
 
         *   In the example below, the instance's state is ``check_warnings``.
-            The ``alerts`` section informs that privileges to the ``books`` space for ``sampleuser`` cannot be granted because the ``books`` space has not created yet:
+            The ``alerts`` section informs that privileges to the ``books`` space for ``sampleuser`` cannot be granted because the ``books`` space has not been created yet:
 
             ..  code-block:: console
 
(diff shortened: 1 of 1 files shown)


doc/reference/reference_lua/config.rst
```diff
@@ -121,10 +134,19 @@ config API
 
     .. _config_api_reference_info:
 
-    ..  method:: info()
+    ..  method:: info([version])
```
Totktonada, line 137: I would update the examples below to use `v2` format. It is more explicit and I would recommend users to use it.

The `v2` format is also used in `box.info.config` (see #4238).

doc/reference/reference_lua/config.rst
```diff
@@ -205,6 +227,65 @@ config API
...
+        :param string uri_type: a URI type. There are the following URI types are supported:
+
+                                * ``peer`` -- a URI used to advertise the instance to other cluster members. See also: :ref:`iproto.advertise.peer <configuration_reference_iproto_advertise_peer>`.
+                                * ``sharding`` -- a URI used to advertise the current instance to a router and rebalancer. See also: :ref:`iproto.advertise.sharding <configuration_reference_iproto_advertise_sharding>`.
+
+        :param table opts: options to pass. The following options are available:
+
+                           -   ``instance`` -- the name of a remote instance whose URI should be obtained
+
+        :return: an instance URI
```
Totktonada: I agree that all the details how the URI is constructed are redundant here. I think I should create a section in the internals and link it from here. It's on me.

But I would add information about the fields in the returned table right into the reference: it is part of the method contract. Also, it worth to add a note that the return value is suitable to pass into `net_box.connect()` (otherwise it may look logical to extract the `uri` field from it before calling `net_box.connect()`).

doc/reference/reference_lua/config.rst
```diff
@@ -205,6 +227,65 @@ config API
...
+
+        ..  code-block:: lua
+
+            local config = require('config')
+            for instance_name in pairs(config:instances()) do
+                local connpool = require('experimental.connpool')
+                local conn = connpool.connect(instance_name)
+                local log = require('log')
+                log.info(string.format("Connection URI for '%s': %s:%s", instance_name, conn.host, conn.port))
+            end
```
Totktonada: Maybe it wotrh to add some example that uses the second iteration value: `for instance_name, def in pairs(config:instances()) do <...> end`?

#4105 has an example with filtering instances of the same replicaset.

doc/reference/reference_lua/config.rst
```diff
@@ -205,6 +227,65 @@ config API
...
+        The example below shows how to use ``instances()`` to get the names of all instances in the cluster, create a connection to each instance using the :ref:`connpool <connpool_module>` module, and log connection URIs using the :ref:`log <log-module>` module:
+
+        ..  code-block:: lua
+
+            local config = require('config')
+            for instance_name in pairs(config:instances()) do
+                local connpool = require('experimental.connpool')
+                local conn = connpool.connect(instance_name)
+                local log = require('log')
+                log.info(string.format("Connection URI for '%s': %s:%s", instance_name, conn.host, conn.port))
```
Totktonada: BTW, `%q` can be used instead of `"%s"` for quoting.

```lua
tarantool> log.info('xxx %q zzz', 'yyy')
xxx "yyy" zzz
```

And an explicit `string.format` is not needed, `log.info` calls it itself.

doc/reference/reference_lua/config.rst
```diff
@@ -205,6 +227,65 @@ config API
...
+
+        The example below shows how to use ``instances()`` to get the names of all instances in the cluster, create a connection to each instance using the :ref:`connpool <connpool_module>` module, and log connection URIs using the :ref:`log <log-module>` module:
+
+        ..  code-block:: lua
+
+            local config = require('config')
+            for instance_name in pairs(config:instances()) do
+                local connpool = require('experimental.connpool')
+                local conn = connpool.connect(instance_name)
+                local log = require('log')
```
Totktonada: Stylistic suggestion: I would move `require` calls to the top.

Totktonada (approved):

doc/reference/reference_lua/config.rst
Totktonada: Nice, thanks!

doc/reference/reference_lua/config.rst
```diff
@@ -83,19 +89,26 @@ config API
 
     .. _config_api_reference_get:
 
-    ..  method:: get([param])
+    ..  method:: get([param, opts])
 
-        Get a configuration applied to the current instance.
-        Optionally, you can pass a configuration option name to get its value.
+        Get a configuration applied to the current or remote instance.
+        Note that there is the following difference between getting a configuration for the current and remote instance:
```
lenkis:
```suggestion
        Note the following differences between getting a configuration for the current and remote instance:
```

doc/reference/reference_lua/config.rst
```diff
@@ -83,19 +89,26 @@ config API
...
+
+        -   For the current instance, ``get()`` returns an instance configuration considering :ref:`environment variables <configuration_environment_variable>`.
```
lenkis:
```suggestion
        -   For the current instance, ``get()`` returns instance configuration considering :ref:`environment variables <configuration_environment_variable>`.
```

doc/reference/reference_lua/config.rst
```diff
@@ -83,19 +89,26 @@ config API
...
+        -   For a remote instance, ``get()`` only considers a cluster configuration and ignores environment variables.
```
lenkis, line 98:
```suggestion
        -   For a remote instance, ``get()`` only considers cluster configuration and ignores environment variables.
```

doc/reference/reference_lua/config.rst
```diff
@@ -121,89 +134,204 @@ config API
...
+              meta:
+                last: &0 []
+                active: *0
+              alerts: []
+            ...
+
+        **Example: configuration warnings**
+
+        In the example below, the instance's state is ``check_warnings``.
+        The ``alerts`` section informs that privileges to the ``bands`` space for ``sampleuser`` cannot be granted because the ``bands`` space has not created yet:
```
lenkis:
```suggestion
        The ``alerts`` section informs that privileges to the ``bands`` space for ``sampleuser`` cannot be granted because the ``bands`` space has not been created yet:
```

doc/reference/reference_lua/config.rst
```diff
@@ -121,89 +134,204 @@ config API
...
+
+    .. _config_api_reference_instance_uri:
+
+    ..  method:: instance_uri([uri_type, opts])
+
+        **Since:** :doc:`3.1.0 </release/3.1.0>`
+
+        Get a URI of the current or remote instance.
+
+        :param string uri_type: a URI type. There are the following URI types are supported:
```
lenkis:
```suggestion
        :param string uri_type: a URI type. The following URI types are supported:
```

doc/reference/reference_lua/config.rst
```diff
@@ -121,89 +134,204 @@ config API
...
-                - status: check_errors
-                  meta: []
-                  alerts:
-                  - type: error
-                    message: '[cluster_config] log.level: Got 8, but only the following values are
-                      allowed: 0, fatal, 1, syserror, 2, error, 3, crit, 4, warn, 5, info, 6, verbose,
-                      7, debug'
-                    timestamp: 2024-02-29T12:55:54.366810+0300
-                ...
+        In this example, the same actions are performed for instances from the specific replica set:
```
lenkis:
```suggestion
        In this example, the same actions are performed for instances from a specific replica set:
```

lenkis (approved):

PR #4412, opened 2024-08-06
Add a what's new page for the 3.2 release

### LINKED ISSUE #4391: Prepare a what's new page for the 3.2.0 release
**Product:** Tarantool
**Since:** 3.2.0
**Root document:** https://www.tarantool.io/en/doc/latest/release/
**SME:** @ KAPANDR

# Details

- [ ] Prepare a what's new page before the release: https://www.tarantool.io/en/doc/latest/release/3.2.0/
- [ ] Add a release date and a link to GitHub after the release
- [ ] Add a new release to the https://www.tarantool.io/en/doc/latest/release/ page

## Possible list of features

Config:

- [ ] https://github.com/tarantool/tarantool/issues/10117

Protobuf:

- [ ] https://github.com/tarantool/tarantool/issues/9844

Console:

- [ ] https://github.com/tarantool/tarantool/issues/9986

NP index:

- [ ] https://github.com/tarantool/tarantool/issues/9994

andreyaksenov:
Fixes https://github.com/tarantool/doc/issues/4391.

Staging: 
- https://docs.d.tarantool.io/en/doc/3.2.0-whats-new/release/3.2.0/
- https://docs.d.tarantool.io/en/doc/3.2.0-whats-new/release/#supported-versions

> [!NOTE]  
> Also added the 2.11.4/3.1.2 releases to the calendar and to the corresponding release pages: 
> - `latest`: https://docs.d.tarantool.io/en/doc/3.2.0-whats-new/release/2.11.0/, https://docs.d.tarantool.io/en/doc/3.2.0-whats-new/release/3.1.0/
> - `2.11`: https://github.com/tarantool/doc/pull/4466


doc/release/3.2.0.rst
```diff
@@ -0,0 +1,301 @@
...
+..  code-block:: lua
+
+    local schema = require('experimental.config.utils.schema')
+
+The ``validate_port()`` function can be used to check that a port value is between ``1`` and ``65535``:
+
+..  code-block:: lua
+
+    local function validate_port(port, w)
+        if (port <= 1 or port >= 65535) then
```
Totktonada: Nit: redundant parentheses.

doc/release/3.2.0.rst
```diff
@@ -0,0 +1,301 @@
...
+            w.error("'port' should be between 1 and 65535, got %q", port)
```
Totktonada: Nit: `%q` is generally for strings, `%d` fits better there.

doc/release/3.2.0.rst
```diff
@@ -0,0 +1,301 @@
...
+
+..  code-block:: lua
+
+    local listen_address_schema = schema.new('listen_address', schema.record({
+        host = schema.scalar({
+            type = 'string',
+            allowed_values = { '127.0.0.1', '0.0.0.0' },
+        }),
+        port = schema.scalar({
+            type = 'number',
```
Totktonada: Nit: `number` accepts numbers with fractional parts. I suggest to use `integer` instead.

doc/release/3.2.0.rst
```diff
@@ -0,0 +1,301 @@
...
+            validate = validate_port,
+        }),
+    }))
+
+Finally, you can pass the specified schema to the :ref:`validate() <roles_api_reference_validate>` role's function:
+
+..  code-block:: lua
+
+    local function validate(cfg)
+        if (cfg.host and cfg.port) then
```
Totktonada: Nit: redundant parentheses.

doc/release/3.2.0.rst
```diff
@@ -0,0 +1,301 @@
...
+
+-   ``host`` should be a string value from the ``allowed_values`` array.
+-   ``port`` should be a number that is checked using the ``validate_port()`` function declared above.
+
+..  code-block:: lua
+
+    local listen_address_schema = schema.new('listen_address', schema.record({
+        host = schema.scalar({
+            type = 'string',
+            allowed_values = { '127.0.0.1', '0.0.0.0' },
```
Totktonada: It looks unusual to list acceptable IP addresses. Can we add some other value to show how to use `allowed_values`?

For example,

```lua
<...> schema.new(<...>, schema.record({
    log = schema.enum({
        'info', -- only log errors
        'verbose', -- log request/response method and path
        'debug', -- includes request/response content
    }),
}))
```

BTW, there is `schema.enum()` derived constructor and the given code can be written shorter: `schema.enum({'127.0.0.1', '0.0.0.0'})`.

doc/release/3.2.0.rst
Totktonada: If `host` is set, but `port` isn't (and vice versa), is it a valid configuration?

doc/release/3.2.0.rst
```diff
@@ -0,0 +1,301 @@
...
+Once the protocol is specified, use the ``encode()`` method to encode data:
+
+..  code-block:: lua
+
+    local sample_customer = customer_protocol:encode('Customer',
+            { id = 3,
+              firstName = 'Andrew',
+              lastName = 'Fuller',
+              customerType = 1
+            })
```
Totktonada: This code formatting way looks unusual. Is it intentional?

doc/release/index.rst
```diff
@@ -61,6 +61,12 @@ For information about earlier versions, see :doc:`eol_versions`.
...
 
+        *   -   :doc:`3.2 </release/3.2.0>`
+            -   **August 21, 2024**
+            -   **August 21, 2026**
+            -   **Not planned yet**
+            -   | :tarantool-release:`3.2.0`
+
         *   -   :doc:`3.1 </release/3.1.0>`
             -   **April 16, 2024**
             -   **April 16, 2026**
```
Totktonada, line 72: We have no `3.0`, `3.1` and `3.2` as separate objects of support. In fact, `3.0` is already EOL (not sure about EOS) -- the development is now on 3.2 with backporting bugfixes to 3.1. We need to elaborate a good way to present the real commitments.

It is likely not a goal of this PR, so just noted.

Totktonada (approved):

doc/release/index.rst
andreyaksenov: Good point, thanks. Will discuss with PM.

doc/release/3.2.0.rst
andreyaksenov: Given that the schema demonstrated in this section doesn't provide default values for the host and port, both values are necessary. I can change this code snippet to smth like this ...

```Lua
local function validate(cfg)
    if cfg.host and cfg.port then
        listen_address_schema:validate(cfg)
    else
        error("Set both host and port")
    end
end
```

... , but not sure it is necessary for demo purposes.

doc/release/3.2.0.rst
andreyaksenov: Updated `allowed_values` to `host = schema.enum({ '127.0.0.1', '0.0.0.0' })` as we discussed.

doc/release/3.2.0.rst
Totktonada: If we have a small, but complete example, it is less confusing and raises less questions. I think that it would be good to handle this situation somehow (or leave a comment in the code that a real code should handle it).

xuniq (approved):

PR #4436, opened 2024-08-14
Database hot standby

### LINKED ISSUE #3858: [Config] Document the 'database.hot_standby' option
**Product:** Tarantool
**Since:** 3.0
**Root document:** https://www.tarantool.io/en/doc/latest/reference/configuration/configuration_reference/#configuration-reference-database-hot-standby
**SME:** @ totktonada

# Details

Old content: https://www.tarantool.io/en/doc/latest/reference/configuration/#hot-standby
Dev issue: https://github.com/tarantool/tarantool/issues/775

andreyaksenov:
Fixes https://github.com/tarantool/doc/issues/3858.

Staging: https://docs.d.tarantool.io/en/doc/database-hot-standby/reference/configuration/configuration_reference/#configuration-reference-database-hot-standby

### DIFF
--- a/doc/reference/configuration/configuration_reference.rst
+++ b/doc/reference/configuration/configuration_reference.rst
@@ -1428,6 +1428,40 @@ The ``database`` section defines database-specific configuration parameters, suc
 
 .. confval:: database.hot_standby
 
+    Whether to start the server in the hot standby mode.
+    This mode can be used to provide failover without :ref:`replication <replication>`.
+
+    Suppose there are two cluster applications.
+    Each cluster has one instance with the same configuration:
+
+    ..  literalinclude:: /code_snippets/snippets/config/instances.enabled/hot_standby_1/config.yaml
+        :language: yaml
+        :dedent:
+
+    In particular, both instances use the same directory for storing write-ahead logs and snapshots.
+
+    When you start both cluster applications on the same machine, the instance from the first one will be the primary instance and the second will be the standby instance.
+    In the :ref:`logs <configuration_reference_log>` of the second cluster instance, you should see a notification:
+
+    ..  code-block:: text
+
+        main/104/interactive I> Entering hot standby mode
+
+    This means that the standby instance is ready to take over if the primary instance goes down.
+    The standby instance initializes and tries to take a lock on a directory for storing write-ahead logs
+    but fails because the primary instance has made a lock on this directory.
+
+    If the primary instance goes down for any reason, the lock is released.
+    In this case, the standby instance succeeds in taking the lock and becomes the primary instance.
+
+    ``database.hot_standby`` has no effect:
+
+    * If :ref:`wal.mode <configuration_reference_wal_mode>` is set to ``none``.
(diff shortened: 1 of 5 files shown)


Totktonada (approved):
Thanks!

doc/reference/configuration/configuration_reference.rst
```diff
@@ -1428,6 +1428,40 @@ The ``database`` section defines database-specific configuration parameters, suc
 
 .. confval:: database.hot_standby
 
+    Whether to start the server in the hot standby mode.
+    This mode can be used to provide failover without :ref:`replication <replication>`.
+
+    Suppose there are two cluster applications.
+    Each cluster has one instance with the same configuration:
```
p7nov, line 1435: I can't guess from this example the correct way to use this mode in multi-instance apps. Set it for all instances of two identical apps? Or maybe select a subset of instances to run in standby?
Not sure though if the deeper explanation is needed in the reference. Perhaps somewhere in a task-oriented page in the admin's guide (out of this PR's scope).

doc/reference/configuration/configuration_reference.rst
```diff
@@ -1428,6 +1428,40 @@ The ``database`` section defines database-specific configuration parameters, suc
...
+    but fails because the primary instance has made a lock on this directory.
+
+    If the primary instance goes down for any reason, the lock is released.
+    In this case, the standby instance succeeds in taking the lock and becomes the primary instance.
+
+    ``database.hot_standby`` has no effect:
+
+    * If :ref:`wal.mode <configuration_reference_wal_mode>` is set to ``none``.
+    * If :ref:`wal.dir_rescan_delay <configuration_reference_wal_dir_rescan_delay>` is set to a large value on macOS or FreeBSD. On these platforms, the hot standby mode is designed so that the loop repeats every ``wal.dir_rescan_delay`` seconds.
+    * If spaces are created with :ref:`engine <space_opts_engine>` set to ``vinyl``.
```
p7nov: The old wording was clearer:
`for spaces created with engine = ‘vinyl’`

Now it's ambiguous:
- if _there are_ vinyl spaces?
- if _all_ spaces are vinyl?
- has no effect _only on_ vinyl spaces?

p7nov (approved):
LGTM 👍

doc/reference/configuration/configuration_reference.rst
andreyaksenov: > I can't guess from this example the correct way to use this mode in multi-instance apps.

The correct way is using `two cluster applications` with `each cluster having one instance`. And the GitHub examples at the end should help a reader. I agree that we can document this feature better. But, honestly, I don't think it is a very popular use case for Tarantool (maybe, I'm wrong).

> Set it for all instances of two identical apps? Or maybe select a subset of instances to run in standby?

This sounds like a feature request for a new configuration approach. For example, we can allow a user to add two instances, set `replication.failover` to smth like `hot_standby`, and provide the same path to WALs/snapshots for both instances. But still not sure that this scenario might be popular.  cc @Totktonada

PR #4469, opened 2024-08-26
Document the 'experimental.config.utils.schema' module

### LINKED ISSUE #4279: experimental.config.utils.schema: schema-aware hierarchical data processing
**Main related dev. issue:**

- https://github.com/tarantool/tarantool/issues/10117

**Other related dev. issues/commits:**

- https://github.com/tarantool/tarantool/issues/8725
- Commit [config/schema: store parent's annotations in nodes](https://github.com/tarantool/tarantool/commit/4ddb9f21f6b6651618551a7cf43acf364acbc30d) (from https://github.com/tarantool/tarantool/pull/8933)
- https://github.com/tarantool/tarantool/issues/10190
- https://github.com/tarantool/tarantool/issues/10193
- https://github.com/tarantool/tarantool/issues/10194
- https://github.com/tarantool/tarantool/issues/10204
- https://github.com/tarantool/tarantool/issues/10205

**Product:** Tarantool
**Since:** 3.2
**Audience/target:** module and application developers
**Root document:** 
- https://www.tarantool.io/en/doc/latest/reference/reference_lua/config/utils_schema/
- https://www.tarantool.io/en/doc/latest/platform/app/app_roles/
- https://www.tarantool.io/en/doc/latest/reference/configuration/configuration_reference/#configuration-reference-app-cfg

**SME:** @ Totktonada

# Details

## Introduction

Tarantool offers a [declarative configuration way](https://github.com/tarantool/tarantool/issues/8724) since 3.0.0. The configuration has strictly defined schema, but there are sections that accept arbitrary values. These are `app.cfg.*` and `roles_cfg.*`.

An author of an application or a role defines how these values are validated and processed.

Tarantool now offers a tool for validating and processing of data using a declarative schema: the `experimental.config.utils.schema` module.

As the name of the module says, it is in the experimental status. The API may be changed in a backward incompatible way. However, tarantool developers are conservative regarding such changes in experimental modules.

Application and role developers are encouraged to use the schema module and provide a [feedback](https://github.com/tarantool/tarantool/issues/new/choose).

## Brief API description

<details>
<summary><em>Click to expand...</em></summary>

*(This API description mostly repeats the one from https://github.com/tarantool/tarantool/issues/8725 with some updates and enhancements. It is copied here primarily to ease reading and to have all the related documentation in one place.)*

Schema node constructors:

* Define a scalar.

  ```lua
  schema.scalar({
      type = 'string', -- number, integer, boolean, any
      <..annotations..>
  })
  ```
* Define a record (an object with certain field names and field value types).

  ```lua
  schema.record({
      foo = <schema node>,
  }, {
      <..annotations..>
  })
  ```
* Define a map (an object with arbitrary key names, but strict about keys and values types).

  ```lua
  schema.map({
      key = <schema node>,
      value = <schema node>,
      <..annotations..>
  })
  ```
* Define an array.

  ```lua
  schema.array({
    items = <..schema node..>,
    <..annotations..>
  ```

Two supplementary schema node constructors are defined: `schema.enum()` and `schema.set()`.

* ```lua
  -- Accepts 'foo', 'bar' or 'baz'.
  schema.enum({
    'foo',
    'bar',
    'baz',
  }, {
      <..annotations..>
  })
  ```
* ```lua
  -- Accepts ['foo'], ['foo', 'bar'] and so on.
  schema.set({
    'foo',
    'bar',
    'baz',
  }, {
      <..annotations..>
  })
  ```

The schema object constructor wraps a schema node and adds schema name and user-provided methods.

```lua
schema.new('myschema', <schema node>[, {methods = <...>}])
```

The schema object has the following methods:

* Traversing the schema.

  ```lua
  <schema object>:pairs() -> luafun iterator
  ```

* Validate data against the schema.

  ```lua
  <schema object>:validate(data)
  ```
* Filter data based on the schema annotations.

  ```lua
  <schema object>:filter(data, f) -> luafun iterator
  ```
* Map data based on the schema annotations.

  ```lua
  <schema object>:map(data, f, f_ctx) -> new_data
  ```
* Apply default values.

  ```lua
  <schema object>:apply_default(data) -> new_data
  ```
* Get/set a nested value.

  ```lua
  <schema object>:get(data, path) -> requested field
  <schema object>:set(data, path, value)
  ```
* Merge two values.

  ```lua
  <schema object>:merge(a, b) -> new_data
  ```

The following annotations are interpreted by the module itself:

* type (string)
* validate (function)
* allowed_values (table)
* default (any)
* apply_default_if (function)

Other annotations are ignored by the module and may be used for arbitrary purposes.

The module supports parsing of data from environment variables with conversions to appropriate data types (for example, converting `MAYVAR=3301` to a Lua number), parsing of comma separated array items and key-value pairs (for example, `TT_REPLICATION_PEERS=localhost:3301,localhost:3302,localhost:3303`) and decoding JSON values.

```lua
schema.fromenv('MYVAR', os.getenv('MYVAR'), <schema>)
```
</details>

### Type system

<details>
<summary><em>Click to expand...</em></summary>

There are scalar and composite types.

A scalar type accepts a primitive value (except a special scalar type `any`). The scalar types are listed in the next section.

A composite type accepts a collection of values. There the following composite types.

* `record` describes an object with certain fields of certain types.

  It is an analogue of `struct` in C, `record` in Avro Schema and `message` in Protocol Buffers.

  A record has the following constraints:

  * The only accepted Lua type is `table`.
  * All the keys of the table are strings.
  * Only listed keys are accepted.
  * The fields have certain types (different ones in the general case).

  All the fields are optional if there are no additional constraints.

* `map` describes an object with arbitrary named keys (of the same type) with values of the same type.

  It is an analogue of `unordered_map` in C++, `map` in Avro Schema, `map` in Protocol Buffers.

  A map has the following constraints:

  * The only accepted Lua type is `table`.
  * All keys of the table have the same certain type.
  * All values of the table have the same certain type.

* `array` describes an ordered collection of elements with the same type.

  It is an analogue of an array in C, `array` in Avro Schema, a `repeated` field in Protocol Buffers.

  An array has the following constraints:

  * The only accepted Lua type is `table`.
  * All the keys of the table are numeric, without a fractional part.
  * The lower key is 1.
  * The higher key is equal to the number of items.
  * All items of the table have the same certain type.
</details>

### Scalars

<details>
<summary><em>Click to expand...</em></summary>

| Scalar type | Lua type        | Constraints              |
| ----------- | --------------- | ------------------------ |
| `string`    | `string`        |                          |
| `number`    | `number`        |                          |
| `integer`   | `number`        | `x - math.floor(x) == 0` |
| `boolean`   | `boolean`       |                          |
| `string, number` | `string` or `number` |                |
| `number, string` | `string` or `number` |                |
| `any`       | <..arbitrary..> |                          |

`any` accepts an arbitrary Lua type, including `table`. A scalar of the `any` type may be used to declare an arbitrary value that doesn't need any validation.
</details>

### Schema node constructors: scalar, record, map, array

<details>
<summary><em>Click to expand...</em></summary>

The module provides the following functions to create schema node objects.

```lua
-- Create a scalar.
--
-- Returns a table of the following shape.
--
-- {
--     type = <...>,
--     <..annnotations..>
-- }
schema.scalar({
    type = one-of(
        'string',
        'number',
        'integer',
        'boolean',
        'any',
        'string, number',
        'number, string',
    ),
    <..annotations..>
})

-- Example.
schema.scalar({
    type = 'string',
    description = 'lorem ipsum',
})
-- =>
{
    type = 'string',
    description = 'lorem ipsum',
}
```

```lua
-- Create a record.
--
-- Returns a table of the following shape.
--
-- {
--     type = 'record',
--     fields = <...>,
--     <..annotations..>
-- }
schema.record({
    [<field name>] = <schema node>,
    <...>,
}, {
    <..annotations..>
})

-- Example.
schema.record({
    foo = schema.scalar({type = 'string'}),
    bar = schema.scalar({type = 'integer'}),
}, {
    description = 'lorem ipsum',
})
-- =>
{
    type = 'record',
    fields = {
        foo = {type = 'string'},
        bar = {type = 'integer'},
    },
    description = 'lorem ipsum',
}
```

```lua
-- Create a map.
--
-- Returns a table of the following shape.
--
-- {
--     type = 'map',
--     key = <...>,
--     value = <...>,
--     <..annotations..>
-- }
schema.map({
    key = <schema node>,
    value = <schema node>,
    <..annotations..>
})

-- Example.
schema.map({
    key = schema.scalar({type = 'string'}),
    value = schema.scalar({type = 'string'}),
    description = 'lorem ipsum',
})
-- =>
{
    type = 'map',
    key = {type = 'string'},
    value = {type = 'string'},
    description = 'lorem ipsum',
}
```

```lua
-- Create an array.
--
-- Returns a table of the following shape.
--
-- {
--     type = 'array',
--     items = <...>,
--     <..annotations..>
-- }
schema.array({
    items = <schema node>,
    <..annotations..>
})

-- Example.
schema.array({
    items = schema.scalar({type = 'string'}),
    description = 'lorem ipsum',
})
-- =>
{
    type = 'array',
    items = {type = 'string'},
    description = 'lorem ipsum',
}
```

A general structure of a schema node is the following.

```lua
{
    -- One of scalar types, 'record', 'map' or 'array'.
    type = <string>,
    -- For a record.
    fields = <table>,
    -- For a map.
    key = <table>,
    value = <table>,
    -- For an array.
    items = <table>,
    -- Arbitrary user specified annotations.
    <..annotations..>
}
```
</details>

### Derived schema node type constructors: enum, set

<details>
<summary><em>Click to expand...</em></summary>

`schema.enum` accepts a string from the given set of allowed string values. It uses an `allowed_values` annotation, see `<schema object>:validate()` for details about this annotation.

```lua
-- Shortcut for a string scalar with the given allowed values.
--
-- Returns a table of the following shape.
--
-- {
--     type = 'string',
--     -- As given in the first argument.
--     allowed_values = <...>,
--     -- As given in the second argument.
--     <..annotations..>
-- }
schema.enum(allowed_values, annotations)

-- Example.
schema.enum({'foo', 'bar'}, {description = 'lorem ipsum'})
-- =>
{
    type = 'string',
    allowed_values = {'foo', 'bar'},
    description = 'lorem ipsum',
}
```

`schema.set` accepts an array of unique string values, where each of the strings is from the given set of allowed string values.

It uses `allowed_values` and `validate` annotations, see `<schema object>:validate()` for details about these annotations.

```lua
-- Shortcut for array of unique string values from the given list
-- of allowed values.
--
-- Returns a table of the following shape.
--
-- {
--     type = 'array',
--     items = {
--         type = 'string',
--         -- As given in the first argument.
--         allowed_values = <...>,
--     },
--     -- This validate function checks that the incoming
--     -- array has no repeated values.
--     validate = <function>,
--     -- As given in the second argument.
--     <..annotations..>
-- }
schema.set(allowed_values, annotations)

-- Example.
schema.set({'foo', 'bar'}, {description = 'lorem ipsum'})
-- =>
{
    type = 'array',
    items = {
        type = 'string',
        allowed_values = {'foo', 'bar'},
    },
    validate = <function>,
    description = 'lorem ipsum',
}
```
</details>

### Schema object constructor: new

<details>
<summary><em>Click to expand...</em></summary>

A schema node can be transformed to a schema object. Unlike a schema node, the schema object has a name, has methods described below and may have user-provided methods.

```lua
-- Create a schema object.
--
-- name: string, schema object name
-- schema: table, schema node
-- opts:
--   methods: table, user-provided methods
--
-- Returns a table of the following shape.
--
-- {
--     name = <string>,
--     schema = <schema node>,
--     methods = <table>,
-- }
--
-- It has a metatable with methods described in this document.
schema.new(name, schema, opts)

-- Example.
schema.new('foo', schema.scalar({type = 'string'}))
-- =>
-- {
--     name = 'foo',
--     schema = {
--         type = 'string',
--         computed = {
--             annotations = {},
--         },
--     }
-- }
```

The given schema node is recursively copied and computed fields are added to each schema node into the `computed` field. See the 'Computed annotations' sections for details.

Indexing a schema object is performed as follows:

* Look up a user-provided method.
* Look up a module provided method.
* Look up a field in the schema object table.
* Return nil.

An example of a user-provided method:

```lua
local point_schema = schema.new('point', schema.record({
    x = schema.scalar({type = 'number'}),
    y = schema.scalar({type = 'number'}),
}), {
    methods = {
        distance = function(_self, a, b)
            return math.sqrt((a.x - b.x)^2 + (a.y - b.y)^2)
        end,
    },
})

point_schema:distance({x = 0, y = 0}, {x = 3, y = 4})
-- => 5
```
</details>

### `<schema object>:validate()`

<details>
<summary><em>Click to expand...</em></summary>

Validate the given data against the given schema.

```lua
-- Raise an error if the given data doesn't adhere the given
-- schema.
<schema object>:validate(data)
```

The method performs a recursive type checking. See the 'Type system' and 'Scalars' sections for details how exactly it is performed for each of the given types.

Aside of the type checking the method performs validation based on a user-provided annotations. It is described below in this section.

Nuances:

* `schema.new('<...>', schema.scalar(<...>))` doesn't accept `nil` and `box.NULL`. However,
* All fields in a record are optional: they accept `nil` and `box.NULL`.
* There is no record/map/array autoguessing: the given data is validated against the given schema node by rules defined by this schema node. Also, `mt.__serialize` marks in the data are not involved anyhow.
* An array shouldn't have any holes (`nil` values in a middle).

Annotations taken into accounts:

* `allowed_values` (table) -- whitelist of values
* `validate` (function) -- schema node specific validator

The `validate` annotation is a user-provided function that accepts the following arguments.

```lua
{
    <...>
    validate = function(data, w)
        -- w.schema -- current schema node
        -- w.path -- path to the node
        -- w.error -- function that prepends a caller provided
        --            error message with context information;
        --            use it for nice error messages
    end,
}
```

The user-provided `validate` function is called after all the type validation is done, including nested nodes, and, also, after the `allowed_values` check.

The contract is that the `validate` annotation raises an error to fail the check. There is a convenient `w.error()` helper that raises an error formatted with schema name and path to the current schema node.

Example:

```lua
#!/usr/bin/env tarantool

local schema = require('experimental.config.utils.schema')

local function validate_email(email, w)
    if email:find('@') == nil then
        w.error('A email must contain @ symbol, got %q', email)
    end
end

local personal_info_schema = schema.new('personal_info', schema.record({
    email = schema.scalar({
        type = 'string',
        validate = validate_email,
    }),
}))

personal_info_schema:validate({email = 'foo'})
-- error: [personal_info] email: A email must contain @ symbol, got "foo"
```

The user-provided `validate` function may use computed annotations. See an example in the 'Computed annotations' section below.

Beware: The user-provided `validate` function is not called on a record's field that has `nil`/`box.NULL` value. If an error should be raised on a missing field, add the `validate` annotation on the outer level (on a record, not the field itself).
</details>

### `<schema object>:get()`

<details>
<summary><em>Click to expand...</em></summary>

Get nested data that is pointed by the given path.

```lua
<schema object>:get(data, path)
```

Important: the data is assumed as already validated against the given schema.

The indexing is performed in the optional chaining manner (`'foo.bar'` works like `foo?.bar` in TypeScript).

The method checks the path against the schema: it doesn't allow to use a non-existing field or index a scalar value.

The path is either array-like table or a string in the dot notation.

```lua
local myschema = schema.new(<...>)

local data = {foo = {bar = 'x'}}
myschema:get(data, 'foo.bar') -- => 'x'
myschema:get(data, {'foo', 'bar'}) -- => 'x'

local data = {}
myschema:get(data, 'foo.bar') -- => nil
```

#### Nuances

* `nil`, `''`, `{}` in the `path` argument means the root node: IOW, returns the `data` argument as is.
* Array indexing is not supported yet.
* A scalar of the `any` type can be indexed if it is a `table` or `nil`/`box.NULL`. In this case a tail of the path that is inside the `any` type is not checked against a schema. Indexing `nil`/`box.NULL` always returns `nil`.
</details>

### `<schema object>:set()`

<details>
<summary><em>Click to expand...</em></summary>

Set the given `rhs` value at the given path in the `data`.

```lua
<schema object>:set(data, path, rhs)
```

Important: `data` is assumed as already validated against the given schema, but `rhs` is validated by the method before the assignment.

The method checks the path against the schema: it doesn't allow to use a non-existing field or index a scalar value.

The path is either array-like table or a string in the dot notation.

```lua
local myschema = schema.new(<...>)

local data = {}
myschema:set(data, 'foo.bar', 42)
print(data.foo.bar) -- 42

local data = {}
myschema:set(data, {'foo', 'bar'}, 42)
print(data.foo.bar) -- 42
```

#### Nuances

* A root node (pointed by the empty path) can't be set using this method.
* Array indexing is not supported yet.
* ~A scalar of the 'any' type can't be indexed, even when it is a table. It is OK to set the whole value of the 'any' type.~ (This restriction will be relaxed in 3.2.0 in the scope of https://github.com/tarantool/tarantool/issues/10204.)
* Assignment of a non-nil `rhs` value creates intermediate tables over the given `path` instead of `nil` or `box.NULL` values.

#### Field deletion

If `rhs` is `nil`, it means deletion of the pointed field.

How it works (in examples):

Intermediate tables are not created:

```lua
local myschema = <...>
local data = {}
myschema:set(data, 'foo.bar', nil)
-- data is {}, not {foo = {}}
```

Existing tables on the `path` are not removed:

```lua
local myschema = <...>
local data = {
    instances = {
        foo = {x = 1},
        bar = {},
    },
}
myschema:set(data, 'instances.foo.x', nil)
-- data is {
--     instances = {
--         foo = {},
--         bar = {},
--     },
-- }
```
</details>

### `<schema object>:filter()`

<details>
<summary><em>Click to expand...</em></summary>

Filter data based on the schema annotations.

```lua
<schema object>:filter(data, f) -> luafun iterator
```

Important: the data is assumed as already validated against the given schema. (A fast type check is performed on composite types, but it is not recommended to lean on it.)

The user-provided filter function `f` receives the following table as the argument:

```lua
w = {
    path = <array-like table>,
    schema = <schema node>,
    data = <data at the given path>,
}
```

The filter function returns a boolean value that is interpreted as 'accepted' or 'not accepted'.

The user-provided function `f` is called for each schema node, including ones that have box.NULL value (but not nil). A node of a composite type (record/map/array) is not traversed down if it has nil or box.NULL value.

The `:filter()` function returns a luafun iterator by all `w` values accepted by the `f` function.

A composite node that is not accepted still traversed down.

Examples:

```lua
-- Do something for each piece of data that is marked by the
-- given annotation.
s:filter(function(w)
    return w.schema.my_annotation ~= nil
end):each(function(w)
    do_something(w.data)
end)
```

```lua
-- Group data by a value of an annotation.
local group_by_my_annotation = s:filter(function(w)
    return w.schema.my_annotation ~= nil
end):map(function(w)
    return w.schema.my_annotation, w.data
end):tomap()
```

#### Nuances

* box.NULL is assumed as an existing value, so the user-provided filter function `f` is called for it. However, it is not called for `nil` values. See details below.
* While it is technically possible to pass information about a field name for record/map field values and about an item index for an array item value, it is not implemented for simplicity.
* `w.path` for a map key and a map value are the same. It seems, we should introduce some syntax to point a key in a map, but it is not implemented yet.

#### nil/box.NULL nuances explanation

Let's assume that a record defines three scalar fields: 'foo', 'bar' and 'baz'. Let's name a schema object that wraps the record as `s`.

* `s:filter(nil, f)` calls `f` only for the record itself.
* `s:filter(box.NULL, f)` works in the same way.
* `s:filter({foo = box.NULL, bar = nil}, f)` calls `f` two times: for the record and for the 'foo' field.

This behavior is needed to provide ability to handle box.NULL values in the data somehow. It reflects the pairs() behavior on a usual table, so it looks quite natural.
</details>

### `<schema object>:map()`

<details>
<summary><em>Click to expand...</em></summary>

Transform data by the given function.

```lua
<schema object>:map(data, f, f_ctx) -> new_data
```

Leave the shape of the data unchanged.

Important: the data is assumed as already validated against the given schema. (A fast type check is performed on composite types, but it is not recommended to lean on it.)

The user-provided transformation function receives the following three arguments in the given order:

* `data` -- value at the given path
* `w` -- walkthrough node, described below
* `ctx` -- user-provided context for the transformation function

The walkthrough node `w` has the following fields:

* `w.schema` -- schema node at the given path
* `w.path` -- path to the schema node
* `w.error` -- function that prepends a caller provided error message with context information; use it for nice error messages

An example of the mapping function:

```lua
local function f(data, w, ctx)
    if w.schema.type == 'string' and data ~= nil then
        return data:gsub('{{ *foo *}}', ctx.foo)
    end
    return data
end
```

The :map() method is recursive with certain rules:

* All record fields are traversed unconditionally, including ones with nil/box.NULL values. Even if the record itself is nil/box.NULL, its fields are traversed down (assuming their values as nil).

  It is important when the original data should be extended using some information from the schema: say, default values.
* It is not the case for a map and an array: nil/box.NULL fields and items are preserved as is, they're not traversed down. If the map/the array itself is nil/box.NULL, it is preserved as well.

  A map has no list of fields in the schema, so it is not possible to traverse it down. Similarly, an array has no items count in the schema.

The method attempts to preserve the original shape of values of a composite type:

* nil/box.NULL record is traversed down, but if all the new field values are nil, the return value is the original one (nil/box.NULL), not an empty table.
* nil/box.NULL values for a map and an array are preserved.

#### Nuances

* The user-provided transformation function is called only for scalars.
* nil/box.NULL handling for composite types. Described above.
* `w.path` for a map key and a map value are the same. It seems, we should introduce some syntax to point a key in a map, but it is not implemented yet.
</details>

### `<schema object>:apply_default()`

<details>
<summary><em>Click to expand...</em></summary>

Apply default values from the schema.

```lua
<schema object>:apply_default(data) -> new_data
```

Important: the data is assumed as already validated against the given schema. (A fast type check is performed on composite types, but it is not recommended to lean on it.)

Annotations taken into accounts:

* `default` -- the value to be placed instead of a missed one
* `apply_default_if` (function) -- whether to apply the default

  ```lua
  {
      default = <...>,
      apply_default_if = function(data, w)
         -- w.schema -- current schema node
         -- w.path -- path to the node
         -- w.error -- for nice error messages
      end,
  }
  ```

  If there is no `apply_default_if` annotation, the default is assumed as to be applied.

#### Nuances:

* Defaults are taken into account only for scalars.
* The method works for static defaults, but it doesn't work for dynamic default values or for defaults that depend on the data somehow. Use `:map()` for such scenarios.
</details>

### `<schema object>:merge()`

<details>
<summary><em>Click to expand...</em></summary>

Merge two hierarical values (prefer the latter).

```lua
<schema object>:merge(a, b) -> new_data
```

Important: the data is assumed as already validated against the given schema. (A fast type check is performed on composite types, but it is not recommended to lean on it.)

`box.NULL` is preferred over `nil`, any `X` where `X ~= nil` is preferred over `nil`/`box.NULL`.

Records and maps are deeply merged. Scalars and arrays are all-or-nothing: the right hand one is chosen if both are not `nil`/`box.NULL`.

The formal rules are below.

Let's define the merge result for `nil` and `box.NULL` values:

1. `merge(nil, nil)` -> `nil`
2. `merge(nil, box.NULL)` -> `box.NULL`
3. `merge(box.NULL, nil)` -> `box.NULL`
4. `merge(box.NULL, box.NULL)` -> `box.NULL`

Let's define `X` as a value that is not `nil` and is not `box.NULL`.

5. `merge(X, nil)` -> `X`
6. `merge(X, box.NULL)` -> `X`
7. `merge(nil, X)` -> `X`
8. `merge(box.NULL, X)` -> `X`

If the above conditions are not meet, the following type specific rules are in effect.

9. `merge(<scalar A>, <scalar B>)` -> `<scalar B>`
10. `merge(<array A>, <array B>)` -> `<array B>`
11. `merge(<record A>, <record B>)` -> `deep-merge(A, B)`
12. `merge(<map A>, <map B>)` -> `deep-merge(A, B)`

For each key `K` in `A` and each key `K` in `B`: `deep-merge(A, B)[K]` is `merge(A[K], B[K])`.

#### Nuances

* A scalar of the `any` type is NOT deeply merged even if it is a table, however it may be useful. We'll consider adding support of such a behaviour in some backward-compatible way in a future.
* Arrays are not concatenated (the right hand one wins), however it may be useful too. The original idea is that we don't know, whether ordinals in the array are important, but the practice shows that arrays in configuration data are always just sets with an order -- ordinals do not matter.

  Also, the given behavior (the right hand one wins) allows to discard items from the left hand side array that may be useful too. A concatenation behavior wouldn't allow it.

  Anyway, the concatenation behavior may be considered to implement in some backward-compatible way in a future.
</details>

### `<schema object>:pairs()`

<details>
<summary><em>Click to expand...</em></summary>

Walk over the schema and return scalar, array and map schema nodes (all nodes except records).

```lua
<schema object>:pairs() -> luafun iterator
```

Usage example:

```lua
for _, w in schema:pairs() do
    local path = w.path
    local schema = w.schema
    <...>
end
```
</details>

### Parse an environment variable

<details>
<summary><em>Click to expand...</em></summary>

Parse data from an environment variable as a value of the given type.

```lua
schema.fromenv(env_var_name, raw_value, schema) -> data
```

Important: the result is not necessarily valid against the given schema node. It should be validated using the `<schema object>:validate()` method before further processing.

`env_var_name` is used for error messages.

`raw_value` is to be received using `os.getenv()` or `os.environ()`.

`schema` is a schema node, not a schema object.

How the raw value is parsed depends on the schema node type.

#### Scalars

* `string` -- return the raw value as is
* `number` -- attempt to parse a number using `tonumber()`, fail if unsuccessful
* `string, number` (and its alias `number, string`) -- attempt to parse a number using `tonumber()`; if it is unsuccessful return the raw value as is
* `integer` -- attempt to parse an integral number using `tonumber64()`, fail if unsuccessful
* `boolean` -- accept `0`/`1` and `true`/`false` (case insensitively), fail on other values
* `any` -- parse as JSON, fail if the decoding fails

#### Record

Not supported.

It is technically possible to implement parsing of records similarly how it is done for maps, but it is not implemented yet.

#### Map

Accepts two formats:

1. JSON format (if data starts from `{`).
2. Simple `foo=bar,baz=fiz` object format (otherwise).

The simple format is applicable for a map with string keys and scalar values of all types except `any`.

In the simple format the field values are parsed according to its type in the schema (see the rules for scalars above).
#### Array

Accepts two formats:

1. JSON format (if data starts from `[`).
2. Simple `foo,bar,baz` array format (otherwise).

The simple format is applicable for an array with scalar item values of all types except `any`.

In the simple format the item values are parsed according to its type in the schema (see the rules for scalars above).
</details>

### Computed annotations

<details>
<summary><em>Click to expand...</em></summary>

The idea is to have information from the ancestor nodes accessible from the given schema node.

Example:

```lua
local schema = require('experimental.config.utils.schema')

local abilities = schema.record({
    walking = schema.scalar({type = 'boolean'}),
    swimming = schema.scalar({type = 'boolean'}),
    flying = schema.scalar({type = 'boolean'}),
}, {
    validate = function(data, w)
        local kind = w.schema.computed.annotations.kind
        if kind == 'penguin' and data.flying then
            w.error('A penguin is unable to fly')
        end
    end,
})

local duck_schema = schema.new('duck', schema.record({
    name = schema.scalar({type = 'string'}),
    abilities = abilities,
}, {
    kind = 'duck',
}))

local penguin_schema = schema.new('penguin', schema.record({
    name = schema.scalar({type = 'string'}),
    abilities = abilities,
}, {
    kind = 'penguin',
}))

local bird_data = {
    name = 'Gurr',
    abilities = {
        walking = true,
        swimming = true,
        flying = true,
    },
}

duck_schema:validate(bird_data)
penguin_schema:validate(bird_data)
-- error: [penguin] abilities: A penguin is unable to fly
```

The example demonstrates how the `kind` annotation from the outermost schema node is used in a `validate` function of a nested schema node.

`schema.new` call prepares each schema node in such a way that the `computed.annotations` field contains all the annotations merged from the root schema node down to the given one. If the same annotation is present in an ancestor node and in an descendant node, the latter is preferred.

There are two classes of schema node table fields that are not considered as annotations to merge into the `computed.annotations` field:

* keys that are part of the schema node tree structure: `type`, `fields`, `key`, `value`, `items`
* known annotations that barely has any sense in context of descendant schema nodes: `allowed_values`, `validate`, `default`, `apply_default_if`
</details>

# Definition of done

This is relatively large topic and it seems logical to split the 'document everything' goal to some subtasks.

- [ ] Schema node constructors are documented
- [ ] Schema constructor is documented
- [ ] Schema object methods are documented
- [ ] `schema.fromenv` is documented
- [ ] A list of annotation that influence module's functions/methods behavior is documented

# Planning checklist

- [x] Pick product label: server.
- [x] Pick type label: reference.
- [ ] Estimate complexity in storypoints in the title
- [ ] Add to **Documentation** board → **Backlog**

andreyaksenov:
Fixes https://github.com/tarantool/doc/issues/4279.

# New samples

The following new samples were added to demonstrate the main use cases of this module.

### Config hierarchy

- `config_schema_nodes_scalar`
- `config_schema_nodes_record`
- `config_schema_nodes_record_hierarchy` (based on `config_schema_node_record`)
- `config_schema_nodes_array` (based on `config_schema_node_record`)
- `config_schema_nodes_map` (based on `config_schema_node_record`)

### Annotations

- `config_schema_annotations` (based on `config_schema_node_record`)

### Helper methods

- `config_schema_methods` (based on `config_schema_node_record`)

### Environment variables

- `config_schema_fromenv` (based on `config_schema_node_record`)

### Computed annotations

- `config_schema_computed_annotations` (based on `config_schema_nodes_record_hierarchy`)

# Updated samples

The `application_role_http_api` sample has been updated to use the new module.

# Docs

- New Lua reference page [Submodule experimental.config.utils.schema](https://docs.d.tarantool.io/en/doc/config_utils_schema/reference/reference_lua/config/utils_schema)
- [Platform > Applications > Application roles](https://docs.d.tarantool.io/en/doc/config_utils_schema/platform/app/app_roles/#validating-a-role-configuration): mention the new module and update sample.
- [Configuration reference > app.cfg](https://docs.d.tarantool.io/en/doc/config_utils_schema/reference/configuration/configuration_reference/#configuration-reference-app-cfg) and [roles_cfg](https://docs.d.tarantool.io/en/doc/config_utils_schema/reference/configuration/configuration_reference/#configuration-reference-roles-cfg): mention the new module


doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1163 @@
...
+        :dedent:
+
+4.  Obtain values of configuration options using :ref:`get() <config-schema_object-get>`.
+    In case of a role, call it inside the role's :ref:`apply() <roles_create_custom_role_apply>` function:
+
+    ..  literalinclude:: /code_snippets/snippets/config/instances.enabled/config_schema_nodes_scalar/http_api.lua
+        :language: lua
+        :start-at: local function apply
+        :end-before: local function stop
+        :dedent:
```
Totktonada: The code here is the following:

```lua
local function apply(cfg)
    local http_api_cfg = http_api_schema:get(cfg)
    log.info("HTTP API endpoint: %s", http_api_cfg)
end
```

But there is no sense to call `<schema object>:get(data, path)` if the `path` is `nil`. It returns the same `cfg` value in this case.

There is some sense to use `:get()` when we have two nesting levels in the configuration, for example:

```lua
local bar = myschema:get(cfg, 'foo.bar')
```

Unlike `cfg.foo.bar` here we have the optional chaining semantic: if there is no `foo`, we get `nil`, instead of `error: attempt to index nil value`.

Even using `:get()` with just one level (with a constant path) has no sense:

```lua
local foo = myschema:get(cfg, 'foo')
-- roughtly same as
local foo = cfg.foo
```

The former checks that the `foo` field exists in the schema, but the value is anyway constant.

doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1163 @@
...
+or a :ref:`roles_cfg <configuration_reference_roles_cfg>` section. It includes
+option names, types, hierarchy, and other aspects of a configuration.
+
+To create a schema, use the :ref:`schema.new() <config-utils-schema-new>` function.
+It has the following arguments:
+
+-   schema name -- an arbitrary string to use as an identifier
+-   root schema node -- a table describing the hierarchical schema structure
+    starting from the root.
+-   (optional) methods -- functions that can be called on this schema object.
```
Totktonada: I would say 'user-defined methods' here to highlight that it is not some kind of filter of the built-in methods.

BTW, I would say 'method' instead of 'function' for functions that take the schema object as the first argument and can be called using a colon: `<schema object>:foo(<...>)`.

doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1163 @@
...
+    :dedent:
+
+.. _config_utils_schema_annotation:
+
+Annotations
+***********
+
+Node *annotations* are named attributes that define the its various aspects. For example,
+scalar nodes have a required annotation ``type`` that defines the node value type.
+Other annotations can optionally set default values, validation function, and
```
Totktonada: And?.. :)

doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1163 @@
...
+
+.. _config_utils_schema_built_in_annotations:
+
+Built-in annotations
+^^^^^^^^^^^^^^^^^^^^
+
+Built-in annotations are interpreted by the module itself. There are the following
+built-in annotations:
+
+-   :ref:`type <config-schema_node_annotation-type>` -- the node value type. Mandatory for scalar nodes, except for those created with ``schema.enum()``.
```
Totktonada: > Mandatory for scalar nodes

The `type` field is mandatory for all the schema nodes (including ones of a composite type). However, there are schema node constructors that set it on its own: `schema.record()`, `schema.map()`, `schema.array()`, `schema.enum()`, `schema.set()`.

Let's describe this nuance in regards to the schema node *constructors*, not schema nodes.

doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1163 @@
...
+
+    **Example:** see :ref:`config_utils_schema_transform_configuration`
+
+    **See also:** :ref:`schema_object:apply_default() <config-schema_object-apply_default>`
+
+..  _config-schema_node_annotation-type:
+
+-   ``type``
+
+    A value type.
```
Totktonada: Maybe we should be a bit more precise here and say 'a schema node type' instead. A value with type `number` may be accepted by a node with type `number`, `integer`, `any` (and `string, number`, `number, string` as well). These are different things.

doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1163 @@
...
+
+    **See also:** :ref:`config_utils_schema_data_types`
+
+..  _config-schema_node_annotation-validate:
+
+-   ``validate``
+
+    A boolean function used to validate node data. Node data is valid if this function
+    returns ``true``. The function is called upon the :ref:`schema_object:validate() <config-schema_object-validate>`
+    function calls.
```
Totktonada: > Node data is valid if this function returns `true`.

From where this information comes? This is incorrect.

doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1163 @@
...
+    :param table opts: a table in the following format:
+
+                       ..  code-block:: text
+
+                           { items = <schema_node_object>, <..annotations..> }
+
+                       See also: :ref:`schema_node_object <config-utils-schema_node_object>`, :ref:`schema_node_annotation <config-utils-schema_node_annotation>`.
+
+    :return: the created schema node
+    :rtype: table
```
Totktonada: The return value is not somewhat implementation-defined and incapsulates inside this implementation, it is a part of the public API of the module. We let a user to operate on schema nodes in `validate` annotation, `:map()` method and other parts of the API.

Let's describe what exactly the constructor returns.

doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1163 @@
...
+
+..  function:: schema.array(opts)
+
+    Create an array node of a configuration schema.
+
+    :param table opts: a table in the following format:
+
+                       ..  code-block:: text
+
+                           { items = <schema_node_object>, <..annotations..> }
```
Totktonada: Please, don't write 'concise' code by merging different things into a single line.

We have items and other annotations here. Let's format this code as usual for tables with several fields:

```lua
{
    items = <schema node>,
    <..annotations..>
}
```

doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1163 @@
...
+..  _config-utils-schema_functions:
+
+Functions
+~~~~~~~~~
+
+..  module:: config.utils.schema
+
+..  _config-utils-schema-array:
+
+..  function:: schema.array(opts)
```
Totktonada: `opts` is usually used for some optional parameters. I think it is barely applicable here.

I use `array_def` for this argument in the source code of the module.

Similar remark is applicable for other schema node constructors.

doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1163 @@
...
+            -   Create a schema
+
+        *   -   :ref:`schema.record() <config-utils-schema-record>`
+            -   Create a record
+
+        *   -   :ref:`schema.scalar() <config-utils-schema-scalar>`
+            -   Create a scalar
+
+        *   -   :ref:`schema.set() <config-utils-schema-set>`
+            -   Create a set
```
Totktonada: TBH, I don't like that all the functions are in one bag, while we have separate groups in the head. At least, I would separate schema node constructors into its own group.

doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1163 @@
...
+
+    How the raw value is parsed depends on the ``schema_node`` type:
+
+    -   Scalar:
+
+        -   ``string``: return the value as is
+        -   ``number`` or ``integer``: parse the value as a number or an integer
+        -   ``string, number``: attempt to parse as a number; in case of a failure
+            return the value as is
+        -   ``boolean``: accept ``true`` or ``1`` for ``true``, ``false`` or ``0`` for ``false``
```
Totktonada: These `true` and `false` source values are taken case-insensitively. I think it is a part of the API and should be mentioned.

doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1163 @@
...
+        -   ``any``: parse the value as a JSON
+
+    -   Map: parse either as JSON (if the raw value starts with ``{``)
+        or as a comma-separated string of ``key=value`` pairs: ``key1=value1,key2=value2``
+    -   Array: parse either as JSON (if the raw value starts with ``[``)
+        or as a comma-separated string of items: ``item1,item2,item3``
```
Totktonada, line 706: We say nothing about a record. It is not supported yet, right. I would say it explicitly, because now it is unclear.

doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1163 @@
...
+
+    :param string env_var_name: environment variable name
```
Totktonada: We can explicitly say that it is used for an error reporting. Otherwise, one may assume that the function calls `os.getenv()` on its own.

doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1163 @@
...
+    :param string raw_value: environment variable value
+    :param schema_node_object schema_node: a schema node (see :ref:`schema_node_object <config-utils-schema_node_object>`)
+
+    :return: the created schema node
+    :rtype: table
```
Totktonada: `schema.fromenv()` returns the parsed value.

doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1163 @@
...
+..  function:: schema.new(schema_name, schema_node[, { methods = <...> }])
+
+    Create a schema object.
+
+    :param string schema_name: a name
+    :param table schema_node: a root schema node
+    :param table methods: methods
+
+    :return: a new schema instance (see :ref:`schema_object <config-utils-schema_object>`)
+    :rtype: userdata
```
Totktonada: From where this information comes? It is incorrect.

doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1163 @@
...
+        :return: configuration data with applied schema defaults
+
+        **See also:** :ref:`default <config-schema_node_annotation-default>`, :ref:`apply_default_if <config-schema_node_annotation-apply_default_if>`
+
+    ..  _config-schema_object-filter:
+
+    ..  method:: filter(data, f)
+
+        Filter data based on the schema annotations. The methods returns an iterator
+        by configuration nodes for which the given filter function ``f`` returns ``true``.
```
Totktonada: Typo: the methods -> the method.

doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1163 @@
...
+..  _config-utils-schema_object:
+
+schema_object
+~~~~~~~~~~~~~
+
+..  class:: schema_object
+
+    ..  _config-schema_object-apply_default:
+
+    ..  method:: apply_default(data)
```
Totktonada, line 859: From the documentation request:

> Important: the data is assumed as already validated against the given schema.

And it is actually important -- this is a part of the method's contract. Please, include it into the API reference.

The same remark is applicable for some other methods -- see the documentation request.

doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1163 @@
...
+        **Example:**
+
+        ..  literalinclude:: /code_snippets/snippets/config/instances.enabled/config_schema_nodes_record_hierarchy/http_api.lua
+            :language: lua
+            :start-at: local scheme
+            :end-before: local host
+            :dedent:
+
+        :param any data: configuration data
+        :param any path: path to the target node
```
Totktonada: From where the `path`'s type information comes? It is either a string in the dot-notation or an array-like table. Not something arbitrary.

doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1163 @@
...
+            end
+
+        The ``map()`` method traverses all fields of the schema records,
+        even if they are ``nil`` or ``box.NULL`` in the provided configuration.
+        This allows using this method to set computed default values for missing
+        fields. Note that this is not the case for maps and arrays since the schema
+        doesn't define their fields to traverse.
+
+        :param any data: value at the given path
+        :param any f: walkthrough node, described below
```
Totktonada: `f` is not a walkthrough node, but the transformation function. And it a `function` or at least a callable.

doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1163 @@
...
+                return data
+            end
+
+        The ``map()`` method traverses all fields of the schema records,
+        even if they are ``nil`` or ``box.NULL`` in the provided configuration.
+        This allows using this method to set computed default values for missing
+        fields. Note that this is not the case for maps and arrays since the schema
+        doesn't define their fields to traverse.
+
+        :param any data: value at the given path
```
Totktonada: Which `path`?

doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1163 @@
...
+        :param any f: walkthrough node, described below
+        :param any f_ctx: user-provided context for the transformation function
+
+        :return: transformed configuration data
+
+    ..  _config-schema_object-merge:
+
+    ..  method:: merge(data_a, data_b)
```
Totktonada, line 1000: I would like to have one important remark here. Quoted from the documentation request:

> A scalar of the `any` type is NOT deeply merged even if it is a table, however it may be useful. We'll consider adding support of such a behaviour in some backward-compatible way in a future.

I want to change it in https://github.com/tarantool/tarantool/issues/10450 (no precise plans here, though), but until that it remains to be the important detail.

doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1163 @@
...
+    ..  _config-schema_object-set:
+
+    ..  method:: set(data, path, value)
+
+        Set a given value at the given path in a configuration.
+        The path can be either a dot-separated string (``http.scheme``) or
+        an array-like table (``{ 'http', 'scheme'}``).
+
+        :param any data: configuration data
+        :param any path: path to the target node
```
Totktonada: I wouldn't say that `path` has any type. It is either a string in the dot-notation or an array-like table of strings.

Totktonada (approved):
No strict objections from my side.

There are points that require some polishing, but everything is generally good.

Thanks for working on it!

doc/reference/reference_lua/config/utils_schema.rst
p7nov: Thanks. 
- Added `user-defined` to this line (methods definition)
- Replaced `function` with `method` when referring to schema object's methods across this page.

doc/reference/reference_lua/config/utils_schema.rst
p7nov: My bad, fixed.

doc/reference/reference_lua/config/utils_schema.rst
p7nov: Added this info.

doc/reference/reference_lua/config/utils_schema.rst
p7nov: This list is pretty short, and items have obvious descriptions (`Create a .. [node]`), so I'd keep the structure plain.
Also, I see only one function that is not a constructor - `fromenv`. It would look rather weird to create a dedicated group for it. Perhaps, we can move it to the bottom of the list.

doc/reference/reference_lua/config/utils_schema.rst
p7nov: Fixed

doc/reference/reference_lua/config/utils_schema.rst
p7nov: Fixed

doc/reference/reference_lua/config/utils_schema.rst
p7nov: Added

doc/reference/reference_lua/config/utils_schema.rst
p7nov: Added

doc/reference/reference_lua/config/utils_schema.rst
p7nov: Added

doc/reference/reference_lua/config/utils_schema.rst
p7nov: A copy-paste error, fixed

doc/reference/reference_lua/config/utils_schema.rst
p7nov: Fixed: `rtype: table`, added the table structure

doc/reference/reference_lua/config/utils_schema.rst
p7nov: Added this note where applicable

doc/reference/reference_lua/config/utils_schema.rst
p7nov: Thanks, fixed

doc/reference/reference_lua/config/utils_schema.rst
p7nov: Fixed

doc/reference/reference_lua/config/utils_schema.rst
p7nov: Fixed

doc/reference/reference_lua/config/utils_schema.rst
p7nov: Fixed

doc/reference/reference_lua/config/utils_schema.rst
p7nov: Added

doc/reference/reference_lua/config/utils_schema.rst
p7nov: Fixed

doc/reference/reference_lua/config/utils_schema.rst
p7nov: changed as suggested

doc/reference/reference_lua/config/utils_schema.rst
p7nov: My bad, misinterpreted this.
Rewrote the description as suggested in the doc request.

doc/reference/reference_lua/config/utils_schema.rst
p7nov: Added the return table structure for node constructors

doc/reference/reference_lua/config/utils_schema.rst
p7nov: Replaced the code with a link to a correct example below.

doc/reference/reference_lua/config/utils_schema.rst
Totktonada: I see schema node constructors (scalar, record, map, array), constructor shortcuts (enum, set), schema object constructor (new) and a separate functions (fromenv).

It would be schema node constructors (scalar, record, map, array, enum, set) and other functions (new, fromenv).

I'll resolve this comment -- it is a thing of taste.

doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1251 @@
...
+            -   ``fields``: a table describing the record fields
+            -   annotations, if provided
+
+    :rtype: table
+
+    **See also:** :ref:`config_utils_schema_nodes_record`
+
+..  _config-utils-schema-scalar:
+
+..  function:: schema.scalar(type[, annotations])
```
Totktonada: It is called as follows.

```lua
schema.scalar({
    type = <...>,
    <..annotations..>
})
```

It seems the prototype is incorrect.

p7nov: Thanks, fixed.

doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1262 @@
...
+    ``experimental.config.utils.schema`` is an experimental submodule and is subject to changes.
+
+
+.. _config_utils_schema_getting_started:
+
+Getting started with config.utils.schema
+----------------------------------------
+
+As an example, consider an :ref:`application role <configuration_application_roles>`
+that has a single configuration option - an HTTP endpoint address.
```
lenkis:
```suggestion
that has a single configuration option -- an HTTP endpoint address.
```

doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1262 @@
...
+
+.. _config_utils_schema_nodes:
+
+Schema nodes
+~~~~~~~~~~~~
+
+Schema nodes describe the hierarchy of options within a schema. There are two types of schema nodes:
+
+-   *Scalar* nodes hold a single value of a supported primitive type. For example,
+    a string configuration option of a role is a scalar node its schema.
```
lenkis:
```suggestion
    a string configuration option of a role is a scalar node.
```

doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1262 @@
...
+    :start-at: local listen_address_schema
+    :end-before: local function validate
+    :dedent:
+
+.. _config_utils_schema_annotation:
+
+Annotations
+***********
+
+Node *annotations* are named attributes that define the its various aspects. For example,
```
lenkis:
```suggestion
Node *annotations* are named attributes that define its various aspects. For example,
```

doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1262 @@
...
+..  literalinclude:: /code_snippets/snippets/config/instances.enabled/config_schema_annotations/http_api.lua
+    :language: lua
+    :start-at: scheme = schema.scalar
+    :end-before: host = schema.scalar
+    :dedent:
+
+Node annotations fall into three groups:
+
+-   *Built-in annotations* are handled by the module. These are: ``type``, ``validate``, ``allowed_values``, ``default`` and ``apply_default_if``.
+    Note that ``validate``, ``allowed_values`` are used for validation only. ``default`` and ``apply_default_if`` can transform the configuration.
```
lenkis:
```suggestion
    Note that ``validate`` and ``allowed_values`` are used for validation only. ``default`` and ``apply_default_if`` can transform the configuration.
```

doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1262 @@
...
+^^^^^^^^^^^^^^^^^^^^
+
+Built-in annotations are interpreted by the module itself. There are the following
+built-in annotations:
+
+-   :ref:`type <config-schema_node_annotation-type>` -- the node value type.
+    The type must be explicitly specified for scalar nodes, except for those created with ``schema.enum()``.
+    For composite nodes and scalar enums, the corresponding constructors ``schema.record()``, ``schema.map()``, ``schema.array()``,
+    ``schema.set()``, and ``schema.enum()`` set the type automatically.
+-   :ref:`allowed_values <config-schema_node_annotation-allowed_values>` -- (optional) a list of possible node values .
```
lenkis:
```suggestion
-   :ref:`allowed_values <config-schema_node_annotation-allowed_values>` -- (optional) a list of possible node values.
```

doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1262 @@
...
+    Shortcut for creating an array node of unique string values from the given list of allowed values.
+
+    :param table allowed_values: allowed values of array items
+    :param table annotations: annotations (see :ref:`config_utils_schema_annotation`)
+
+    :return: the created array node as a table with the following fields:
+
+             -   ``type``: ``array``
+             -   ``items``: a table describing an array item as a schema node
+             -   ``validate``: an auto-generated validation function that check
```
lenkis:
```suggestion
             -   ``validate``: an auto-generated validation function that checks
```

doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1262 @@
...
+        ..  literalinclude:: /code_snippets/snippets/config/instances.enabled/config_schema_nodes_record_hierarchy/http_api.lua
+            :language: lua
+            :start-at: local scheme
+            :end-before: local host
+            :dedent:
+
+        :param any data: configuration data
+        :param string/table path: path to the target node as:
+                                  -   a string in the dot notation
+                                  -   an array-like table
```
lenkis, line 945:
```suggestion
                                  - a string in the dot notation
                                  - an array-like table
```

lenkis: The nested bullet list is rendered incorrectly. I suppose, extra spaces are the reason. Please check.

doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1262 @@
...
+            ``value`` is validated by the method before the assignment.
+
+        Set a given value at the given path in a configuration.
+        The path can be either a dot-separated string (``http.scheme``) or
+        an array-like table (``{ 'http', 'scheme'}``).
+
+        :param any data: configuration data
+        :param string/table path: path to the target node as:
+                                  -   a string in the dot notation
+                                  -   an array-like table
```
lenkis, line 1068:
```suggestion
                                  - a string in the dot notation
                                  - an array-like table
```

lenkis: Same as above ^^

doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1262 @@
...
+        :param any data_a: configuration data
+        :param any data_b: configuration data
+
+        :return: merged configuration data
+
+    ..  _config-schema_object-pairs:
+
+    ..  method:: pairs()
+
+        Walk over the schema and return scalar, array, and map schema nodes
```
lenkis, line 1033: Above, in the reference summary, we annotate the function as "Traverse a configuration". "Walk over" and "traverse" are synonyms, but I'd rather not use them interchangeably for annotating the same function in order to avoid confusion for non-native technical readers. I'd vote for "walk over" in both places.

doc/reference/reference_lua/config/utils_schema.rst
```diff
@@ -0,0 +1,1262 @@
...
+    -   ``data`` -- the configuration data
+    -   ``w`` -- *walkthrough node* with the following fields:
+
+        -   ``w.schema`` -- schema node
+        -   ``w.path`` -- the path to the schema node
+        -   ``w.error()`` -- a function for printing human-readable error messages
+
+    **Example:**
+
+    A function that checks that a string a valid IP address:
```
lenkis:
```suggestion
    A function that checks that a string is a valid IP address:
```

doc/platform/app/app_roles.rst
```diff
@@ -44,10 +42,15 @@ This example shows how to enable and configure the ``greeter`` role, which is im
     :start-at: instance001
     :dedent:
 
-The role's configuration provided in ``roles_cfg`` can be accessed when :ref:`validating <roles_create_custom_role_validate>` and :ref:`applying <roles_create_custom_role_apply>` this configuration.
+The role configuration provided in ``roles_cfg`` can be accessed when :ref:`validating <roles_create_custom_role_validate>` and :ref:`applying <roles_create_custom_role_apply>` this configuration.
+
+Tarantool includes the :ref:`experimental.config.utils.schema <config_utils_schema_module>`
+built-in module that provides an for managing user-defined configurations
```
lenkis:
```suggestion
built-in module that provides tools for managing user-defined configurations
```

doc/platform/app/app_roles.rst
```diff
@@ -85,18 +89,37 @@ The examples below show how to do this.
...
+values, for example, ``cfg.greeting``.
 
 .. _roles_create_custom_role_validate:
 
 Validating a role configuration
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-To validate a role's configuration, you need to define the :ref:`validate([cfg]) <roles_api_reference_validate>` function.
+To validate a role configuration, you need to define the :ref:`validate([cfg]) <roles_api_reference_validate>` function.
 The ``cfg`` argument provides access to the :ref:`role's configuration <roles_create_custom_role_config>` and check its validity.
```
lenkis:
```suggestion
The ``cfg`` argument provides access to the :ref:`role's configuration <roles_create_custom_role_config>` and checks its validity.
```

lenkis (changes requested):

doc/platform/app/app_roles.rst
p7nov: This sentence seems irrelevant with the new schema validation module, I'm removing it. Thanks for spotting.

doc/reference/reference_lua/config/utils_schema.rst
p7nov: Added the missing preposition

doc/reference/reference_lua/config/utils_schema.rst
p7nov: Fixed

doc/reference/reference_lua/config/utils_schema.rst
p7nov: Fixed (added a line break before the list)

lenkis (approved):

PR #4663, opened 2024-11-28
Add 3.3 release and what's new

### LINKED ISSUE #4653: Prepare What's new for Tarantool 3.3
**Product:** Tarantool
**Since:** 3.3.0
**Root document:** https://www.tarantool.io/en/doc/latest/release/
**SME:** @ KAPANDR, @ sergepetrenko 

# Details

- [ ] Prepare a what's new page before the release: https://www.tarantool.io/en/doc/latest/release/3.3.0/
- [ ] Add a release date and a link to GitHub after the release
- [x] Add a new release to the https://www.tarantool.io/en/doc/latest/release/ page

##  Features to cover in What's new

From @ sergepetrenko 

- **offset_of for indexes and spaces, offset counting for O(log(size)); same for read views in EE** (https://github.com/tarantool/tarantool/issues/8204, https://github.com/tarantool/tarantool-ee/issues/858).
  Related changelogs:
  - Improved performance of tree index methods: `select()` with the `offset`
    specified and `count()` method. The underlying algorithm for these methods is
    changed: the old one's time complexity was `O(n)`, where `n` is the value of
    `offset` or the amount of counted tuples. The new algorithm's complexity is
    `O(log(size))`, where `size` is the number of tuples stored in the index. Now
    it does not depend on the `offset` value or the amount of tuples to count. It
    is safe to use these functions with arbitrary big offset values and tuple
    count (gh-8204).
  - Introduced a new `offset` parameter in the `index:pairs` method. It allows to
    skip the first tuples of the iterator (gh-8204).
  - Introduced a new `index:offset_of` method. It allows to get the position in
    the index relative to the iterator direction of a tuple matching the given
    key and iterator (gh-8204).
  - Introduced the new `count` method of an index and space read view. The method
    returns the amount of tuples in the index (or primary key of a space) matching
    the given key and iterator type (ee-858).
  - Improved performance of memtx tree index read view `select()` method with
    the `offset` specified. The underlying algorithm for the method is changed.
    The time complexity of the old algorithm was `O(n)`, where `n` is the value
    of `offset`. The new algorithm's complexity is `O(log(size))`, where `size`
    is the number of tuples stored in the index. Now it does not depend on the
    `offset` value. It is safe to use the function with arbitrary big offset
    values (ee-858).
  - Introduced a new `offset` parameter in the `index_read_view:pairs` method. It
    allows to skip the first tuples of the iterator (ee-858).
  - Introduced a new `index_read_view:offset_of` Lua API. It allows to get the
    0-based offset in the index relative to the iterator direction of a tuple
    that matches the given key and iterator type (ee-858).
- **persistent gc state of replicas** (https://github.com/tarantool/tarantool/issues/10154). Changelog:
  - Now all replicas have WAL GC consumers persisted in the `_gc_consumers`
     space (gh-10154). The `wal_cleanup_delay` option is no longer needed,
     so it is deprecated.
- **don't rollback synchronous transaction upon reaching replication_synchro_timeout** 
  (https://github.com/tarantool/tarantool/issues/7486). Changelogs:
  - A new compat option `compat.replication_synchro_timeout` has been added.
     This option determines whether the `replication.synchro_timeout` option rolls
     back transactions. When set to 'new', transactions are not rolled back due to
     a timeout. In this mode `replication.synchro_timeout` is used to wait
     confirmation in promote/demote and gc-checkpointing. If 'old' is set, the
     behavior is no different from what it was before this patch appeared.
  - A new `replication.synchro_queue_max_size` option limits the number of
     transactions in the master synchronous queue.
    `replication.synchro_queue_max_size` is measured in the number of bytes to be
     written (0 means unlimited, which was the default behavior before).
     Currently, this option defaults to 16 megabytes. (gh-7486)
- **C API for sending work to TX thread**. Changelog:
   - The new public C API functions `tnt_tx_push(callback)` and `tnt_tx_flush()`
      allow to send work to TX thread from any other thread.

p7nov:
Resolves #4653 

- Add [Tarantool 3.3 what's new](https://docs.d.tarantool.io/en/doc/gh-4653-whatsnew-3-3/release/3.3.0/)
- Add 3.3 to the [release calendar](https://docs.d.tarantool.io/en/doc/gh-4653-whatsnew-3-3/release/#supported-versions)
- Add 3.3 to the list of [supported versions](https://docs.d.tarantool.io/en/doc/gh-4653-whatsnew-3-3/release/#supported-versions)

Deployment: https://docs.d.tarantool.io/en/doc/gh-4653-whatsnew-3-3/release/3.3.0/


doc/release/3.3.0.rst
```diff
@@ -0,0 +1,375 @@
...
+*   **Community Edition (CE)**
+
+    *   Improvements around queries with offsets.
+    *   Improvement in Raft implementation.
+    *   Persistent replication state.
+    *   New C API for sending work to the TX thread.
+
+*   **Enterprise Edition (EE)**
+
+    *   Offset-related improvements in read views.
```
Totktonada: Should the newly added features be here as well?

doc/release/3.3.0.rst
```diff
@@ -0,0 +1,375 @@
...
+
+
+.. _3-3-features-for-developers:
+
+Developing applications
+-----------------------
+
+* JSON-схема для кластерного конфига.
+* В ролях появился коллбек on_event для реакции на RO/RW.
+* Возможность поднять или убрать алерт из роли или приложения.
```
Totktonada: Is it something temporary?

doc/release/3.3.0.rst
```diff
@@ -0,0 +1,375 @@
...
+
+Additionally, there is the `latest <https://download.tarantool.org/tarantool/schema/config.schema.json>`__
+schema that reflects the latest configuration schema in development (master branch).
+
+You can use the URIs of these schemas to validate you configuration with `check-jsonschema <https://pypi.org/project/check-jsonschema/>`__
+as follows:
+
+.. code-block:: console
+
+    $ check-jsonschema --schemafile https://download.tarantool.org/tarantool/schema/config.schema.3.3.0.json config.yaml
```
Totktonada: Ability to check user's config is good, but what is really nice: we can add this schema into an IDE's YAML plugin that supports JSON Schemas (at least for VSCode has one) and get completion for the options and hints with the documentation.

This is more important for UX than the validation. I would include both scenarios, but if we have place only for one, let's drop the `check-jsonschema` one.

doc/release/3.3.0.rst
```diff
@@ -0,0 +1,375 @@
...
+
+    local config = require('config')
+    local alerts = config:new_alerts_namespace('my_alerts')
+
+Alerts namespaces provides methods for managing alerts within them. All user-defined
+alerts raised in all namespaces are shown in ``box.info.config.alerts``.
+
+To raise an alert, use the namespace methods ``add`` or ``set``:
+The difference between them is that ``set()`` accepts a key to use as an alert
+identifier for further operations. An alert must be a table; it's ``message`` field is logged.
```
Totktonada: > The difference between them is that `set()` accepts a key to use as an alert identifier for further operations.

The key point behind `set()` is that you can raise an alert several times with the same key, but only the last one is saved. Ability to call `unset()` later is a good thing, but it seems less important.

doc/release/3.3.0.rst
```diff
@@ -0,0 +1,375 @@
...
+.. _3-3-upgrade-ddl:
+
+DDL before upgrade
+~~~~~~~~~~~~~~~~~~~
+
+* Апгрейд лишился ограничений на DDL по сравнению с версиями 3.0, 3.1, 3.2.
+Это позволяет, например, выдать права на функции до вызова box.schema.upgrade()..
+
+Tarantool now allows DDL operations before calling ``box.schema.upgrade()`` during
+an upgrade from an earlier version. For example, you can
```
Totktonada: It seems, this section is not finished.

doc/release/3.3.0.rst
```diff
@@ -0,0 +1,375 @@
...
+Use the isolated mode to temporarily isolate instances for maintenance, debugging,
+or other actions that should not affect other cluster instances.
+
+.. _3-3-replica-unregister:
+
+Automatic expulsion configuration
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A new configuration section ``replication.autoexpel`` allows to automatically expel
+instances after they are removed from the YAML configuration.
```
Totktonada: It is OK to use `expulse` and `expel` both? I would attempt to use one term for one feature.

doc/release/3.3.0.rst
```diff
@@ -0,0 +1,375 @@
...
+The section includes three options:
+
+-   ``enabled``: whether automatic expulsion logic is enabled in the cluster.
+-   ``by``: a criterion for selecting instances that can be expelled automatically.
+    In version 3.3, the only available criterion is ``prefix``.
+-   ``prefix``: a prefix with which an instance name should start to make automatic expulsion possible.
+
+.. note::
+
+    Automatic expulsion happens after a :ref:`configuration reload <config_api_reference_reload>`.
```
Totktonada: From the [commit message](https://github.com/tarantool/tarantool/commit/b2c11535d87d4cb0e9c02fbe53d57887f8887fcd):

> The autoexpel logic works on startup and reacts on the reconfiguration and the `box.status` watcher event.

If we say that it reacts on reload, it is incomplete and may confuse. Maybe better to say nothing in such a brief description.

doc/release/3.3.0.rst
```diff
@@ -0,0 +1,375 @@
...
+Lua memory size
+~~~~~~~~~~~~~~~
+
+A new configuration option ``lua.memory`` specifies the maximum amount of memory
+for Lua scripts execution, in bytes. For example, this configuration sets the Lua memory limit to 2 GB:
+
+.. code-block:: yaml
+
+    lua:
+        memory: 2147483648
```
Totktonada: This is the default, so there is no sense to set the limit to 2GiB. We can given an example that reduces the limit or uplifts it.

Totktonada (approved):
Aside of some parts that look in-progress everything seems correct. I left some non-blocking remarks, though.

doc/release/3.3.0.rst
p7nov: Most dictionaries state that `expulsion` is the noun form of `expel` (a surprise for me), so it's the same term.

Totktonada: Hm, OK so.

doc/release/3.3.0.rst
```diff
@@ -1,7 +1,7 @@
 Tarantool 3.3
 =============
 
-Release date: November 29, 2024
+Release date: November 30, 2024
```
Totktonada: It was released 29 Nov.

doc/release/index.rst
```diff
@@ -62,8 +62,8 @@ For information about earlier versions, see :doc:`eol_versions`.
             -   Versions
 
         *   -   :doc:`3.3 </release/3.3.0>`
-            -   **November 29, 2024**
-            -   **November 29, 2026**
+            -   **November 30, 2024**
+            -   **November 30, 2026**
```
Totktonada: Same remark: the release appeared 29 Nov.

doc/release/3.3.0.rst
p7nov: Thanks, fixed

doc/release/index.rst
```diff
@@ -61,6 +61,12 @@ For information about earlier versions, see :doc:`eol_versions`.
             -   End of support
             -   Versions
 
+        *   -   :doc:`3.3 </release/3.3.0>`
+            -   **November 29, 2024**
+            -   **November 29, 2026**
+            -   **Not planned yet**
+            -   | :tarantool-release:`3.3.0`
+
```
sergepetrenko, line 69: This doesn't concern this particular change, but I doubt we'll provide updates and fixes for 3.1 and 3.0 until April 2026/December 2025. We don't do that already.

Am I misunderstanding something?
CC @Totktonada

doc/release/3.3.0.rst
```diff
@@ -0,0 +1,375 @@
...
+Releases on GitHub: :tarantool-release:`3.3.0`
+
+The 3.3 release of Tarantool adds the following main product features and improvements for the Community and Enterprise editions:
+
+*   **Community Edition (CE)**
+
+    *   Improvements around queries with offsets.
+    *   Improvement in Raft implementation.
+    *   Persistent replication state.
+    *   New C API for sending work to the TX thread.
```
sergepetrenko: I'd add '... from user threads.'

sergepetrenko (approved):
Thanks!
Looks good to me.

doc/release/index.rst
Totktonada: The object of support is the 3.x series at whole. I guess it is what actually meant here, but the representation is confusing. (Likely it is not a problem of this particular 'what's new' article, but a structure of our release informational docs at all.)

p7nov: OK, thanks for spotting, we'll recheck this representation with PMs later to avoid blocking this PR.

doc/release/3.3.0.rst
```diff
@@ -0,0 +1,375 @@
...
+Release date: November 29, 2024
+
+Releases on GitHub: :tarantool-release:`3.3.0`
+
+The 3.3 release of Tarantool adds the following main product features and improvements for the Community and Enterprise editions:
+
+*   **Community Edition (CE)**
+
+    *   Improvements around queries with offsets.
+    *   Improvement in Raft implementation.
```
lenkis, line 13: (Here and same term below) I see 2 ways of capitalization for Raft in the doc: Raft and RAFT. 
https://www.tarantool.io/en/doc/latest/search/?q=raft 
All caps seems more popular -- and "more correct" as for me.

doc/release/3.3.0.rst
```diff
@@ -0,0 +1,375 @@
...
+
+Tarantool 3.3 brings a number of improvements around queries with offsets.
+
+-   The performance of tree index :ref:`select() <box_index-select>` with offset and
+    :ref:`count() <box_index-count>` methods was improved.
+    Previously, the algorithm complexity had a linear dependency on the
+    provided offset size (``O(offset)``) or the number of tuples to count. Now,
+    the new algorithm complexity is ``O(log(size))`` where ``size`` is the number of tuples
+    in the index. This change also eliminates the dependency on the offset value or
+    the number of tuple to count.
```
lenkis:
```suggestion
    the number of tuples to count.
```

doc/release/3.3.0.rst
```diff
@@ -0,0 +1,375 @@
...
+-   The :ref:`index <box_index>` and :ref:`space <box_space>` entities get the new
```
lenkis:
```suggestion
-   The :ref:`index <box_index>` and :ref:`space <box_space>` entities get a new
```

doc/release/3.3.0.rst
```diff
@@ -0,0 +1,375 @@
...
+
+No rollback on timeout for synchronous transactions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To better match the canonical Raft algorithm design, Tarantool no longer rolls
+back synchronous transactions on timeout (upon reaching :ref:`replication.synchro_timeout <cfg_replication-replication_synchro_timeout>`).
+In the new implementation, transactions can only be rolled back by a new leader after it is elected.
+Otherwise, they can wait for a quorum infinitely.
+
+Given that this changes the behavior, a new ``replication_synchro_timeout`` :ref:`compat <compat-module>` option is introduced.
```
lenkis:
```suggestion
Given this change in behavior, a new ``replication_synchro_timeout`` :ref:`compat <compat-module>` option is introduced.
```

doc/release/3.3.0.rst
```diff
@@ -0,0 +1,375 @@
...
+
+-   In Lua code:
+
+    ..  code-block:: tarantoolsession
+
+        tarantool> require('compat').replication_synchro_timeout = 'new'
+        ---
+        ...
+
+There is also the new ``replication.synchro_queue_max_size`` configuration option
```
lenkis:
```suggestion
There is also a new ``replication.synchro_queue_max_size`` configuration option
```

doc/release/3.3.0.rst
```diff
@@ -0,0 +1,375 @@
...
+that limits the total size of transactions in the master synchronous queue. By default,
+it's 16 megabytes.
```
lenkis:
```suggestion
that limits the total size of transactions in the master synchronous queue. The default
value is 16 megabytes.
```

doc/release/3.3.0.rst
```diff
@@ -0,0 +1,375 @@
...
+
+.. _3-3-c-api-tx-thread:
+
+C API for sending work to TX thread
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The new public C API functions ``tnt_tx_push()`` and ``tnt_tx_flush()``
```
lenkis:
```suggestion
New public C API functions ``tnt_tx_push()`` and ``tnt_tx_flush()``
```

doc/release/3.3.0.rst
```diff
@@ -0,0 +1,375 @@
...
+allow to send work to the :ref:`TX thread <thread_model>` from any other thread:
+
+-   ``tnt_tx_push()`` schedules the given callback to be executed with the provided
+    arguments.
+
+-   ``tnt_tx_flush()`` sends all the pending callback for execution in the TX thread.
```
lenkis:
```suggestion
-   ``tnt_tx_flush()`` sends all pending callbacks for execution in the TX thread.
```

doc/release/3.3.0.rst
```diff
@@ -0,0 +1,375 @@
...
+    The execution is started in the same order as the callbacks were pushed.
```
lenkis:
```suggestion
    Execution is started in the same order as the callbacks were pushed.
```

doc/release/3.3.0.rst
```diff
@@ -0,0 +1,375 @@
...
+-   `3.1.1 <https://download.tarantool.org/tarantool/schema/config.schema.3.1.1.json>`__
+-   `3.1.2 <https://download.tarantool.org/tarantool/schema/config.schema.3.1.2.json>`__
+-   `3.2.0 <https://download.tarantool.org/tarantool/schema/config.schema.3.2.0.json>`__
+-   `3.2.1 <https://download.tarantool.org/tarantool/schema/config.schema.3.2.1.json>`__
+-   `3.3.0 <https://download.tarantool.org/tarantool/schema/config.schema.3.3.0.json>`__
+
+Additionally, there is the `latest <https://download.tarantool.org/tarantool/schema/config.schema.json>`__
+schema that reflects the latest configuration schema in development (master branch).
+
+Use these schemas to add completion for YAML configuration files and get
```
lenkis:
```suggestion
Use these schemas to add code completion for YAML configuration files and get
```

doc/release/3.3.0.rst
```diff
@@ -0,0 +1,375 @@
...
+the ``config:jsonschema()`` function.
+
+.. _3-3-roles-on-event:
+
+on_event callbacks in roles
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Now :ref:`application roles <application_roles>` can have ``on_event`` callbacks.
+They are executed every time a ``box.status`` :ref:`system event <system-events>` is
+broadcasted or the configuration is updated. The callback has three arguments:
```
lenkis: I see that in the doc we opted for "broadcast", not "broadcasted" as the verb's 2nd and 3rd forms: https://www.tarantool.io/en/doc/latest/search/?q=broadcasted

doc/release/3.3.0.rst
```diff
@@ -0,0 +1,375 @@
...
+
+-   ``config`` -- the current configuration.
+-   ``key`` -- an event that has triggered the callback: ``config.apply`` or ``box.status``.
+-   ``value`` -- is the value of the ``box.status`` :ref:`system event <system-events>`.
```
lenkis:
```suggestion
-   ``value`` -- the value of the ``box.status`` :ref:`system event <system-events>`.
```

doc/release/3.3.0.rst
```diff
@@ -0,0 +1,375 @@
...
+
+    local config = require('config')
+    local alerts = config:new_alerts_namespace('my_alerts')
+
+Alerts namespaces provide methods for managing alerts within them. All user-defined
+alerts raised in all namespaces are shown in ``box.info.config.alerts``.
+
+To raise an alert, use the namespace methods ``add()`` or ``set()``:
+The difference between them is that ``set()`` accepts a key to refer to the alert
+later: overwrite or discard it. An alert is a table with one mandatory  field ``message``
```
lenkis:
```suggestion
later: overwrite or discard it. An alert is a table with one mandatory field ``message``
```

doc/release/3.3.0.rst
```diff
@@ -0,0 +1,375 @@
...
+before the schema is upgraded.
+
+.. _3-3-isolated-instances:
+
+Isolated instances
+~~~~~~~~~~~~~~~~~~
+
+A new instance-level configuration option ``isolated`` puts an instance into the
+*isolated* mode. In this mode, an instance doesn't accept updates from other members
+of its replica set and other iproto requests. It also doesn't perform background
```
lenkis:
```suggestion
of its replica set and other iproto requests. It also performs no background
```

doc/release/3.3.0.rst
```diff
@@ -0,0 +1,375 @@
...
+
+        failover:
+          replicasets:
+            replicaset-001:
+              priority:
+                instance-001: 5
+                instance-002: -5
+                instance-003: 4
+
+    Additionally, there is ``failover.learners`` section that lists instances
```
lenkis:
```suggestion
    Additionally, there is a ``failover.learners`` section that lists instances
```

doc/release/3.3.0.rst
```diff
@@ -0,0 +1,375 @@
...
+    and ``failover.log.file``:
+
+    .. code-block:: yaml
+
+        failover:
+          log:
+            to: file # or stderr
+            file: var/log/tarantool/failover.log
+
+Learn more about supervised failover :ref:`repl_supervised_failover`.
```
lenkis:
```suggestion
Learn more about supervised failover: :ref:`repl_supervised_failover`.
```

doc/release/3.3.0.rst
```diff
@@ -0,0 +1,375 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Tarantool :ref:`persistence mechanism <concepts-data_model-persistence>` uses
+two types of files: snapshots and write-ahead log (WAL) files. These files are also used
+for replication: read-only replicas receive data changes from the replica set leader
+by reading these files.
+
+The :ref:`garbage collector <configuration_persistence_garbage_collector>` that
+cleans up obsolete snapshots and WAL files it doesn't remove the files while they
+are in use for replication. To make such a check possible, the replica set leaders
```
lenkis:
```suggestion
The :ref:`garbage collector <configuration_persistence_garbage_collector>`
cleans up obsolete snapshots and WAL files, but it doesn't remove the files while they
are in use for replication. To make such a check possible, the replica set leaders
```

doc/release/3.3.0.rst
```diff
@@ -0,0 +1,375 @@
...
+store the replication state in connection with files. However, this information
+was not persisted, which could lead to issues in case of the leader restart.
+The garbage collector could delete WAL files after the restart even if there were
+replicas that still read these files. The :ref:`wal.cleanup_delay <configuration_reference_wal_cleanup_delay>`
+configuration option was used to prevent such situations.
+
+Since version 3.3, leader instances persist the information about WAL files in use
+in the new system space ``_gc_consumers``. After a restart, the replication state
```
lenkis:
```suggestion
in a new system space ``_gc_consumers``. After a restart, the replication state
```

lenkis (changes requested):

doc/release/index.rst
sergepetrenko: Sure.

doc/release/3.3.0.rst
p7nov: I've checked the [Raft website](https://raft.github.io/), [Wikipedia](https://en.wikipedia.org/wiki/Raft_(algorithm)), and some [docs from software vendors](https://developer.hashicorp.com/consul/docs/architecture/consensus), and haven't seen the all caps spelling. Sticking to the original (Raft website) is usually correct.

doc/release/3.3.0.rst
p7nov: Thanks, I didn't know that this second form is also correct.

doc/release/3.3.0.rst
lenkis: Would be nice to ask our devs to update log messages then: they mostly say "RAFT".

lenkis (approved):

PR #5082, opened 2025-04-18
Add a detailed review of the key features in release 3.4

lenkis:


doc/release/3.4.0.rst
```diff
@@ -0,0 +1,249 @@
...
+
+.. _3-4-native:
+
+[EE] New boostrap strategy in failover: ``native``
+--------------------------------------------------
+
+Now supervised failover coordinator supports three bootstrap strategies:
+native, supervised, auto.
+
+The new ``native `` strategy acts more or less similar to the ``auto`` strategy,
```
Totktonada: Nit: GitHub highlight this part is a bit strage way due to a whitespace before double backtick (after `native`). Not sure how it is rendered by Sphinx, but maybe it worth to drop the whitespace to be on the safe side.

doc/release/3.4.0.rst
```diff
@@ -0,0 +1,249 @@
...
+but relaxes its limitations. It is based on the supervised strategy and basically
+performs two things:
+* issues ``box.ctl.make_bootstrap_leader({graceful = true})`` to bootstrap
+  a replicaset;
+* issues ``box.ctl.make_bootstrap_leader()`` to keep the bootstrap leader
+  record pointing to the instance that is currently in the RW mode (to register
+  new replicas).
+
+To enable the ``native `` bootstrap strategy, set it in the ``replication`` section
```
Totktonada: Nit: Same here regarding the whitespace.

doc/release/3.4.0.rst
```diff
@@ -0,0 +1,249 @@
...
+native, supervised, auto.
+
+The new ``native `` strategy acts more or less similar to the ``auto`` strategy,
+but relaxes its limitations. It is based on the supervised strategy and basically
+performs two things:
+* issues ``box.ctl.make_bootstrap_leader({graceful = true})`` to bootstrap
+  a replicaset;
+* issues ``box.ctl.make_bootstrap_leader()`` to keep the bootstrap leader
+  record pointing to the instance that is currently in the RW mode (to register
+  new replicas).
```
Totktonada: I'm not sure that the calls under the hood are really interesting for our reader. Maybe some motivational things would work better here.

Quoted from my mini-announce in a team chat (in Russian):

> Очень похоже на дефолтную auto, но под капотом работает по-другому (поверх стратегии supervised). Цель была в том, чтобы решить следующие проблемы:
>
> * Чтобы не стреляла ошибка Some replica set members were not specified in box.cfg.replication при одновременном подключении нескольких реплик, при наличии неанонимных CDC в репликасете, при наличии старых реплик в _cluster.
> * Чтобы база бутстрапилась по команде координатора, а не инстансами самостоятельно <...>.

doc/release/3.4.0.rst
```diff
@@ -0,0 +1,249 @@
...
+[CE] New methods in the ``config`` module to check instance roles
+-----------------------------------------------------------------
+
+Three new methods are now available in the ``config`` module:
+
+* ``has_role(<role_name>, {instance = <instance_name})`` returns ``true`` if
+  the instance with the name ``<instance_name>`` has the role ``<role_name>``
+  enabled in the current configuration, or ``false`` if not.
+  The second argument is optional: if not provided, the check is performed
+  for the instance the method is called on.
```
Totktonada: Nit: It feels a bit too formal for the announcement. Would look appropriate in an API documentation, but here I would make it simpler:

> `config:has_role('myrole')` tells whether the current instance has role `myrole`. `config:has_role('myrole', {instance = 'i-001'})` do the same for the given instance.

Same for `config:is_router()`, `config:is_storage()`.

Totktonada (approved):
Thanks! Nice list :)

doc/release/3.4.0.rst
lenkis: @Totktonada Please approve this text:

```
The new ``native`` strategy relaxes the limitations of the ``auto`` strategy,
but has different under-the-hood implementation (based on the ``supervised`` strategy).
Otherwise, it acts similar to the ``auto`` strategy.

In effect, it helps resolve these two problems:
* Avoid the error ``Some replica set members were not specified in box.cfg.replication``
  in the following cases:
  * several replicas join at the same time,
  * the replica set includes non-anonymous CDC instances,
  * ``_cluster`` contains expelled replicas.
* Make the database get bootstrapped upon the coordinator's command rather than
  let the instances boostrap it on their own.
```

doc/release/3.4.0.rst
lenkis: @Totktonada Please approve the text:

```
Three new methods are now available in the ``config`` module:

* ``config:has_role('myrole')`` tells whether the current instance has the role ``myrole``, and
  ``config:has_role('myrole', {instance = 'i-001'})`` does the same for the specified instance (``i-001``).

* ``config:is_router()`` tells whether the current instance is a vshard router, and
  ``config:is_router({instance = 'i-002'})`` does the same for the specified instance (``i-002``).

* ``config:is_storage()`` tells whether the current instance is a vshard storage, and
  ``config:is_storage({instance = 'i-003'})`` does the same for the specified instance (``i-003``).
```

doc/release/3.4.0.rst
Totktonada: Nice! The only remark: in my understanding an instance is expelled, when it is removed from `_cluster`. So, I would say '`_cluster` contains old unneeded replicas'.

doc/release/3.4.0.rst
Totktonada: I'm OK.

doc/release/3.4.0.rst
```diff
@@ -0,0 +1,244 @@
...
+--------------------------------------------------
+
+Now supervised failover coordinator supports three bootstrap strategies:
+native, supervised, auto.
+
+The new ``native`` strategy relaxes the limitations of the ``auto`` strategy,
+but has different under-the-hood implementation (based on the ``supervised`` strategy).
+Otherwise, it acts similar to the ``auto`` strategy.
+
+In effect, it helps resolve these two problems:
```
hackallcode, line 130: It seems that the new strategy is better than the auto strategy, isn't it? Or is the automatic strategy better in some cases?

lenkis: Выяснила подробности. предлагаю вот так:

```
The ``native`` strategy is the recommended choice for highly dynamic clusters with
automatic scaling, as well as in most other cases.
The only scenario when this strategy may work worse than ``auto`` is when a current
bootstrap leader in a multi-master configuration goes off or becomes inaccessible.
Another RW instance doesn't pick up this role automatically and some RW -> RO -> RW
configuration changes are needed to make another instance take the role. In this regard,
the ``native`` strategy seems worse than ``auto``.
```

Totktonada: We hope that the new strategy works better in all the cases except a multi-master setup.

I think that describing certain problems, when `auto` is better, is overkill for the announcement.

> The only scenario when this strategy may work worse than ``auto`` is when <...>

I propose to replace it with:

> The new strategy is the recommended choice for all the cases except multi-master.

I think that's enough.

lenkis: По итогу оставила только рекомендацию, когда хорошо использовать native, и убрала все подробности про нерабочий сценарий.

PR #5136, opened 2025-05-27
Adds isolated instance mode description

### LINKED ISSUE #4632: config: isolated mode for an instance
**Related dev. issue(s):** https://github.com/tarantool/tarantool/issues/10796

**Product:** Tarantool
**Since:** 3.3.0
**Audience/target:** application developers, administrators
**Root document:** 
- (possibly somewhere in) https://www.tarantool.io/en/doc/latest/admin/replication/
**SME:** @ Totktonada

# Details

The option is to temporary isolate an instance to perform replicaset repairing activities: debug a problem on the isolated instance (without affecting the non-isolated part) or extract some data from the isolated instance and apply them on the non-isolated part of the replicaset.

```yaml
groups:
  g:
    replicasets:
      r:
        instances:
          i-001: {}
          i-002: {}
          i-003: {}
          i-004:
            isolated: true
```

In this mode the instance doesn't perform processing of iproto requests, doesn't perform background data modifications and refuses to accept updates from other replicaset members. Also, other replicaset members don't replicate data from the isolated instance.

Technically speaking, when the option is enabled, the following actions are performed:

1. iproto stops listening for new connections.
2. iproto drops all the current connections.
3. The instance goes to read-only mode.
4. The instance is disconnected from all the replication upstreams.
5. Other replicaset members exclude the isolated instance from the replication upstreams.

One new startup constraint is added in the isolated mode:

1. The instance can't be bootstrapped (a local snapshot is required to start).

# Definition of done

- [ ] Describe the new option in the configuration reference
- [ ] Verify whether anything else should be updated in the documentation

AArdeev:
* The ```isolated:true``` option moves the instance to RO mode 
* In this mode, various debugging & maintanance operations can be performed
Fixes #4632


doc/reference/configuration/configuration_reference.rst
```diff
@@ -2836,6 +2836,39 @@ Attributes must be ``key: value`` pairs with string keys and values.
...
+``isolated:true`` moves the instance it has been applied at to the **isolated** mode.
+
+The isolated mode allows the user to temporarily isolate an instance and perform maintenance activities on it. 
+
+The isolated mode moves the instance to the read-only state, in which:
+
+*   iproto stops listening for new connections
+*   iproto drops all the current connections
+*   The instance is disconnected from all the replication upstreams
+*   Other replicaset members exclude the isolated instance from the replication upstreams
```
Totktonada: This wording seems a bit misleading: 'read-only state, which stops iproto and disconnects replication'. In fact, the isolated state/mode sets the read-only state, stops iproto and so on.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -2836,6 +2836,39 @@ Attributes must be ``key: value`` pairs with string keys and values.
...
+
+Once the mode is enabled, any request trying to change persistent data will fail with the ``ER_READONLY`` error.
```
Totktonada: A remote request likely just fails, because we drop iproto connections. Not sure that we should add a note regarding `ER_READONLY`.

doc/code_snippets/snippets/config/instances.enabled/isolated_mode/config.yaml
```diff
@@ -0,0 +1,10 @@
+groups:
+  g:
+    replicasets:
+      r:
+        instances:
+          i-001: {}
+          i-002: {}
+          i-003: {}
+          i-004:
+            isolated: true
```
Totktonada: Nit: GitHub shows a red circle, when there is no newline symbol at the end of the file.

image

doc/admin/troubleshoot.rst
```diff
@@ -254,6 +254,10 @@ This issue can be fixed in two ways:
 -   Manually: :ref:`reseed <replication-master-master-reseed-replica>` one master from another by removing write-ahead logs and snapshots.
 -   Programmatically: set up a :ref:`conflict resolution trigger <replication-problem_solving>`.
 
+..  NOTE::
+
+    To avoid brain split between two masters, one of the instances may be put in the :ref:`isolated mode <_configuration_reference_isolated_mode>`.
```
Totktonada: It gives a false impression that the isolated mode somehow helps to avoid split-brain, but it means that we should detect is somehow faster than tarantool itself. It seems strange.

Here we can fight with consequences of split-brain to some extent. At least, isolate the broken instance from client requests using the config option.

Let's discuss how to better express it.

doc/reference/configuration/configuration_reference.rst
AArdeev: Fixed

doc/reference/configuration/configuration_reference.rst
AArdeev: Fixed

doc/code_snippets/snippets/config/instances.enabled/isolated_mode/config.yaml
AArdeev: fixed

doc/admin/troubleshoot.rst
AArdeev: updated

doc/reference/configuration/configuration_reference.rst
```diff
@@ -2836,6 +2836,38 @@ Attributes must be ``key: value`` pairs with string keys and values.
 
     See also: :ref:`configuration_labels`
 
+..  _configuration_reference_isolated_mode:
```
Totktonada, line 2808: Forgot to say it on the previous review iteration: AFAIR, options/sections on each level are sorted alphabetically. Here `isolated` goes after `labels`.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -2836,6 +2836,38 @@ Attributes must be ``key: value`` pairs with string keys and values.
...
+
+isolated mode
+------
+
+Since version :doc:`3.3.0 </release/3.3.0>`, a new ``isolated`` option is added to instance configuration.
+
+The option takes boolean values, by default it is set to ``false``.
+``isolated:true`` moves the instance it has been applied at to the **isolated** mode.
```
Totktonada: Nit: `isolated: true` looks better for me (comparing to `isolated:true`).

Totktonada (approved):
Thank you for working on it!

PR #5149, opened 2025-05-30
Adds description of ```lua_call`` option specifics

### LINKED ISSUE #4552: Grant runtime access to Lua functions via configuration
**Product:** Tarantool
**Since:** 3.3
**Root document:** 
  
**SME:** @ mandesero @ Totktonada 

# Details

To be done together with #4462.

It is now possible to grant execution privileges for Lua functions
through the declarative configuration, even when the database is in
read-only mode or has an outdated schema version. You might also
permit `guest` to execute Lua functions before the initial bootstrap.

You can specify function permissions using the `lua_call` option in
the configuration, for example:

```yaml
credentials:
  users:
    alice:
      privileges:
        - permissions: [execute]
          lua_call: [my_func]
```

This grants the `alice` user permission to execute the `my_func` Lua
function, regardless of the database's mode or status. The special option
`lua_call: [all]` is also supported, granting access to all global Lua
functions except built-in ones, bypassing database restrictions.

Privileges will still be written to the database when possible to
maintain compatibility and consistency with other privilege types.
Requested by @mandesero in https://github.com/tarantool/tarantool/commit/38c6b0d38254d5ebde82cd469366955e808798b0.

AArdeev:
* ``lua_call`` option allows the specified user to perform the specified lua function on the instance
* permissions can be granted to any user registered on the instance
* ``lua_call: [all]`` grants access to all global Lua functions except built-in ones
* grants are valid until instance is rebooted
* Fixes #4552

### DIFF
--- a/doc/reference/configuration/configuration_reference.rst
+++ b/doc/reference/configuration/configuration_reference.rst
@@ -1453,13 +1453,10 @@ credentials.users.*
     This option should be configured together with the ``execute``
     :ref:`permission <configuration_reference_credentials_privileges_permissions>`.
 
-    To take a previously granted privilege away, specify the ``lua_call`` option without the function name, 
-    or with an empty privileges array (furhter options may retain commented-out).
-
     Since version :doc:`3.3.0 </release/3.3.0>`, the ``lua_call`` option allows granting users privileges to call specified lua function on 
-    the instance in runtime.
+    the instance in runtime (thus it doesn't require an ability to write to the database).
 
-    Example to grant and take away custom functions from the 'alice' user:
+    Example to grant custom functions from the 'alice' user:
 
     ..  literalinclude:: /code_snippets/snippets/config/instances.enabled/lua_call/config.yaml
         :language: yaml
--- a/doc/reference/configuration/configuration_reference.rst
+++ b/doc/reference/configuration/configuration_reference.rst
@@ -1456,7 +1456,7 @@ credentials.users.*
     Since version :doc:`3.3.0 </release/3.3.0>`, the ``lua_call`` option allows granting users privileges to call specified lua function on 
     the instance in runtime (thus it doesn't require an ability to write to the database).
 
-    Example to grant custom functions from the 'alice' user:
+    Example to grant custom functions to the 'alice' user:
 
     ..  literalinclude:: /code_snippets/snippets/config/instances.enabled/lua_call/config.yaml
         :language: yaml
--- a/doc/reference/configuration/configuration_reference.rst
+++ b/doc/reference/configuration/configuration_reference.rst
@@ -1447,10 +1447,23 @@ credentials.users.*
 .. confval:: <user_or_role_name>.privileges.lua_call
 
     A list of global user-defined Lua functions that this user or a user with this role can call.
-    To allow calling all such functions, specify the ``all`` value.
+    To allow calling a specific function, specify its name as the value.
+    To allow calling all global Lua functions except built-in ones functions, specify the ``all`` value.
 
     This option should be configured together with the ``execute``
     :ref:`permission <configuration_reference_credentials_privileges_permissions>`.
+
+    To take a previously granted privilege away, specify the ``lua_call`` option without the function name, 
+    or with an empty privileges array (furhter options may retain commented-out).
+
+    Since version :doc:`3.3.0 </release/3.3.0>`, the ``lua_call`` option allows granting users privileges to call specified lua function on 
+    the instance in runtime.
+
+    Example to grant and take away custom functions from the 'alice' user:
+
+    ..  literalinclude:: /code_snippets/snippets/config/instances.enabled/lua_call/config.yaml
+        :language: yaml
+        :dedent:
 
 ..  _configuration_reference_credentials_privileges_sql:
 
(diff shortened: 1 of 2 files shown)


doc/reference/configuration/configuration_reference.rst
```diff
@@ -3220,6 +3220,26 @@ The ``lua`` section outlines the configuration parameters related to the Lua env
     | Default: 2147483648 (2GB)
     | Environment variable: TT_LUA_MEMORY
 
+.. _configuration_reference_lua_call:
+
+.. confval:: lua_call
```
Totktonada: Is it on the top level? The deployment fails, so I can't check how it is rendered.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -3220,6 +3220,26 @@ The ``lua`` section outlines the configuration parameters related to the Lua env
...
+
+    Since version :doc:`3.3.0 </release/3.3.0>`, the ``lua_call`` option allows the specified user to perform the specified lua function on 
+    the instance during runtime.
```
Totktonada: Perform the Lua function? I guess you tell about granting a permission to call it, not about the call itself.

Totktonada: 'During runtime' and 'in runtime' feels a bit different. The first is just 'when the instance is run', but 'in runtime' may mean 'without persistence'. Maybe the formal documentation should use less informal words.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -3220,6 +3220,26 @@ The ``lua`` section outlines the configuration parameters related to the Lua env
...
+
+    Via the ``lua_call`` option, one can grant permissions to the function to any user registered on the instance. 
+
+    Note that the special option ``lua_call: [all]`` is also supported, granting access to all global Lua functions except built-in ones,   
+    bypassing database restrictions.
```
Totktonada: Imagine I've no needed context: it seems I have no chance to understand what does 'database restriction' mean.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -3220,6 +3220,26 @@ The ``lua`` section outlines the configuration parameters related to the Lua env
...
+
+    After the instance is rebooted, permissions defined via the ``lua_call`` options are reset to the values stored in the database.
```
Totktonada: Hmm. We discussed that and it is correct to some extent. However, the permissions that are granted using `lua_call` via the YAML config are both runtime and persistent. It is possible that we have no RW instance at the moment and the permission is not persisted, right. But usually it is persisted and read from the data files after restart.

Don't know how to better formalize it.

Totktonada:
It seems, we should work together here to provide a correct formal description. It would be nice to give a concise, but easy to understand explanation.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -1447,11 +1447,24 @@ credentials.users.*
...
     A list of global user-defined Lua functions that this user or a user with this role can call.
-    To allow calling all such functions, specify the ``all`` value.
+    To allow calling a specific function, specify its name as the value.
+    To allow calling all global Lua functions except built-in ones functions, specify the ``all`` value.
 
     This option should be configured together with the ``execute``
     :ref:`permission <configuration_reference_credentials_privileges_permissions>`.
 
+    To take a previously granted privilege away, specify the ``lua_call`` option without the function name, 
+    or with an empty privileges array (furhter options may retain commented-out).
```
Totktonada: I think that a section like 'Revoking priveleges' in https://www.tarantool.io/en/doc/latest/platform/connections_and_auth/credentials/#configuration-credentials would be a more appropriate place for this information.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -1447,11 +1447,24 @@ credentials.users.*
...
+
+    Since version :doc:`3.3.0 </release/3.3.0>`, the ``lua_call`` option allows granting users privileges to call specified lua function on 
+    the instance in runtime.
+
+    Example to grant and take away custom functions from the 'alice' user:
+
+    ..  literalinclude:: /code_snippets/snippets/config/instances.enabled/lua_call/config.yaml
+        :language: yaml
+        :dedent:
```
Totktonada: It goes after the 'since 3.3.0' remark and it makes unclear, whether it relates to all the versions or 3.3.0+.

doc/reference/configuration/configuration_reference.rst
```diff
@@ -1447,11 +1447,24 @@ credentials.users.*
...
+    To allow calling all global Lua functions except built-in ones functions, specify the ``all`` value.
 
     This option should be configured together with the ``execute``
     :ref:`permission <configuration_reference_credentials_privileges_permissions>`.
 
+    To take a previously granted privilege away, specify the ``lua_call`` option without the function name, 
+    or with an empty privileges array (furhter options may retain commented-out).
+
+    Since version :doc:`3.3.0 </release/3.3.0>`, the ``lua_call`` option allows granting users privileges to call specified lua function on 
+    the instance in runtime.
```
Totktonada: I would add: 'in runtime (thus it doesn't require an ability to write to the database)'.

doc/reference/configuration/configuration_reference.rst
Totktonada: Typo: furhter -> further.

doc/reference/configuration/configuration_reference.rst
Totktonada: Seems solved.

doc/reference/configuration/configuration_reference.rst
Totktonada: Seems solved.

Totktonada: Seems solved.

doc/reference/configuration/configuration_reference.rst
Totktonada: Seems solved.

doc/reference/configuration/configuration_reference.rst
Totktonada: There is no such a paragraph for now. Resolving the discussion.

doc/reference/configuration/configuration_reference.rst
AArdeev: Sure, i'll remove it from here and create another issue

doc/reference/configuration/configuration_reference.rst
AArdeev: Changed text and example

doc/reference/configuration/configuration_reference.rst
AArdeev: Added

doc/reference/configuration/configuration_reference.rst
AArdeev: removed

Totktonada: For the record: #5152.

Totktonada (approved):

PR #5150, opened 2025-06-03
Adds hierarchy table description

### LINKED ISSUE #4608: Configured names in `config:info()` and `box.info.config`
**Product:** Tarantool
**Since:** 3.3.0
**Root document:** TBD, possibly
- https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_info/info/
- https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_info/config/
- https://www.tarantool.io/en/doc/latest/reference/reference_lua/config/#config-api-reference-info
- https://www.tarantool.io/en/doc/latest/admin/upgrades/

**SME:** @ Totktonada 

# Details

An instance in the YAML configuration always has a name and always
corresponds to a particular replicaset and a particular group. This is
forced by the hierarchical configuration format.

These names are now reported in `config:info()` and `box.info.config`.

```yaml
# tarantool> config:info()
# or
# tarantool> config:info('v2')
# or
# tarantool> box.info.config
---
- <..status, alerts, meta..>
  hierarchy:
    group: group-001
    replicaset: replicaset-001
    instance: instance-001
...
```

The `hierarchy` field is an empty table if tarantool is started with a
Lua script or in the interactive mode without a script/a config.

Usually `box.info.name` equals to `box.info.config.hierarchy.instance`,
but there are situations, when only one of them is reported.

`box.info.config.hierarchy.instance` corresponds to the *configured*
instance name and it is reported when the YAML configuration is in use.
This instance name is written to the database as well, but the write
operation is postponed if the database schema version is lower then
needed to support the persistent names (see [1] for details).

A user can meet this situation while running tarantool 3.x on data from
tarantool 2.x, before the upgrading procedure is performed.

On the other hand, a user may use the persistent names feature (see [2])
together with the classic approach, when tarantool is run without the
configuration file, using a Lua script. In this case,
`box.info.config.hierarchy` is empty, but `box.info.name` shows the name
passed to `box.cfg({instance_name = <...>})` (and written to the
database).

The same is applicable to `box.info.replicaset.name` and
`box.info.config.hierarchy.replicaset`.

[1]: https://github.com/tarantool/doc/issues/4592
[2]: https://github.com/tarantool/doc/issues/3468
Requested by @Totktonada in https://github.com/tarantool/tarantool/commit/caf08d41c013a0b022289a28d04b959ef1cf70ba.

AArdeev:
* Since 3.3.0 the ``hierarchy`` table shows names of the group, replicaset, and the instance itself 
* Table is reported in ``config:info()``; ``config:info('v2')``; ``box.info.config`` 
* Fixes #4608

### DIFF
--- a/doc/reference/reference_lua/box_info/info.rst
+++ b/doc/reference/reference_lua/box_info/info.rst
@@ -17,7 +17,7 @@ box.info()
     :return: keys and values in the submodule
     :rtype:  table
 
-    Since version :doc:`3.3.0 </release/3.3.0>`
+    Since version :doc:`3.3.0 </release/3.3.0>`.
     Returns the ``hierarchy`` table, showing names of the group, replicaset, and the instance itself.
     These names are taken directly from the ``--name`` CLI option (or the ``TT_INSTANCE_NAME`` environment variable) 
     and the cluster configuration. This means they are always present if the YAML configuration flow is in use, 
--- a/doc/reference/reference_lua/box_info/info.rst
+++ b/doc/reference/reference_lua/box_info/info.rst
@@ -17,8 +17,7 @@ box.info()
     :return: keys and values in the submodule
     :rtype:  table
 
-    Since version :doc:`3.3.0 </release/3.3.0>`.
-    Returns the ``hierarchy`` table, showing names of the group, replicaset, and the instance itself.
+    Since version :doc:`3.3.0 </release/3.3.0>` returns the ``hierarchy`` table, showing names of the group, replicaset, and the instance itself.
     These names are taken directly from the ``--name`` CLI option (or the ``TT_INSTANCE_NAME`` environment variable) 
     and the cluster configuration. This means they are always present if the YAML configuration flow is in use, 
     disregarding the database status (whether upgraded, writable or not).
--- a/doc/reference/reference_lua/box_info/info.rst
+++ b/doc/reference/reference_lua/box_info/info.rst
@@ -19,6 +19,9 @@ box.info()
 
     Since version :doc:`3.3.0 </release/3.3.0>`
     Returns the ``hierarchy`` table, showing names of the group, replicaset, and the instance itself.
+    These names are taken directly from the ``--name`` CLI option (or the ``TT_INSTANCE_NAME`` environment variable) 
+    and the cluster configuration. This means they are always present if the YAML configuration flow is in use, 
+    disregarding the database status (whether upgraded, writable or not).
 
     **Example**
 
@@ -98,7 +101,7 @@ box.info()
               active: *0
             alerts: []
             hierarchy:
-              group: group-001
-              replicaset: replicaset-001
-              instance: instance-001
+              group: storages
+              replicaset: storage-a
+              instance: storage-a-002
         ...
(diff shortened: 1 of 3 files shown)


doc/reference/reference_lua/box_info/info.rst
```diff
@@ -94,4 +97,8 @@ box.info()
               last: &0 []
               active: *0
             alerts: []
+            hierarchy:
+              group: group-001
+              replicaset: replicaset-001
+              instance: instance-001
```
Totktonada: It is `storage-a-002` (see the persistent names above).

doc/reference/reference_lua/box_info/info.rst
```diff
@@ -94,4 +97,8 @@ box.info()
               last: &0 []
               active: *0
             alerts: []
+            hierarchy:
+              group: group-001
+              replicaset: replicaset-001
```
Totktonada: The replicaset name is `storage-a` (see above).

doc/reference/reference_lua/box_info/info.rst
```diff
@@ -94,4 +97,8 @@ box.info()
               last: &0 []
               active: *0
             alerts: []
+            hierarchy:
+              group: group-001
```
Totktonada: The related example (`doc/code_snippets/snippets/sharding/instances.enabled/sharded_cluster_crud/`) would issue `storages` as the group name.

doc/reference/reference_lua/config.rst
```diff
@@ -248,6 +264,10 @@ config API
                   8, but only the following values are allowed: 0, fatal, 1, syserror, 2, error,
                   3, crit, 4, warn, 5, info, 6, verbose, 7, debug'
                 timestamp: 2024-07-03T15:22:06.438275Z
+              hierarchy:
+                group: group-001
+                replicaset: replicaset-001
+                instance: instance-001
```
Totktonada: The alert message above shows other names.

doc/reference/reference_lua/config.rst
```diff
@@ -159,6 +159,10 @@ config API
 
                  -   ``alerts`` -- warnings or errors raised on an attempt to apply the configuration
 
+                Since version :doc:`3.3.0 </release/3.3.0>`
+
+                 -   ``hierarchy`` -- table, showing names of the group, replicaset, and the instance itself.
```
Totktonada, line 164: Should we say something like the following?

> Unlike the persistent `box.info.name` and `box.info.replicaset.name` values, these names are taken directly from the `--name` CLI option (or `TT_INSTANCE_NAME` environment variable) and the cluster configuration. It means that they always present if the YAML configuration flow is in use, disregarding the database status (upgraded or not, writable or not).

Totktonada (approved):
Mostly OK, left some consistency remarks.

doc/reference/reference_lua/config.rst
AArdeev: Not sure we need to refer to ``persistent box.info.name`` and ``box.info.replicaset.name`` values - this will invoke questions and call for references. I suggest:
"These names are taken directly from the ``--name`` CLI option (or the ``TT_INSTANCE_NAME`` environment variable) and the cluster configuration. This means they are always present if the YAML configuration flow is in use, disregarding the database status (whether upgraded, writable or not)."

PR #5232, opened 2025-07-04
Adds description of the on_event callback to custom application roles

### LINKED ISSUE #4666: `on_event` callback for roles
**Dev. issue:** https://github.com/tarantool/tarantool/issues/10538

**Product:** Tarantool
**Since:** 3.3.0
**Root document:** https://www.tarantool.io/en/doc/latest/platform/app/app_roles/
**SME:** @ grafin 

# Details

Roles can now set a `on_event` callback, which will be called
every time a `box.status` system event is broadcasted or the
configuration has been updated (after the `apply` has finished).

`on_event` callbacks are called one after another in the order
the roles are applied (according to their dependencies). Each
event (`box.status` or configuration change) triggers a series
of callbacks, next series (for a new event) will be triggered
after the previous finishes.

`on_event` callbacks are executed inside a `pcall` and if error is
raised, it is logged with 'error' level and the series execution
continues.

It is guaranteed that all of the `on_event` triggers are executed as
part of the configuration process (and 'ready' or 'check_warnings'
status is reached only after all of them are done).

When the callback is called, three arguments are provided,
`on_event(config, key, value)` where:
`config` is the current configuration,
`key` is `'config.apply'` if the callback was trigger by configuration
update, or `'box.status'` if it was triggered by `box.status` system
event.
`value` is the same as in `box.status` system event.

Example usage:

```lua
local config = require('config')

return {
    validate = ...,
    apply = ...,
    stop = ...,
    -- TODO: config parameter should contain role's config,
    -- but 3.3.0 passes the whole instance config to it. It
    -- is a bug (gh-10934).
    on_event = function(_config, key, value)
        local log = require('log')

        log.info('on_event is triggered by ' .. key)
        log.info('is_ro: ' .. value.is_ro)
        log.info('roles_cfg.my_role.foo: ' .. config:get('roles_cfg.my_role.foo'))
    end,
}
```
Requested by @grafin in https://github.com/tarantool/tarantool/commit/88e656cdf1cefbe7a30bdfe19099d08cec4ff443.

AArdeev:
* Added general info about custom roles as an intro to the section
* Fixes #4666

Deployment: https://docs.d.tarantool.io/ru/doc/gh-4666-on-event-callback-for-roles/platform/app/app_roles/#creating-a-custom-role

### DIFF
--- a/doc/platform/app/app_roles.rst
+++ b/doc/platform/app/app_roles.rst
@@ -81,16 +81,13 @@ The ``on_event`` callback returns 3 arguments, when it is called:
    - ``config.apply`` if the callback was triggered by a configuration update;
 
    - ``box.status`` if it was triggered by the ``box.status`` system event.
-- ``value``, which shows and logs the information about the instance status as in the trigger ``box.status`` system event.
-If the callback is triggered by a configuration update, the ``value`` shows the information of the most recent ``box.status`` system event.
+- ``value``, which shows and logs the information about the instance status as in the trigger ``box.status`` system event. If the callback is triggered by a configuration update, the ``value`` shows the information of the most recent ``box.status`` system event.
 
 ..  NOTE::
 
-- All ``on_event`` callbacks with the ``config.apply`` key  are executed as a part of the configuration process.
-Process statuses ``ready`` or ``check_warnings`` are reached only after all such ``on_event`` callbacks are done.
+   - All ``on_event`` callbacks with the ``config.apply`` key  are executed as a part of the configuration process. Process statuses ``ready`` or ``check_warnings`` are reached only after all such ``on_event`` callbacks are done.
 
-- All ``on_event`` callbacks are executed inside of a ``pcall``. If an error is raised for a callback,
-it is logged with the ``error`` level and the series execution continues.
+   - All ``on_event`` callbacks are executed inside of a ``pcall``. If an error is raised for a callback, it is logged with the ``error`` level and the series execution continues.
 
 Creating a custom role includes the following steps:
 
--- a/doc/platform/app/app_roles.rst
+++ b/doc/platform/app/app_roles.rst
@@ -90,7 +90,6 @@ The ``on_event`` callback provides 3 arguments, when it is called:
     - All ``on_event`` callbacks are executed inside of a ``pcall``. If an error is raised for a callback, it is logged with ``error`` level 
     and the series execution continues.
 
-
 Creating a custom role includes the following steps:
 
 #.  (Optional) Define the role configuration schema.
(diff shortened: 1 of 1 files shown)


doc/platform/app/app_roles.rst
```diff
@@ -81,6 +110,13 @@ As a result, a role module should return an object that has corresponding functi
         apply = function() -- ... -- end,
         stop = function() -- ... -- end,
         dependencies = { -- ... -- },
+        on_event = function(config, key, value)
```
grafin, line 113: Better to indent the function contents:
```lua
        on_event = function(config, key, value)
            local log = require('log')
            log.info('roles_cfg.my_role.foo: ' .. config.foo)
            log.info('on_event is triggered by ' .. key)
            log.info('is_ro: ' .. value.is_ro)
        end,
```

doc/platform/app/app_roles.rst
```diff
@@ -64,13 +64,42 @@ Creating a custom role
 Overview
 ~~~~~~~~
 
+A custom application role is an object which implements custom functions or logic adding to Tarantool's built-in roles and roles provided by third-party Lua modules.
+For example, a logging role can be created to add logging functionality on top of the built-in one.
+
+Since version :doc:`3.4.0 </release/3.4.0>`, you can define an ``on_event`` callback for custom roles. The ``on_event`` callback is called
+every time a ``box.status`` system event is broadcasted, or after the ``apply`` action of the configuration update is finished.
```
grafin: IMO "the ``apply`` action of " this is not needed here. `apply` step is a part of the current inner implementation, and should not be important for the users.

grafin (approved):
Thanks for the work! Just some minor comments from me.

doc/platform/app/app_roles.rst
```diff
@@ -64,13 +64,42 @@ Creating a custom role
 Overview
 ~~~~~~~~
 
+A custom application role is an object which implements custom functions or logic adding to Tarantool's built-in roles and roles provided by third-party Lua modules.
+For example, a logging role can be created to add logging functionality on top of the built-in one.
+
+Since version :doc:`3.4.0 </release/3.4.0>`, you can define an ``on_event`` callback for custom roles. The ``on_event`` callback is called
```
xuniq, line 70: 3.4.0 or 3.3.0?
I see 3.3.0 version in ticket info and in the changelog: https://github.com/tarantool/tarantool/releases/tag/3.3.0

AArdeev: 3.4.0 As agreed with SME

xuniq (approved):

PR #5538, opened 2026-01-15
What's new for release 3.6

lenkis:


doc/release/3.6.0.rst
```diff
@@ -0,0 +1,83 @@
...
+* Null layout columns passed to ``insert_arrow``,
+* Inserting into the middle of MemCS primary index with aggregates,
+* Inserting columns in the REE encoding (except for columns that are part of index keys or skip indexes)
+* Specifying per-column layout in space format and secondary index definitions,
+* LT, LE, and GT iterator types in Arrow streams,
+* ``index:quantile()``,
+* Bloom aggregates,
+* Statistics of index aggregates and scanner,
+* Index aggregates exported to their Lua objects.
+
```
sergepetrenko, line 114: I'm not sure if we should publish C API improvements in this section.

For the end users, there is no way to use this functionality without either their own development team, which makes use of C API features, or without getting an appropriate TCS release where this functionality is used.

Here are the bullets related to C API:
```
* Null layout columns passed to ``insert_arrow``,
* Inserting columns in the REE encoding (except for columns that are part of index keys or skip indexes)
* LT, LE, and GT iterator types in Arrow streams,
```

Maybe mention explicitly that these are the C API improvements?
Or not mention them at all?
Or mention them generally, like `Some C API improvements related to inserting and scanning data in Arrow format`?

I'm not sure which option we should take, what's your opinion on this?

lenkis: Good point. Let's wrap up as "C API improvements..." -- that's the closest to our "what's new" idea.

sergepetrenko: Thanks!

sergepetrenko (approved):

doc/release/3.6.0.rst
```diff
@@ -0,0 +1,115 @@
...
+The failover coordinator automatically decreases the quorum size if connection
+to the second data center is lost, and restores the cluster's operability within
+the available data center.
+When connection to the second data center is up again, the failover coordinator
+automatically restores the normal quorum size.
+So, no transactions are lost.
+
+However, there are some limitations.
+An important thing is that while data storages can be located in just 2 DCs,
+the configuration storage (based on etcd or Tarantool) must be located in yet another DC
```
sergepetrenko: I'm not sure I can parse this sentence correctly. Does it mean configuration storage has to be located in 3 datacenters or just in another DC, separate from the 2 used for storing data?

AFAIU, we should set up configuration storage in 3 DCs, the easiest way to do so is by using 2 existing "data" DCs + 1 extra "quorum" DC.

doc/release/3.6.0.rst
```diff
@@ -0,0 +1,115 @@
...
+
+Here are the cases when a topology with data storages located in 2 DCs may fail.
+
+**Case #1:** "interrupted data enrichment after DC failure"
+
+a. DC #1 goes down, while DC #2 is up. The failover coordinator decreases the quorum size
+   to 1. The cluster keeps serving the clients.
+b. DC #1 goes up and starts obtaining missed data from DC #2. The quorum size is still 1.
+c. The data enrichment process for DC #1 is still in progress, but DC #2 goes down now.
+   In this case, the cluster becomes unavailable.
```
sergepetrenko, line 84: I'm not sure about case 1, let's ask @georgiy-belyanin for clarifications. Is this really what's going to happen?

sergepetrenko (commented):
Thanks for the clarifications in the 2DC synchro section!
Please find a couple more comments below.

doc/release/3.6.0.rst
lenkis: Huh, I deduced this from Alex Klenov's feedback: "Чтобы всё устойчиво работало, нужно чтобы хранилище конфигурации располагалось на трёх площадках. То есть данные мы храним в двух ЦОДах. Но при этом у нас есть третяя площадка для принятия решений."

So, I've guessed it wrong that we need only 1 configuration storage here -- the one located in DC #3.

Re-iterated it with Alex. Indeed, we need 3 config storages. Thanks for spotting!

sergepetrenko: Thanks for the fix!

doc/release/3.6.0.rst
georgiy-belyanin: Thanks for reaching out. It seems to be described properly. The synchronous replication mechanism ensures no transactions are lost and in that case the cluster would remain stale until DC #1 is recovered.

PR #5666, opened 2026-05-18
3.7.0 what's new

lenkis:


Gumix (commented):
Please add one more limitation to dictionary and compression, like:

Currently the layout/compression options can be specified only during the space creation. It's not possible to change them for the existing space.

Gumix (approved):

PR #5683, opened 2026-06-04
doc: add storing data with memcs chapter

### LINKED ISSUE #5630: Dictionary-encoded columns in MemCS
Since: Tarantool EE 3.7

MemCS engine is populated with dictionary-encoding columns. They can be
created with `layout` option (here should be link to `layout` doc page):

```lua
local s = box.schema.create_space('test', {
    engine = 'memcs', format = format, field_count = field_count,
})
local pk = s:create_index('pk', {layout = 'dict'})
```

Only string non-key columns can use this layout. Other columns silently
ignore this option.

The maximal amount of unique values in such column (in other words,
maximal dictionary size) is UINT16_MAX (65536). When it's full, writes
of new unique values fail with an error. All indexes occupy 2 bytes
(`uint16` type is used for indexes under the hood). Hence, such column
will occupy `2 * space_size + dict_size` amount of bytes.

The dictionary is accounted in the `space:bsize()` statistics.

ArrowStream of dictionary-encoded columns always return values in
dictionary-encoded Arrow layout. The dictionary are returned in
string-view layout, the indexes have `uint16` type. When Arrow Stream
is used, we have some guarantees for dictionaries:
1. Unless the space was populated with a new unique value, all batches
   have the same dictionary.
2. Dictionaries are not copied so their dump to ArrowArray is cheap.
3. Dictionary can only grow, so values in the middle of the dictionary
   will never be deleted. Hence, after dictioniary was changed, it can
   be used for batches used old dictioinary.
Requested by @drewdzzz in https://github.com/tarantool/tarantool-ee/commit/cd7bd1ca233b1db4211f57439ef08364d4d27c6b.

maryiaLichko:
Fixes: https://github.com/tarantool/doc/issues/5630

Deployment: https://docs.d.tarantool.io/en/doc/doc-memcs-engine/platform/engines/memcs/

**done with AI help**


doc/platform/engines/memcs.rst
```diff
@@ -0,0 +1,223 @@
...
+.. _memcs-features:
+
+Key Features
+------------
+
+* Columnar data organization — data is stored column-wise, enabling efficient aggregations, filters, and scans.
+* Apache Arrow support — data can be exported in Arrow format without conversion, enabling zero-copy interoperability.
+* Dictionary encoding — reduces memory usage for string columns with repeated values.
+* `LZ4 <https://en.wikipedia.org/wiki/LZ4_(compression_algorithm)>`_ compression — compresses column data to reduce memory footprint.
+* SQL integration — supports querying via Tarantool SQL engine.
```
Gumix: SQL is not supported by memcs.

doc/platform/engines/memcs.rst
```diff
@@ -0,0 +1,223 @@
...
+
+MemCS supports a wide range of data types, including:
+
+- Integer types: ``uint64``, ``int64``, ``uint32``, ``int32``
+- Floating-point types: ``double``, ``float``
+- Strings: ``string``
+- Boolean: ``boolean``
+- Temporal types: ``datetime``
+- UUID: ``uuid``
+- Decimal: ``decimal``
```
Gumix: decimal, decimal32, decimal64, decimal128, and decimal256

doc/platform/engines/memcs.rst
```diff
@@ -0,0 +1,223 @@
...
+
+Supported Data Types
+--------------------
+
+MemCS supports a wide range of data types, including:
+
+- Integer types: ``uint64``, ``int64``, ``uint32``, ``int32``
+- Floating-point types: ``double``, ``float``
+- Strings: ``string``
+- Boolean: ``boolean``
```
Gumix: Booleans aren't supported correctly, let's remove them from here.

doc/platform/engines/memcs.rst
```diff
@@ -0,0 +1,223 @@
...
+
+
+.. _memcs-data:
+
+Supported Data Types
+--------------------
+
+MemCS supports a wide range of data types, including:
+
+- Integer types: ``uint64``, ``int64``, ``uint32``, ``int32``
```
Gumix: int8, uint8, int16, uint16, int32, uint32, int64, uint64

doc/platform/engines/memcs.rst
```diff
@@ -0,0 +1,223 @@
...
+- Floating-point types: ``double``, ``float``
```
Gumix: float32 and float64

doc/platform/engines/memcs.rst
```diff
@@ -0,0 +1,223 @@
...
+- Strings: ``string``
+- Boolean: ``boolean``
+- Temporal types: ``datetime``
```
Gumix: Not supported.

doc/platform/engines/memcs.rst
```diff
@@ -0,0 +1,223 @@
...
+- UUID: ``uuid``
```
Gumix: Not supported.

doc/platform/engines/memcs.rst
```diff
@@ -0,0 +1,223 @@
...
+        {name = "c3", type = "number"},
+        {name = "c4", type = "number", is_nullable = true, layout = 'null_rle'},
+        {name = "c5", type = "number", is_nullable = true},
+    })
+
+**Supported layouts:**
+
+- ``plain`` — default layout, no encoding
+- ``null_rle`` — RLE encoding for nullable fields
+
```
Gumix, line 168: and `dict` layout

doc/platform/engines/memcs.rst
```diff
@@ -0,0 +1,223 @@
...
+- **Analytical queries** with aggregations and column filters
+- **High-speed data transfer** via Arrow
+- **Efficient memory usage** through dictionary encoding and compression
+
+Using **dictionary encoding** and **LZ4 compression** together typically reduces memory usage by **8–13x** compared to uncompressed data.
+
+.. _memcs-memory:
+
+Memory Consumption
```
Gumix: Let's remove from here till the end, it looks too much AI.

doc/platform/engines/memcs.rst
```diff
@@ -0,0 +1,223 @@
...
+
+**ArrowStream guarantees:**
+
+- Dictionary is returned as ``string-view``, indices as ``uint16``
+- All batches share the same dictionary unless new unique values are inserted
+- Dictionaries are returned by reference (no copying), so ArrowArray export is cheap
+- Dictionaries only grow — previously produced batches remain compatible
+
+.. _memcs-column:
+
```
Gumix, line 111: Please add a chapter "RLE encoding of NULLs":

By default, NULL values are stored explicitly and use up the same space as any other valid column value (1, 2, 4, 8, 16 or 32 bytes depending on an exact field type), however RLE encoding of NULLs is also supported (`null_rle`). For reference, RLE-encoding of a column with 90% evenly distributed NULL values reduces memory consumption of that column by around 5 times.

doc/platform/engines/memcs.rst
```diff
@@ -0,0 +1,223 @@
+.. _engines-memcs:
+
+Storing data with memcs
+=======================
+
+..  admonition:: Enterprise Edition
+    :class: fact
+
+The `memcs` engine uses a single-threaded transaction processor (TX thread), similar to `memtx`. However, unlike `memtx`,
```
Gumix: Please change to something like:

The `memcs` engine uses a single-threaded transaction processor (TX thread), similar to `memtx`, and stores data in the memtx arena but in contrast to memtx it doesn’t organize data in tuples. Instead, it stores data in columns. Each format field is assigned its own BPS tree-like structure (BPS vector), which stores values only of that field. If the field type fits in 32 bytes, raw field values are stored directly in tree leaves without any encoding. The strings are stored in the format similar to "Arrow Variable-size Binary View Layout", also called "German Strings".

The main benefit of such data organization is a significant performance boost of columnar data sequential scans compared to memtx thanks to CPU cache locality. That’s why memcs supports a special C api for such columnar scans: see `box_index_arrow_stream()` and `box_raw_read_view_arrow_stream()`. Peak performance is achieved when scanning embedded field types.

Querying full tuples, like in memtx, is also supported, but the performance is worse compared to memtx, because a tuple has to be constructed on the runtime arena from individual field values gathered from each column tree.

Other features include:
- Point lookup.
- Stable iterators.
- Insert / replace / delete / update.
- Batch insertion in the Arrow format.
- Transactions, including cross-engine transactions with memtx (with `memtx_use_mvcc_engine = false`).
- Read view support.
- Secondary indexes with an ability to specify covered columns and sequentially scan indexed + covered columns.

Gumix (approved):

PR #5696, opened 2026-07-03
box.error: added e:unpack() raise example

tmr-g:


doc/reference/reference_lua/box_error.rst
```diff
@@ -175,6 +175,51 @@ Error details may include an error code, type, message, and trace.
     :end-before: Get error details: end
     :dedent:
 
+.. NOTE::
+
+    You can use the error details table to raise an error, but that error would not be as the original.
```
AArdeev: You can use the error details table to raise an error, but that error would not be the same as the original.

AArdeev (approved):

doc/reference/reference_lua/box_error.rst
tmr-g: Thanks! Fixed

PR #5729, opened 2026-09-11
doc: drop layout specification at index level

nshy:
We drop layout specification at index level altogether in https://github.com/tarantool/tarantool-ee/commit/6a01330df4fea5c608b2b9b36b137fa83e9269b8. So let's clean it from documentation skipping usual procedure with TarantoolBot request to avoid users confusion.

Created with the help of AI.


nshy:
PR is will be redone in Gitlab repo.

